PR #3, opened 2021-09-11
Fix links in registration email

adelq:
Missing forward slash in privacy policy link and remove HTML from plaintext

I am not experienced with elixir and was not able to test this pull request but is relatively self-explanatory.


Teifion:
A fantastic pull request, especially for someone not used to Elixir. Approved!

PR #11, opened 2021-10-31
name + password handler for get_token

Jazcash:


Teifion:
As per discussion on Discord we won't be adding getting users based on name.

PR #23, opened 2022-05-08
Help for individual commands

L-e-x-o-n:


lib/teiserver/coordinator/coordinator_lib.ex
```diff
@@ -2,166 +2,106 @@ defmodule Teiserver.Coordinator.CoordinatorLib do
...
+      {"dosplit", [], "Completes the lobby split now.", :moderator},
+      {"cancelsplit", [], "Cancels the lobby split.", :moderator},
+      {"vip", ["name"], "Places that user at the front of the queue. This command will always output it's use even if used with the % operator.", :moderator},
+      {"reset", [], "Resets the coordinator bot for this lobby to the default.", :moderator},
+    ]
     # $command - Coordinator command
     # $%command - Don't echo command back to chat
+  result = commands
+    |> Enum.filter(fn {cmd, _args, _desc, group} ->
+      (cmd == command or command == "") and can_use?(user, host, group)
```
Teifion, line 86: Why `or command == ""` ?

lib/teiserver/coordinator/coordinator_lib.ex
```diff
@@ -2,166 +2,106 @@ defmodule Teiserver.Coordinator.CoordinatorLib do
...
+    [{cmd, args, desc, _group}] ->
+      arg_str = args |> Enum.map(fn a -> "<#{a}>" end) |> Enum.join(" ")
+      "#{cmd} #{arg_str}\n#{desc}"
+    _ ->
+      if command != "" do
+        "No commands matching that filter"
+      else
+
+
+        Enum.map(result, fn {cmd, args, desc, _group} ->
```
Teifion, line 99: Please use the style:

```
result
|> Enum.map(fn
```

lib/teiserver/coordinator/coordinator_lib.ex
```diff
@@ -2,166 +2,106 @@ defmodule Teiserver.Coordinator.CoordinatorLib do
...
+        arg_str = args |> Enum.map(fn a -> "<#{a}>" end) |> Enum.join(" ")
+        "\n#{cmd} #{arg_str}\n#{desc}\n" end)
```
Teifion, line 101: the `end)` should be on it's own line as the opening is on a different line

lib/teiserver/coordinator/coordinator_lib.ex
L-e-x-o-n: It's for the case where only `$help` is given with no specific command. I wanted the `result` to have all commands and this looked like a good place to do it and filter commands that don't match the user's group.

lib/teiserver/coordinator/coordinator_lib.ex
L-e-x-o-n: Ok

lib/teiserver/coordinator/coordinator_lib.ex
L-e-x-o-n: Ok

lib/teiserver/coordinator/coordinator_lib.ex
Teifion: Works for me, I'm happy with that solution :)

test/teiserver/coordinator/coordinator_commands_test.exs
```diff
@@ -32,6 +32,73 @@ defmodule Teiserver.Coordinator.CoordinatorCommandsTest do
...
+    message_coordinator(socket, "$help pull")
+    [reply] = _tachyon_recv(socket)
+    assert reply == %{"cmd" => "s.lobby.send_direct_message", "result" => "success"}
+
+    [reply] = _tachyon_recv(socket)
+    message = reply["message"]
+
+    assert Enum.member?(message, "No commands matching that filter.")
+    assert Enum.count(message) == 1
+    assert not Enum.member?(message, "Pulls a given user into the battle.")
```
Teifion, line 86: instead of `assert not` use `refute`

test/teiserver/coordinator/coordinator_commands_test.exs
```diff
@@ -32,6 +32,73 @@ defmodule Teiserver.Coordinator.CoordinatorCommandsTest do
...
+    #Normal pull test
+    message_coordinator(socket, "$help pull")
+    [reply] = _tachyon_recv(socket)
+    assert reply == %{"cmd" => "s.lobby.send_direct_message", "result" => "success"}
+
+    [reply] = _tachyon_recv(socket)
+    message = reply["message"]
+
+    assert Enum.member?(message, "No commands matching that filter.")
+    assert Enum.count(message) == 1
```
Teifion, line 85: In this case you can use `assert message == ["No commands matching that filter."]`, this will work the same as both the above lines

PR #25, opened 2022-05-15
Joinq on full battle room

L-e-x-o-n:
Trying to join a full battle room as a player will add you to the join queue.


L-e-x-o-n:
Don't merge yet

PR #39, opened 2022-06-06
Fix for missing user data in s.auth.login response

### LINKED ISSUE #31: Missing user data in s.auth.login response
`country` and `icons` properties are returned for users in the `s.user.user_and_client_list` response, but that data seems to be missing from `s.auth.login`. I suspect it'll be missing from `s.auth.verify` too

L-e-x-o-n:
Fixes #31.
Adds icons to s.auth.login and s.auth.verify, Teifion added country.

### DIFF
--- a/lib/teiserver/protocols/tachyon_v1/auth_out.ex
+++ b/lib/teiserver/protocols/tachyon_v1/auth_out.ex
@@ -42,11 +42,10 @@ defmodule Teiserver.Protocols.Tachyon.V1.AuthOut do
   end
 
   def do_reply(:verify, {:success, user}) do
-    icons = %{ icons: %{"play_time_rank" => user.rank}}
     %{
       "cmd" => "s.auth.verify",
       "result" => "success",
-      "user" => Tachyon.convert_object(:user_extended, Map.merge(user, icons))
+      "user" => Tachyon.convert_object(:user_extended_icons, user)
     }
   end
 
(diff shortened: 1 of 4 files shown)


lib/teiserver/protocols/tachyon_v1/auth_out.ex
```diff
@@ -42,10 +42,11 @@ defmodule Teiserver.Protocols.Tachyon.V1.AuthOut do
   end
 
   def do_reply(:verify, {:success, user}) do
+    icons = %{ icons: %{"play_time_rank" => user.rank}}
```
Teifion, line 45: Instead of manually adding the icon here, it should be added to the list of icons generated for a user. The rank icon won't be specific to just Tachyon and anything else (e.g. website) that uses the icon list will have to duplicate this code.

test/teiserver/protocols/tachyon/v1/tachyon_raw_test.exs
```diff
@@ -71,13 +71,15 @@ defmodule Teiserver.Protocols.V1.TachyonRawTest do
     existing_user = new_user()
     data = %{cmd: "c.auth.register", username: "test_name", email: existing_user.email, password: "password"}
     _tachyon_send(socket, data)
+    :timer.sleep(3000)
```
Teifion, line 74: Why a sleep period of 3000?

L-e-x-o-n: Test weren't passing without timer for me and 3000 worked so I didn't try lowering it. I did some more tests, sometimes it works with as low as 10 but not always. 300 seems to work reliably. Removed in last commit.

lib/teiserver/protocols/tachyon_v1/auth_out.ex
L-e-x-o-n: Fixed.

PR #44, opened 2022-07-30
Returns a reply if a given command doesn't exist

L-e-x-o-n:


lib/teiserver/coordinator/consul_server.ex
```diff
@@ -255,6 +255,12 @@ defmodule Teiserver.Coordinator.ConsulServer do
         Coordinator.cast_coordinator({:consul_command, Map.merge(cmd, %{lobby_id: state.lobby_id, host_id: state.host_id})})
         {:noreply, state}
 
+      Enum.member?(@always_allow, cmd.command) == false and
+      Enum.member?(@boss_commands, cmd.command) == false and
+      Enum.member?(@host_commands, cmd.command) == false ->
+        LobbyChat.sayprivateex(state.coordinator_id, cmd.senderid, "No command of name '#{cmd.command}'", state.lobby_id)
+        {:noreply, state}
```
Teifion, line 262: Moderator commands are not part of these lists, this means every time a moderator command is called it will be caught by this block instead of the allowed check.

L-e-x-o-n: Fixed

lib/teiserver/coordinator/consul_server.ex
```diff
@@ -651,9 +651,31 @@ defmodule Teiserver.Coordinator.ConsulServer do
       client == nil -> false
       Enum.member?(@always_allow, cmd.command) -> true
       client.moderator == true -> true
-      Enum.member?(@host_commands, cmd.command) and is_host -> true
-      Enum.member?(@boss_commands, cmd.command) and (is_host or is_boss) -> true
-      true -> false
+
+      true ->
+        false
```
Teifion, line 656: This `false` doesn't do anything

PR #51, opened 2022-08-29
Added tzdata package to get timezones

Benluc:
Mac-Users are not able to get timezone information with `timedatectl list-timezones`, because of that I added a general solution for getting a list of timezones by adding the hex package [tzdata](https://hexdocs.pm/tzdata/readme.html)


Teifion:
A nice fix, thank you!

PR #58, opened 2022-09-30
Fixes typo

alenros:
removes extra colon


Teifion:
Thanks for the fix :)

Issue #67, opened 2022-10-14
Trying to add non-exisitng entity to Clan results in "User already in clan." error


p2004a:
When you open the clans invite page and in the search form type some random characters and press "add member", you get "User already in clan." error on top, instead of some "invalid user" which is a bit confusing.

PR #75, opened 2023-02-27
Add user id check for clan invites

L-e-x-o-n:


lib/teiserver_web/controllers/clans/clan_controller.ex
```diff
@@ -259,6 +259,11 @@ defmodule TeiserverWeb.Clans.ClanController do
     clan = Clans.get_clan!(clan_id)
 
     cond do
+      user_id == nil ->
+        conn
+        |> put_flash(:danger, "User id missing")
```
Teifion, line 264: It's going to be seen by non-developers, I'd rather something more like "Unable to find that user"

Teifion:
I like it

Teifion:
Addresses #67

lib/teiserver_web/controllers/clans/clan_controller.ex
L-e-x-o-n: Changed to "Unable to find user"

PR #76, opened 2023-02-27
Added minimum players option to $splitlobby

L-e-x-o-n:
`$splitlobby` command can now be used with a minimum number of players.

Examples:
- `$splitlobby 4` 
Will pass only if at least 4 people agreed to split to the new lobby.
 - `$splitlobby` 
Works as before (sets minimum number to 1).

In addition to this, fixed a bug where the person who calls  `$splitlobby` would always be moved to a new lobby even if no one else voted to move.

### DIFF
--- a/lib/teiserver/coordinator/consul_commands.ex
+++ b/lib/teiserver/coordinator/consul_commands.ex
@@ -243,7 +243,16 @@ defmodule Teiserver.Coordinator.ConsulCommands do
     ConsulServer.say_command(cmd, state)
     sender_name = User.get_username(senderid)
 
-    min_players = get_default_min_players(rem)
+    min_players = case String.trim(rem) do
+      "" ->
+        1
+
+      _ ->
+        rem
+        |> String.trim
+        |> String.to_integer
+        |> max(1)
+    end
 
     LobbyChat.sayex(state.coordinator_id, "Split lobby sequence started ($y to move, $n to cancel, $follow <name> to follow user)", state.lobby_id)
 
@@ -1131,17 +1140,6 @@ defmodule Teiserver.Coordinator.ConsulCommands do
         |> Map.get(userid)
       false ->
         -1
-    end
-  end
-
-  defp get_default_min_players(str) do
-    case String.trim(str) do
-      "" ->
-      1
-
-      _ ->
-        String.to_integer(str |> String.trim)
-        |> max(1)
     end
   end
 
--- a/lib/teiserver/coordinator/consul_commands.ex
+++ b/lib/teiserver/coordinator/consul_commands.ex
@@ -239,9 +239,13 @@ defmodule Teiserver.Coordinator.ConsulCommands do
     state
   end
 
-  def handle_command(%{command: "splitlobby", senderid: senderid} = cmd, %{split: nil} = state) do
+  def handle_command(%{command: "splitlobby", remaining: rem, senderid: senderid} = cmd, %{split: nil} = state) do
     ConsulServer.say_command(cmd, state)
     sender_name = User.get_username(senderid)
+
+    min_players = get_default_min_players(rem)
+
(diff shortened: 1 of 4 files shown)


lib/teiserver/coordinator/coordinator_lib.ex
```diff
@@ -21,8 +21,8 @@ defmodule Teiserver.Coordinator.CoordinatorLib do
       {"status", [], "Status info about the battle lobby.", :everybody},
       {"afks", [], "Lists possible afk players.", :everybody},
       {"password?", [], "Tells you the room password", :everybody},
-      {"splitlobby", [], "Causes a \"vote\" to start where other players can elect to join you in splitting the lobby, follow someone
-of their choosing or remain in place. After 20 seconds you are moved to a new (empty) lobby and those that voted yes
+      {"splitlobby", ["minimum players"], "Causes a \"vote\" to start where other players can elect to join you in splitting the lobby, follow someone
+of their choosing or remain in place. After 30 seconds, if at the minimum number of players agreed to split, you are moved to a new (empty) lobby and those that voted yes
```
StanczakDominik, line 25:
```suggestion
of their choosing or remain in place. After 30 seconds, if at least the minimum number of players agreed to split, you are moved to a new (empty) lobby and those that voted yes
```

lib/teiserver/coordinator/consul_commands.ex
```diff
@@ -1129,6 +1134,17 @@ defmodule Teiserver.Coordinator.ConsulCommands do
...
 
+  defp get_default_min_players(str) do
+    case String.trim(str) do
+      "" ->
+      1
+
+      _ ->
+        String.to_integer(str |> String.trim)
+        |> max(1)
+    end
```
StanczakDominik, line 1145: Could you explain this part? As a newbie to Elixir this looks pretty magical 😅

StanczakDominik (approved):
The logic itself looks pretty reasonable! That's about as much as I can say.

lib/teiserver/coordinator/consul_commands.ex
L-e-x-o-n: We get the "remaining" string from the handle command function and pass it to this.
This function is used to check if the string is empty (no min number given) by comparing the trimmed string to an empty one.
If it is empty we return 1 as a default min player number value, otherwise we convert it to integer and return the bigger value between it and 1 (in case people enter non positive numbers)

Teifion: Good explanation Lexon :)

I do have a few things to raise on the way it's written (the logic is sound, it's purely Elixir idioms).

```elixir
String.to_integer(str |> String.trim)
|> max(1)
```

I'd replace it with
```
str
|> String.trim
|> String.to_integer
|> max(1)
```

lib/teiserver/coordinator/consul_commands.ex
```diff
@@ -239,17 +239,21 @@ defmodule Teiserver.Coordinator.ConsulCommands do
     state
   end
 
-  def handle_command(%{command: "splitlobby", senderid: senderid} = cmd, %{split: nil} = state) do
+  def handle_command(%{command: "splitlobby", remaining: rem, senderid: senderid} = cmd, %{split: nil} = state) do
     ConsulServer.say_command(cmd, state)
     sender_name = User.get_username(senderid)
 
+    min_players = get_default_min_players(rem)
```
Teifion, line 246: Given the contents and wider file I'd suggest inlining (don't make it a function, put it right here) this function.

Teifion:
Minus the relatively small issues I've flagged up this is a really nice fix. I look forwards to merging it :)

lib/teiserver/coordinator/consul_commands.ex
L-e-x-o-n: Done

lib/teiserver/coordinator/consul_commands.ex
L-e-x-o-n: Done

PR #77, opened 2023-02-28
Allows all party members to create invites

L-e-x-o-n:
This PR should allow every party member to send invites as discussed on Discord.
Previously only the party leader (and moderators) could send party invites.


Teifion:
As per the discord discussion, there are other places this check would need to go (Tachyon `party_in.ex` being the one that was found at the time).

PR #81, opened 2023-03-19
fix fake data user generation to be able to connect via spring lobbies

panzerr:
add a md5 hashed version of the password in password_hash


Teifion:
Nice first PR :)

PR #84, opened 2023-03-22
Readded playtime to $whoami

L-e-x-o-n:


lib/teiserver/bridge/message_commands.ex
```diff
@@ -167,7 +167,7 @@ defmodule Teiserver.Bridge.MessageCommands do
     msg = [
       "You are #{user.name}",
       "Profile link: #{profile_link}",
-      "Rank: #{user.rank+1} with #{player_hours} player hours and #{spectator_hours} spectator hours for a rank hour count of #{rank_time}",
+      "Your total time in BAR (playing, spectating, menu, lobby) is #{total_hours} hours of which you played for #{player_hours} hours and spectated for #{spectator_hours} hours",
```
Teifion, line 170: This is overly verbose and doesn't show rank any more; could you tweak it a bit please?

L-e-x-o-n: Would "Playtime: 100 h (80 playing, 20 spectating)" be better?
Rank was intentionally not shown, I thought there was no reason to show it as it was no longer used in icons.

L-e-x-o-n: Finally changed

PR #100, opened 2023-03-26
Rating graph

panzerr:
Add a page with rating progression graphs to the web interface (similar to the matches ui but with a different presentation)

### DIFF
--- a/lib/teiserver_web/controllers/battle/match_controller.ex
+++ b/lib/teiserver_web/controllers/battle/match_controller.ex
@@ -4,6 +4,7 @@ defmodule TeiserverWeb.Battle.MatchController do
   alias Teiserver.{Battle, Game, Account}
   alias Teiserver.Game.MatchRatingLib
   alias Teiserver.Battle.MatchLib
+  alias Central.Helpers.TimexHelper
 
   plug Bodyguard.Plug.Authorize,
     policy: Teiserver.Battle.Match,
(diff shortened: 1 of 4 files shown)


lib/teiserver_web/controllers/battle/match_controller.ex
```diff
@@ -132,4 +133,52 @@ defmodule TeiserverWeb.Battle.MatchController do
...
+        rating_type_id: filter_type_id
+      ],
+      order_by: "Newest first",
+      limit: 50,
+      preload: [:match, :match_membership]
+    )
+
+    {dates, ratings_data} =
+      logs
+      |> List.foldl( %{}, fn rating, acc ->
```
Teifion, line 167: Putting spaces after the parenthesis isn't used here so the code looks all wrong.

Teifion:
I like that you've done things differently to how I would, as soon as the whitespace stuff is sorted I'll merge.

As it adds a new route it'll require a server restart, I can't just hotfix it in sadly :(

panzerr:
I should have fixed most whitespace issue now,
note: I tried to run mix format to see and it generates a lot of changes but It may be a good idea to do it sometime on the project since it simplifies these kind of things if there are multiple people working on the same codebase

Teifion:
I'll do a mix format soon, I'd _intended_ to prior to this but never got around to it.

PR #112, opened 2023-05-03
Fix case sensitive email field and add tests

Penguin2600:
**Issue:** 
Login page at https://server4.beyondallreason.info/ is case-sensitive for email. Meaning I might be providing the correct email but in the wrong case and I get an "Invalid Credentials" error.

**Proposed Fix:**
- Added tests to expose case-sensitivity
- Added user query for case-insensitive email
- Modified several calls to use case-insensitve email query

I ran the server locally and confirmed this is solving the original bug, I can now type in my email in any cAsE and login successfully.

### DIFF
--- a/lib/central/account/queries/user_queries.ex
+++ b/lib/central/account/queries/user_queries.ex
@@ -77,6 +77,11 @@ defmodule Central.Account.UserQueries do
       where: users.email == ^email
   end
 
+  def _search(query, :email_lower, value) do
+    from users in query,
+      where: lower(users.email) == ^String.downcase(value)
+  end
+
   def _search(query, :name_or_email, value) do
     from users in query,
       where: users.email == ^value or users.name == ^value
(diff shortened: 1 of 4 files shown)


Penguin2600:
Go easy on me :) This is my first time in Elixir, what an.... interesting framework and language.

lib/central/account/queries/user_queries.ex
```diff
@@ -77,6 +77,11 @@ defmodule Central.Account.UserQueries do
       where: users.email == ^email
   end
 
+  def _search(query, :email_lower, value) do
```
Penguin2600, line 80: copied this pattern from the user search where there is a "name" and a "name_lower" query available

https://github.com/beyond-all-reason/teiserver/blob/4c73f45609db6bc93aecbc18990559ab93c6c0e4/lib/central/account/queries/user_queries.ex#L70

PR #113, opened 2023-05-05
Prevent people reporting themselves

### LINKED ISSUE #93: Prevent people reporting themselves
Users can currently report themselves, they should not be able to.

Penguin2600:
Adds a filter to remove report button next to your own name.

Adds checks at the endpoint to deny any requests where submitter id == target id

closes #93


Penguin2600:
@Teifion I was trying to test this and came up with the following but I think the test framework is cycling the user ID out from under me or something? I tried to grab it from the conn but that was also a nogo:

```
  describe "submit matching target and reporter" do
    test "returns error and does not create report", %{conn: conn} do
      conn_target_id = conn.private.guardian_default_resource.id
      assert Enum.empty?(Moderation.list_reports(search: [target_id: conn_target_id]))
      attrs = %{
        "target_id" => conn_target_id,
        "type" => "type",
        "sub_type" => "sub_type"
      }
      conn = assign(conn, :current_user, %{id: conn_target_id})
      conn = post(conn, Routes.moderation_report_form_path(conn, :create), report: attrs)
      assert html_response(conn, 422) =~ "Reporter and target may not be the same user."
      assert Enum.count(Moderation.list_reports(search: [target_id: conn_target_id])) == 0
    end
  end
```
  
  Any thoughts on what I might try next?
  
  Thanks!

Teifion:
I'd suggest looking at the other conn tests, they have the user logged in already. You're not actually logging the user in correctly here.

Penguin2600:
@Teifion Got the tests working as expected! lemme know what you think.

lib/teiserver_web/controllers/moderation/report_form_controller.ex
```diff
@@ -47,63 +47,67 @@ defmodule TeiserverWeb.Moderation.ReportFormController do
...
+          type: report["type"],
+          sub_type: report["sub_type"],
+          extra_text: report["extra_text"],
+          match_id: match_id,
+          relationship: relationship
+        })
+
+      case result do
+        {:ok, report} ->
+          IO.puts("")
```
Teifion, line 97: I think the IO part should be removed. Looks like I left it there by mistake in the first place!

lib/teiserver_web/controllers/moderation/report_form_controller.ex
```diff
@@ -47,63 +47,67 @@ defmodule TeiserverWeb.Moderation.ReportFormController do
...
-                  reporter_member == nil -> nil
-                  target_member.team_id == reporter_member.team_id -> "Allies"
-                  target_member.team_id != reporter_member.team_id -> "Opponents"
-                end
-
-              {match.id, relationship}
-          end
+    if conn.assigns.current_user.id == target_id do
+      conn
+      |> put_status(:unprocessable_entity)
```
Teifion, line 52: This is a HTML request, if it errors we should be presenting people with a HTML response.

test/teiserver_web/controllers/moderation/report_form_controller_test.exs
```diff
@@ -37,4 +37,23 @@ defmodule TeiserverWeb.Moderation.ReportFormControllerTest do
...
+      user = GeneralTestLib.make_user()
+      assert Enum.empty?(Moderation.list_reports(search: [target_id: user.id]))
+
+      attrs = %{
+        "target_id" => user.id-1,
+        "type" => "type",
+        "sub_type" => "sub_type"
+      }
+
+      IO.inspect(attrs)
```
Teifion, line 52: I think the IO should be removed too.

lib/teiserver_web/controllers/moderation/report_form_controller.ex
Penguin2600: I changed the content type and tested that it does now result in an HTML response I didnt know if you wanted a full template or something here, just let me know if you want to see a more sophisticated error page.

Teifion: There should be a moderately user-friendly error page.

Penguin2600: @Teifion I have added a "self_report" page similar to the existing "no_user" page. let me know if that fits the bill!

Penguin2600:
@Teifion ready for re-review at your convenience.

lib/teiserver_web/controllers/moderation/report_form_controller.ex
```diff
@@ -47,63 +47,63 @@ defmodule TeiserverWeb.Moderation.ReportFormController do
...
+      |> render("self_report.html", message: "Reporter and target may not be the same user.")
```
Teifion, line 53: While the statement is accurate, it doesn't read well. I'd rather something like "You cannot report yourself", we are speaking to the reporter themselves rather than a 3rd party trying to enter it into an admin control panel.

lib/teiserver_web/controllers/moderation/report_form_controller.ex
Teifion: Thanks :)

lib/teiserver_web/templates/moderation/report_form/self_report.heex
```diff
@@ -0,0 +1,19 @@
+<div class="row" style="padding-top: 15vh;">
```
Teifion, line 1: Ideally this would be a `.html.heex` file rather than just `.heex`

Teifion:
A few more issues but I think it's nearly done, apologies for the late reply.

Penguin2600:
> A few more issues but I think it's nearly done, apologies for the late reply.

No worries! Thanks for the review, changed those items as requested.

PR #116, opened 2023-05-15
Lingering Friends/Mutes Cleanup

### LINKED ISSUE #35: Lingering friends/mutes
When a user is deleted they are not removed from the friend/mute lists of other users. They should be.

Penguin2600:
- Adds test that exposes remnant data

Closes #35


Penguin2600:
@Teifion I'm on my way to just understanding the scope of the issue and as part of that I wanted to start with a test that exposes the problem. This is what I have come up with, does this test expose the full issue as you understand it as well or have I missed anything?

Teifion:
This has now been addressed in a recent change to the way friends work.

PR #119, opened 2023-05-28
Sticky footer

### LINKED ISSUE #29: Sticky footer
Currently the footer in [standard.html.heex](https://github.com/beyond-all-reason/teiserver/blob/master/lib/central_web/templates/layout/standard.html.heex) is hiding the footer as it isn't sticky.

When the page is small the footer should stick to the bottom of the screen, when the page is large the footer should appear at the bottom of the page after all the content and only visible when the scrollbar is at the maximum level.

fumbleforce:
Credit is important.

Apparently my editor ate some unused indentations as well.

Have not seen any bad side effects of the body/html changes, looking through existing pages.

Screenshot_20230528_101416
Screenshot_20230528_101356
Screenshot_20230528_101328

 fixes #29

### DIFF
--- a/assets/scss/custom/_styles.scss
+++ b/assets/scss/custom/_styles.scss
@@ -4,6 +4,11 @@
   margin: 0;
 }
 
+html, body {
+  min-height: 100%;
+  height: 100%;
+}
+
 body {
   overflow-x: hidden;
   margin: 0;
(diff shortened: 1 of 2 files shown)


Teifion:
Brilliant fix, thank you for addressing it :)

PR #122, opened 2023-06-05
Feature/add uncertainty graph to progression

### LINKED ISSUE #111: Add upper and lower bounds of certainity visualisation to the rating progression plot
add uncertainty to the graph as + and - bounds of the plot
https://server4.beyondallreason.info/teiserver/battle/matches/ratings_graph 

examples: 
image
image

kozak-codes:
Closes #111 

Had to switch to d3 because c3 didn't have any kind of area graph. I've only used d3 a little bit before so it's probably pretty messy. Also, I wasn't quite sure how to apply scss or where it should go, so a lot of the styling is inline in the page.

Let me know if you'd like any other colouring changes, etc. We could also add a list of matches won within the tooltip, but that may be somewhat overkill.

Let me know how I can improve! I'm still pretty new to Elixir and still learning so let me know if you notice any bad patterns.

### DIFF
--- a/lib/teiserver_web/controllers/battle/match_controller.ex
+++ b/lib/teiserver_web/controllers/battle/match_controller.ex
@@ -169,32 +169,25 @@ defmodule TeiserverWeb.Battle.MatchController do
         preload: [:match, :match_membership]
       )
 
-    data =
-      logs
-      |> Enum.map(fn rating -> %{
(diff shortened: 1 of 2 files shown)


Teifion:
I'm not super thrilled about the use of d3 instead of c3 but I wasn't able to find a way either so accepting it :)

PR #125, opened 2023-06-09
Report form adjustments - required description

StanczakDominik:
This makes the description field required for reports and adjusts the wording a little so that the information provided is a little more helpful :)


Teifion:
This doesn't force there to be a description in those cases, I'd like to see server-side logic enforcing this.

StanczakDominik:
You're right, sorry. I'm a bit of a newbie to web dev and, come to think of it, I think I've seen this exact approach as a meme on r/programmerhumor. I'll figure it out :)

Teifion:
@StanczakDominik Does this need to be open?

StanczakDominik:
Not at all, thanks for the reminder!

PR #126, opened 2023-06-10
More info on Ratings page

StanczakDominik:
This does a few things:

- adds Skill Rating and Uncertainty directly. People were already solving sets of equation to get at it, let's not force them to.
- Ties the color associated with the row to the difference in skill rating after the match (green on a win, red on a loss). This was previously done from delta(Balance Rating). Since that's capped at 0, the diffs would show 0 on wins, for example, [in this case](https://discord.com/channels/549281623154229250/1115432854919327745/1116833556417757274).
- Adds notes on how Balance and Leaderboard ratings are calculated and displayed
- Clarifies which rating is being taken for the peak/change summary calculations (it's balance).
- While for many, Supreme Strait may be the name of the game, I still renamed the column from "Game" to "Map name" ;)

This is how it looks for me with http://localhost:4000/teiserver/battle/matches/ratings?filter=Team. I know the colors look weird, but I have a sneaking suspicion that it's because the skill ratings and changes for test data are being generated inconsistently, as two independent random numbers for each match entry, while one the former *should* (but it adds a fair bit of complexity so let's ignore that) be a cumulative sum of the latter.

image

I don't *think* it looks too noisy, but that's me :)

I was also iffy on whether delta uncertainty should be colored, as it most likely will always remain negative, but it's simpler that way and I don't think it hurts at all.

### DIFF
--- a/lib/teiserver_web/templates/battle/match/ratings.html.heex
+++ b/lib/teiserver_web/templates/battle/match/ratings.html.heex
@@ -48,7 +48,7 @@
             </div>
 
             <div class="col-md-3">
-              Peak Balance Rating: <%= Enum.max_by(@logs, fn l -> l.value["rating_value"] end).value[
+              Peak balance rating: <%= Enum.max_by(@logs, fn l -> l.value["rating_value"] end).value[
                 "rating_value"
               ]
               |> round(2) %>
@@ -69,9 +69,8 @@
               <th>Players</th>
               <th>Type</th>
               <th colspan="2">Skill Rating</th>
-              <th colspan="2">Uncertainty</th>
-              <th>Balance rating<br /> (Skill - Uncertainty, non-negative)</th>
-              <th>Leaderboard rating<br /> (Skill - 3 x Uncertainty)</th>
+              <th colspan="2">Leaderboard rating<br /> (Skill - 3 x Uncertainty)</th>
+              <th colspan="2">Balance rating<br /> (Skill - Uncertainty, non-negative)</th>
               <th>Date</th>
               <th>&nbsp;</th>
             </tr>
@@ -92,7 +91,6 @@
                   <td colspan="2"><%= log.value["reason"] || "No match" %></td>
                 <% end %>
                 <td><%= @rating_type_id_lookup[log.rating_type_id] %></td>
-
                 <td><%= round(log.value["skill"], 2) %></td>
                 <td class={text_class}>
                   <i class={"fa-fw fa-solid fa-#{icon}"}></i>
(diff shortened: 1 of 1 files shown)


Teifion:
It does look somewhat too noisy but we've also made a specific choice to not show exact uncertainty levels to most players. The progression graph now shows some and I'm happy with that level as it's very understandable. I don't want to encourage issues from "a little bit of knowledge is a dangerous thing" portion in this regard.

I'll leave the PR open as I'm happy to discuss it but my gut feeling is this will end up being rejected.

StanczakDominik:
May I ask why that specific choice was made? It's already something people are explicitly asking how to figure out. I was not aware of it, but okay - my primary goal here was to prevent the issue of [rating pages like this](https://cdn.discordapp.com/attachments/1115432854919327745/1116833556195463298/image.png) for balance ratings below 0 before the clipping to [0, infty) operation, which have confused people in the past - myself included! Would any of these following solutions be acceptable?

- Drop the uncertainty column and show the skill rating changes; use those to pick the color
- Drop both the uncertainty and skill columns and show the leaderboard rating changes instead of the current balance rating changes
- Revert to the current set of columns (keep the descriptions for Balance/Leaderboard so people can realize at a glance why their balance rating stays 0), show diffs for both Balance and Leaderboard. The latter is not in the database but it should be as simple as `skill_change + 3 * uncertainty_change`.

StanczakDominik:
image
How about this? :)

StanczakDominik:
This should be better. I previously mistakenly added a screenshot of my own current ratings on server4 😅 
image

lib/teiserver_web/templates/battle/match/ratings.html.heex
```diff
@@ -91,13 +91,17 @@
...
 
-                <td><%= round(log.value["skill"] - log.value["uncertainty"] * 3, 2) %></td>
+                <td><%= round(log.value["skill"] - 3 * log.value["uncertainty"], 2) %></td>
+                <td class={text_class}>
+                  <i class={"fa-fw fa-solid fa-#{icon}"}></i>
 
-                <td><%= round(log.value["rating_value"], 2) %></td>
+                  <%= round(log.value["skill_change"] - 3 * log.value["uncertainty_change"], 2) %>
+                </td>
+                <td><%= round(log.value["skill"] - log.value["uncertainty"], 2) %></td>
```
StanczakDominik, line 100:
```suggestion
                <td><%= round(log.value["rating_value"], 2) %></td>l
```

lib/teiserver_web/templates/battle/match/ratings.html.heex
```diff
@@ -91,13 +91,17 @@
...
+                  <i class={"fa-fw fa-solid fa-#{icon}"}></i>
 
+                  <%= round(log.value["skill_change"] - 3 * log.value["uncertainty_change"], 2) %>
+                </td>
                 <td><%= round(log.value["rating_value"], 2) %></td>
                 <td class={text_class}>
                   <i class={"fa-fw fa-solid fa-#{icon}"}></i>
 
-                  <%= round(log.value["rating_value_change"], 2) %>
+                  <%= round(log.value["skill_change"] - log.value["uncertainty_change"], 2) %>
```
StanczakDominik, line 104:
```suggestion
                  <%= round(log.value["rating_value_change"], 2) %>
```

StanczakDominik (commented):
Forgot to revert a thing.

Teifion:
> 1 - Drop the uncertainty column and show the skill rating changes; use those to pick the color

This would be completely acceptable to me :)

StanczakDominik:
image

This should now be ready for review :)

PR #150, opened 2023-07-31
Update battle_status on spectatorCount changes

FIr3baL:
- send UPDATEBATTLEINFO to all server users, added condition: changed_spectator_count
- avoid duplicate sending of UPDATEBATTLEINFO to battle users

### DIFF
--- a/lib/teiserver/tcp/spring/spring_tcp_server.ex
+++ b/lib/teiserver/tcp/spring/spring_tcp_server.ex
@@ -369,7 +369,7 @@ defmodule Teiserver.SpringTcpServer do
       else
         keys = Map.keys(new_values)
 
-        if Enum.member?(keys, :locked) or Enum.member?(keys, :map_name) do
+        if Enum.member?(keys, :locked) or Enum.member?(keys, :map_name) or Enum.member?(keys, :spectator_count) do
           SpringOut.reply(:update_battle, lobby_id, nil, state)
         else
           state
@@ -870,8 +870,7 @@ defmodule Teiserver.SpringTcpServer do
         SpringOut.reply(:enable_all_units, [], nil, state)
         SpringOut.reply(:disable_units, msg.changes.disabled_units, nil, state)
 
-      true ->
-        SpringOut.reply(:update_battle, msg.lobby_id, nil, state)
+      true -> state
     end
   end
 
--- a/lib/teiserver/tcp/spring/spring_tcp_server.ex
+++ b/lib/teiserver/tcp/spring/spring_tcp_server.ex
@@ -870,7 +870,8 @@ defmodule Teiserver.SpringTcpServer do
         SpringOut.reply(:enable_all_units, [], nil, state)
         SpringOut.reply(:disable_units, msg.changes.disabled_units, nil, state)
 
-      true -> state
+      true ->
+        SpringOut.reply(:update_battle, msg.lobby_id, nil, state)
     end
   end
 


lib/teiserver/tcp/spring/spring_tcp_server.ex
```diff
@@ -870,8 +870,7 @@ defmodule Teiserver.SpringTcpServer do
         SpringOut.reply(:enable_all_units, [], nil, state)
         SpringOut.reply(:disable_units, msg.changes.disabled_units, nil, state)
 
-      true ->
-        SpringOut.reply(:update_battle, msg.lobby_id, nil, state)
```
Teifion, line 874: Why do we not want to default to sending out an update_battle message?

FIr3baL: idk, that's why i removed it

Teifion: I'd like a reason to remove it before we remove it, currently that's the fallback of "if stuff has changed, send out an update_battle message to be safe" clause.

FIr3baL: 2nd option: don't change it for now, but just go for the first edit.
Clients will get 2x UPDATEBATTLESTATUS , if user is member of that battle. That's no problem for chobby.
I thought of saving some bandwith and avoid unnessary messages with it.

Otherwise i can help with finding a reason tomorrow or late this evening.

FIr3baL: for now i can say, that we need 
true ->
        SpringOut.reply(:update_battle, msg.lobby_id, nil, state)
otherwise new connected clients don't get UPDATEBATTLEINFO for already existing battles.
Need to find another way to avoid duplicate sending of UPDATEBATTLEINFO for members of a battle.

Teifion: How much help/guidance would you like from me in how to dig into why there's a duplicate message? I know part of the fun of this can be the learning process but I also don't want you to feel unsupported!

Teifion: So the updating of values for a battle would be updating almost any other value so I think it makes sense to have this there or a bunch of other updates won't propogate to clients. It might be worth looking at which changes come through twice because then we can work back from those and deal with them accordingly.

FIr3baL: teiserver_global_lobby_updates
This results in sending updates to every client, which is wanted so that clients can show the current values in a battle list without joining the battle. (just need to add ```or Enum.member?(keys, :spectator_count)``` as mentioned in this PR)

teiserver_lobby_updates
And this one is sending updates to clients only that are member of the updated battle.

With spring protocol the only chance for clients to know the number of players and spectators is:
1. Keep track of battle members count by incrementing/decrementing on JoinedBattle/LeftBattle
2. subtracting spectatorCount from it, when receiving UpdateBattleInfo to know playerCount

Currently teiserver does not send UpdateBattleInfo for updated SpectatorCount, but only for updates to map and locked. If you observer battle list in chobby you can easily see, that the player numbers fix themself as soon as the map was changed or you join that battle by your own.

And if you have joined a battle you will receive UpdateBattleInfo 2 times in a row on changes to map or locked, because you belong to teiserver_global_lobby_updates AND teiserver_lobby_updates.

I would consider changes to spectatorcount being the most frequent updates for a spring server by far. It changes on each Joned/LefBattle and each MyBattleStatus that changes from or to Spectator. Maybe that's why you disabled them in the past ? At least it can't be wrong to minimize the amount of sent UpdateBattleInfo's. It's less load for server and clients.

Option A:
Don't send UpdateBattleInfo for teiserver_lobby_updates

Option B:
Don't send UpdateBattleInfo for teiserver_global_lobby_updates to clients that are member of the updated battle.

FIr3baL:
@Teifion I rolled back the 2nd change, since this is just luxus.
Can we merge it like this first and look for optimizations later ?

lib/teiserver/tcp/spring/spring_tcp_server.ex
Teifion: Both channels should be sending their channel name as part of it, we can probably use that to work out which ones we do and don't want to do pass on. Option B makes the most sense though, it only needs to be a temporary fix for now :)

Teifion:
@FIr3baL sorry for the lag there, I basically had a day of no coding so didn't look at this. It's merged and I'll do a release with it on Monday.

PR #152, opened 2023-08-03
allow tweakdefs through SAYBATTLEEX length filter

metavolution:


metavolution:
not sure if indent is perfect / good enough on this one ...

PR #155, opened 2023-08-09
allow tweakdefs and welcome-message again

metavolution:


Teifion:
Instead of a nested if, please make use of the `cond` statement, there's a tutorial on it here - https://elixirschool.com/en/lessons/basics/control_structures#cond-2 - if that helps :)

PR #161, opened 2023-09-01
Adjust chat search 'When' field

StanczakDominik:
This forces chat search to also display times, not just dates. Sure it's more noise, but it beats binary-searching to figure out the lobby where an outdated, or even yesterday's, report happened.

(I'm having some issues with dependency setup right now, so I wasn't able to test this one, but it should be simple enough.)

### DIFF
--- a/lib/teiserver_web/templates/admin/chat/index.html.heex
+++ b/lib/teiserver_web/templates/admin/chat/index.html.heex
@@ -49,7 +49,7 @@
                       </a>
                     </td>
                     <td class="message"><%= msg.content %></td>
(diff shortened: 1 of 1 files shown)


Teifion:
It used to do this but I changed it to reduce the noise. Is there anything else we could do to make it easier to see when the match started maybe?

StanczakDominik:
I don't think that would help much - when I do a chat search, I mostly want to match up the report date from `https://server4.beyondallreason.info/moderation/report/<report_id>` (which has the full datetime regardless of when the report was made) to a chat entry, to figure out the exact moment someone felt compelled to report someone. Having these visible on the chat search, while I get it's a trade-off vs the noise, would still make it a fair bit easier.

PR #169, opened 2023-09-27
Fix profile_link leading to NotFound error

lolsteamroller:
remove a path for teiserver, as the production API throws Not Found.

image
image

and after investigation the path element teiserver is not needed.


StanczakDominik (approved):
Welp, I got distracted and forgot to push my branch with this exact change. :D

Teifion:
Hotfix deployed.

PR #183, opened 2023-10-22
Add type column to overwatch page

StanczakDominik:
Don't merge this yet, it blows up! I'm getting this error:

```
[error] an exception was raised:
** (Protocol.UndefinedError) protocol Enumerable not implemented for #Ecto.Association.NotLoaded<association :reports is not loaded> of type Ecto.Association.NotLoaded (a struct)
```

So I understand that I would need to pre-load reports for each
`report_group` in `lib/teiserver_web/live/moderation/overwatch/index.ex`
for this to work. I'm not sure how to do that, though... I gather that you're
pre-loading the report groups here:

```
    socket = socket
    |> assign(:site_menu_active, "moderation")
    |> assign(:view_colour, Teiserver.Moderation.colour())
    |> assign(:outstanding_report_groups, 0)
    |> assign(:report_groups, nil)                 <----- this?
    |> default_filters(params)
    |> add_breadcrumb(name: "Moderation", url: ~p"/moderation")
    |> add_breadcrumb(name: "Overwatch", url: ~p"/moderation/overwatch")
```

But I'm not sure how to extend that to also load something "belonging"
to those report groups.

### DIFF
--- a/lib/teiserver_web/live/moderation/overwatch/index.html.heex
+++ b/lib/teiserver_web/live/moderation/overwatch/index.html.heex
@@ -106,6 +106,9 @@
   </:col>
   
   <:col :let={report_group} label="Target"><%= report_group.target.name %></:col>
+  <:col :let={report_group} label="Type">
+    <%= Enum.join(MapSet.new(for report <- report_group.reports, do: report.name), ", ") %>
+  </:col>
   <:col :let={report_group} label="Reports"><%= report_group.report_count %></:col>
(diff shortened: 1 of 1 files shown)


Teifion:
You got the right place as to where the variable is set, that line you commented on though is a default value being set to nil. The `recalculate_outstanding_report_groups` function is used to pull the report groups and that's where the query exists. I know you want to solve as much yourself so I'll stop there but feel free to ask if this isn't enough, I'm happy to help :)

PR #190, opened 2023-11-03
whoIs response: syntax change

FIr3baL:
to have userID and userName at same position in server response:


FIr3baL:
hope the code works :)

PR #193, opened 2023-11-12
Meme Commands Update/Improvements

robertthepie:
**Randomizer Imrpovements:**
- Commanders and constructors can no longer end up without a factory for previous/current/next tier, nor without a metal extractor and energy generator. T1/T2 Factories should also always have a constructor.

**Hoversonly Improvements/Update**
- Balance: Commanders no longer copy hover con build lists - available turrets were worth more than building an army
- Prepare for commander update and hover rework being integrated from always on modoptions to intergrated into unit stats

**Others:**
- Update removed modoption values


Teifion:
Merging, it will go live with the next release.

Issue #178, opened 2023-10-15
Add ability to join user lobby from their profile page


Teifion:

darcymiranda:
@Teifion I opened a PR for this request https://github.com/beyond-all-reason/teiserver/pull/209. Please let me know you can help me test this.

Also, personally, I think this feature would also be nice to have on the party page too.

jauggy:
@StanczakDominik this can be closed.

PR #209, opened 2024-01-07
Add ability to join user lobby from their profile page

darcymiranda:
This is for https://github.com/beyond-all-reason/teiserver/issues/178.

I only tested this through tests and have been unable to test locally. When running locally, I get region errors trying to create a lobby. I've come to understand that it's because I don't have SPADS running, but I'm unable to get SPADS working locally.

Details
--
* Join button is only visible if the user is in a lobby

Screenshots
--
image

### DIFF
--- a/lib/teiserver_web/live/account/profile/overview.ex
+++ b/lib/teiserver_web/live/account/profile/overview.ex
@@ -290,11 +290,11 @@ defmodule TeiserverWeb.Account.ProfileLive.Overview do
     current_user_id = assigns.current_user.id
     lobby_id = assigns.client.lobby_id
 
-    with true <- client_connected(current_user_id),
-         {:ok, _} <- join_lobby(lobby_id, current_user_id) do
+    with :ok <- client_connected(current_user_id),
+         :ok <- server_allows_join(lobby_id, current_user_id),
+         :ok <- join_lobby(lobby_id, current_user_id) do
       {:noreply, put_flash(socket, :success, "Lobby joined")}
-    else
-      {:error, reason} ->
+    else {:error, reason} ->
         {:noreply, put_flash(socket, :warning, reason)}
     end
   end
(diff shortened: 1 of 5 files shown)


lib/teiserver_web/live/account/profile/overview.ex
```diff
@@ -250,6 +285,38 @@ defmodule TeiserverWeb.Account.ProfileLive.Overview do
...
+
+    if not is_nil(client) do
+      true
+    else
+      {:error, "Client is not connected"}
+    end
+  end
+
+  defp join_lobby(lobby_id, user_id) do
+    if :ok == Lobby.force_add_user_to_lobby(user_id, lobby_id) do
```
darcymiranda, line 313: Does this obey rules like user is banned from lobby and what not? I would have preferred to the use `add_user_to_battle`, but it has a `script_password` arg that I have no idea about

Teifion:
Hi @darcymiranda I've left this project (now implementing it as a library over at https://github.com/Teifion/teiserver) and instead I think @Beherith would be the better person to respond to this PR :)

Beherith:
Hit me up in DM on discord :)

lib/teiserver_web/live/account/profile/overview.ex
L-e-x-o-n: I don't think it does, use `Battle.server_allows_join?` to check for that first.

This is how it's done in other places.
```ex
    if Battle.server_allows_join?(userid, lobby_id) == true do
      Battle.force_add_user_to_lobby(userid, lobby_id)
    end
```

darcymiranda: Thanks for the suggestion - I hadn't noticed it being used like that elsewhere. I added it [58d0bd8](https://github.com/beyond-all-reason/teiserver/pull/209/commits/58d0bd82fabc7461acdcfbf52c9606632561f892)

darcymiranda:
> Hit me up in DM on discord, DarcyMiranda :)

I was able to get spads setup (Thanks @jauggy) and successfully tested this manually

PR #210, opened 2024-01-11
removed min limit (1) for shown skills to spads

FIr3baL:


FIr3baL:
this should be testet with webserver, too!
AS i learned now, values from spads go into demo files and those are used as origin for website values.

Beherith:
@FIr3baL scream if this can go in

jauggy:
Currently on the website ratings page 
```
https://server4.beyondallreason.info/profile/157348
```
The match rating will floor at 0.0 if your `skill-uncertainty` is negative. Similarly the database ratin value will floor at 0.0. $whoami rating stats also will floor at 0.0. 

But Chobby currently shows 1 as the OS floor. This PR would make things consistent with the true value.

FIr3baL:
@jauggy thanks for testing.
@Beherith imho, this can go in !

PR #212, opened 2024-01-28
Changed namespace from Teiserver to Barserver

Teifion:
Replaces all `Teiserver` and `TeiserverWeb` namespaces with `Barserver` and `BarserverWeb`. In the future it is possible the project may want to move to make use of the [Teiserver library](https://hex.pm/packages/teiserver) and thus preventing namespace collisions is an important first step. It will also make any integration PRs much easier to review as the find/replace step will have already been completed.

Lowercase `teiserver` entries cannot be find/replaced as easily since they refer to things like the application name, database tables and certain config key names. They will need to be addressed in a more careful manner.


Teifion:
Unfortunately a number of the tests are failing on my system though swapping back to master the same tests fail. I would highly recommend testing this branch on an integration system before putting it live even though it should not break anything.

PR #234, opened 2024-03-16
Fix local setup documentation for root user

darcymiranda:
At some point this broke because `get_user_by_email/1` is now returning a map (cached user) instead of the user ecto struct. Let's just use the general functions provided by Ecto to fetch the root user.


jauggy:
I tested this and can confirm that the docs are wrong and this update fixes it.

StanczakDominik (approved):

Issue #218, opened 2024-03-09
Add balancemode to evenly distribute one chevs


jauggy:
# Problem
<img src="https://github.com/beyond-all-reason/teiserver/assets/1305569/c3b9bb36-75ee-405e-9a30-85407ca30ec0" width="300">

Games like this are unfair and it feels like the match is decided by the balancer. If one side has multiple new players and the other side has none, then if you're a semi-experienced player on the side of the new players you will feel the game is unwinnable.

I made a suggestion in [discord and was told PR's welcome](https://discord.com/channels/549281623154229250/1206459571783073832/1206898868181008404). The suggestion was to allow a better distribution of one chevs across teams. This suggestion is not new. Core BAR developer Firestorm [suggested evenly distributing new players across teams back in 2021](https://discord.com/channels/549281623154229250/549281623577722899/908066499841908746)

# Proposed Solution
  The proposed solution doesn't mean changing the default balance algo but instead allowing another one that can be run if wanted. From the lobby, allow players to make this call (boss required)
```
$balancemode split_one_chevs
```
which will balance the lobby differently by assuring that one chevs will be distributed evenly across teams and drafts them last. There are already a couple of balance algos already (loser_picks is the default; cheeky_switcher_smart is a non-default algo).

## Algorithm Description

All players will be sorted by visible OS. Then all one chevs will be placed at the bottom of the list. The team with the lowest amount of players will pick first from this sorted player list. If there are multiple teams with the lowest player count, then the team with the lowest match rating picks.

Since new players are drafted last, it's similar to them being treated as having negative match rating. They will be seen as being worse than a 1 OS player.

## Players' opinions on the algo
Check out this [reddit comment](https://www.reddit.com/r/beyondallreason/comments/1bl2m6k/the_real_issue_with_new_players/kw2dvd6/) I made. Based on the upvotes it seems the balancemode would be received favourably.

## Considerations
We have an optional balancemode called `cheeky_switcher_smart` that never gets used. My opinion of this is because most people don't know how it works nor can explain how it works. It was therefore important to make my balance algorithm easy to understand to increase the chance it would get used.

## Testing the algo using past replays

 Go to my website [here](https://balance-algo-web.web.app/). Using the website you can try testing with this replay:
```
https://www.beyondallreason.info/replays?gameId=876fc5656af05709119ca358699032cf
```

Its main downside compared to the default algo is that it ignores parties

# Expected Result
[...]

Issue #233, opened 2024-03-16
Add documentation for SPADS installation


jauggy:
Here is draft:

## Purpose of SPADS installation
If you have a local running version of TeiServer, you can point Chobby to localhost and login. However, you won't be able to see any battle rooms. To see battle rooms you will need to install SPADS.

## Create a bot account to be used by SPADS
Run TeiServer locally and go to 
http://localhost:4000/teiserver/admin/user

Search for a user like SpryCeruleanTermite
Then Edit User > Tick "Bot" and "Verified"

## Point Chobby to local TeiServer
Open Beyond-All-Reason exe > Start > Settings > Scroll down and change Server Address to localhost

Go to Multiplayer & Coop > Login as the bot user e.g. SpryCeruleanTermite
All these users should have a password of `password`
Accept any agreement if needed. This is just to test if everything works. You can now Exit the app.

## Install SPADS Prerequisites
Download dependencies by following the prerequisite instructions [here](https://github.com/Yaribz/SPADS/blob/master/INSTALL.md). In theory if you have TeiServer on Windows WSL, you could use either the Linux or Windows prerequisite instructions.

Another dependency we need that is not listed is Inline::Python. This dependency is used by one of BAR's plugins. Here are the install steps for Ubuntu:
```
sudo apt-get install python3-dev 
cpan Inline::Python
```

## Install SPADS
Further down the previous page are [install instructions](https://github.com/Yaribz/SPADS/blob/master/INSTALL.md#installation-instructions) for SPADS. Follow only steps 1 and 2 of the Installation instructions. When downloading spadsInstaller.tar it might be flagged as a virus. (I had to disable Windows Threat Protection. Not sure if there's a better way).

Now for step 3 instead run this
```
perl spadsInstaller.pl --auto BarMinimalMaps
```
This will install SPADS with some BAR-specific configurations. It will ask you for a username/password for SPADS to use. Use the bot user you created earlier e.g.
Username: SpryCeruleanTermite
Password: password

Finally it will ask you for another login for autohost owner. You can enter anything here, even a user that doesn't exist. It's not used.

## Config changes

Inside your SPADS folder, open etc/BarAutoUpdate.conf and blank out the mapInventoryUrl
```
mapInventoryUrl:
```
Don't confuse this with the mapListsUrl which is different. The purpose of this is to prevent SPADS from downloading all maps (15GB) whenever you run it. Instead, you will have 3 maps to use.

[...]

PR #235, opened 2024-03-19
WIP: Add minimal files for split one chevs balancemode

jauggy:
To test this please see instructions in #218


jauggy:
Fixes #218

lib/teiserver/battle/libs/balance_lib.ex
```diff
@@ -374,13 +395,7 @@ defmodule Teiserver.Battle.BalanceLib do
 
           {team_id, players} ->
             top_player =
-              players
-              |> Enum.map(fn userid ->
-                {ratings[userid], userid}
```
jauggy, line 379: This is a bug hence I had to remove and replace with a different function for captains. ratings is a variable defined above as tuple of team ids and team ratings. 
```
ratings[userid]
```
is completely wrong.

lib/teiserver/libs/teiserver_configs.ex
```diff
@@ -572,7 +572,7 @@ defmodule Teiserver.TeiserverConfigs do
       key: "profile.Rank method",
       section: "Profiles",
       type: "select",
-      default: "Leaderboard rating",
+      default: "Role",
```
jauggy, line 575: At one point the chevs were calculated by leaderboard rating. Now they are calculated using time. Search teiserver codebase for
```
calculate_rank(userid, "Role")
``` 
and you'll find the current algo. This algo can be changed via the admin panel and will effect the database.
http://localhost:4000/teiserver/admin/site
Go to profile and you can set the way chevs are calculated.

jauggy:
@badosu this pull request has less changes than the previous pull request but should give the same results.

badosu:
Thank you! I'll check out when I'm able to

FIr3baL:
Tested with previous branch and it lgtm
https://github.com/jauggy/teiserver/tree/add-balancemode-split-one-chevs
default balance
old balance
$balancemode split_one_chevs
balancemode split_one_chevs

FIr3baL:
Same visual result with the branch of this PR
grafik

jauggy:
@badosu Given that fireball has tested, I think code review alone would be fine if you don't have a dev environment to test this.

However if you want to setup dev environment let me know if you have any issues with teiserver/spads setup. There's some docs here for [SPADS](https://github.com/beyond-all-reason/teiserver/issues/233).

Also there is not any documentation currently for launching multiple chobbies so let me know if you need that and I will create it.

badosu:
@jauggy Just to be clear, I'm not the stakeholder for this project, and don't know who that would be.

That said, I want to have a more active role, unfortunately haven't had time to familiarize with the repo myself yet. As soon as I do I'll get back to here. (There are no blockers from me until I can confidently review a PR here)

jauggy:
As requested by Lexon, I'll be adding some unit tests to this PR in the next couple of days.

jauggy:
@L-e-x-o-n Tests have been added. I updated the documentation `documents/guides/testing.md` which will give instructions on how to test just the balance tests.

lib/teiserver/battle/balance/split_one_chevs.ex
```diff
@@ -0,0 +1,205 @@
...
+        # Zipping will create binary tuples from 2 lists
+        {id, rating} <- Enum.zip(members, ratings),
+        # Create result value
+        rank = get_rank(id),
+        do: %{member_id: id, rating: rating, rank: rank}
+  end
+
+  def get_rank(member_id) do
+    Account.get_user_by_id(member_id).rank
+  end
```
geekingfrog, line 60: Hum, I think this is a problem. It does one database call per user, when it could do one for all members of the group at once using `Account.list_users`.

But, and I think this is a more severe problem, it introduce a database call in what used to be a pure function, and this creates issues in testing (see later).
I think a better way to handle that is to change the signature of `BalanceLib.create_balance` so that it takes a map of `user_id => {rating, rank}`

I had a look at the uses of `create_balance` and it's "only" used in 4 places so that shouldn't be too bad to modify, and this should remove the need to muck around with mocks in tests. That may require to adjust a bunch of other existing tests though :/

lib/teiserver/battle/balance/split_one_chevs.ex
```diff
@@ -0,0 +1,205 @@
...
+        do: %{members: [member_id], count: 1, group_rating: rating, ratings: [rating]}
+  end
+
+  def standardise_team_players(raw_input) do
+    Map.new(raw_input, fn x -> {x.team_id, standardise_member_ids(x.members)} end)
+  end
+
+  def standardise_member_ids(members) do
+    for %{member_id: member_id} <- members,
+        do: member_id
```
geekingfrog, line 203: This can be written as `Enum.map(members, fn member -> member[:member_id] end)`

lib/teiserver/battle/balance/split_one_chevs.ex
```diff
@@ -0,0 +1,205 @@
...
+  [
+                 %{rating: 6, rank: 0, member_id: 2},
+                 %{rating: 5, rank: 0, member_id: 4}
+               ]
+  output= [
+                 %{members: [2], count: 1, group_rating: 6, ratings: [6]},
+                 %{members: [4], count: 1, group_rating: 5, ratings: [5]}
+               ]
+  """
+  def standardise_members(members) do
```
geekingfrog, line 192: In general, prefer to use `defp` for functions that are purely internal to your module, makes easier to know what's safe to touch from what requires a bit more thought.

lib/teiserver/battle/balance/split_one_chevs.ex
```diff
@@ -0,0 +1,205 @@
...
+    end)
+  end
+
+  @spec create_empty_teams(any()) :: any()
+  def create_empty_teams(count) do
+    for i <- 1..count,
+        do: %{team_id: i, members: []}
+  end
+
+  @spec get_picking_team(any()) :: any()
```
geekingfrog, line 117: might as well remove the `@spec` if you're typing everything with `any`

lib/teiserver/battle/balance/split_one_chevs.ex
```diff
@@ -0,0 +1,205 @@
...
+
+  Returns %{teams:teams, logs:logs}
+  """
+  def assign_teams(member_list, number_of_teams) do
+    default_acc = %{teams: create_empty_teams(number_of_teams), logs: ["Begin split_one_chevs balance"]}
+
+    Enum.reduce(member_list, default_acc, fn x, acc ->
+      picking_team = get_picking_team(acc.teams)
+      update_picking_team = Map.merge(picking_team, %{members: [x | picking_team.members]})
+      username = Account.get_username_by_id(x.member_id)
```
geekingfrog, line 101: Another DB call that should be provided upfront.

lib/teiserver/battle/balance/split_one_chevs.ex
```diff
@@ -0,0 +1,205 @@
...
+      else
+        # Team is picker if it is tied for least and has lower team rating
+        if(
+          length(x.members) == length(acc.members) && get_team_rating(x) < get_team_rating(acc)
+        ) do
+          x
+        else
+          acc
+        end
+      end
```
geekingfrog, line 134: I believe it's more idiomatic in elixir to use [`cond`](https://hexdocs.pm/elixir/case-cond-and-if.html) over these nested ifs.

geekingfrog:
I think the most important change here is the introduction of DB calls inside the `create_balance` function. Seeing there is a timer `System.system_time(:microsecond)` as well, that makes me think this function is somewhat time sensitive and I'm wary of making it not pure by introducing DB calls inside.

jauggy:
Thanks for the review @geekingfrog . Yeah you caught something that I also wasn't too happy about but at the time I wanted to write minimal code. Now that I have more spare time, I'll see if I can pull the db call outside.

test/teiserver/battle/loser_picks_balance_test.exs
```diff
@@ -361,7 +366,8 @@ defmodule Teiserver.Battle.LoserPicksBalanceTest do
       )
 
     assert Map.drop(result, [:logs, :time_taken]) == %{
-             captains: %{1 => 116, 2 => 115},
+      #The captain should be the user with the highest rating on each team
```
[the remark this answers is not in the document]
jauggy: I don't know but my guess is that captains are used for blue/red colours. I have noticed that sometimes the colours are wrong in production.

jauggy:
@geekingfrog I have made a [new PR](https://github.com/beyond-all-reason/teiserver/pull/248)

lib/teiserver/battle/balance/loser_picks.ex
```diff
@@ -36,6 +36,10 @@ defmodule Teiserver.Battle.Balance.LoserPicks do
   """
   @spec perform([BT.expanded_group_or_pair()], non_neg_integer(), list()) :: BT.algorithm_result()
   def perform(raw_groups, team_count, opts) do
+    # This module doesn't use names or ranks and they need to be dropped from raw_groups to make existing tests pass
+    # However, this module could be improved to use names from raw_groups instead of calling Teiserver.Account
+    raw_groups = Enum.map(raw_groups, fn x-> Map.drop(x,[:ranks, :names]) end)
+
```
jauggy, line 42: If you can think of a way to use the names in the input rather than calling Account module that would help. I couldn't quite understand this module.

lib/teiserver/battle/libs/balance_lib.ex
```diff
@@ -94,11 +106,31 @@ defmodule Teiserver.Battle.BalanceLib do
...
         userids = Map.keys(members)
-        ratings = Map.values(members)
+        ratings = Map.values(members) |> Enum.map(fn x -> x.rating end)
+
+        ranks =
+          Map.values(members)
+          |> Enum.map(fn x ->
+            cond do
+              Map.has_key?(x, :rank) -> x.rank
+              true -> 0
```
geekingfrog, line 116: You can use a shorter version: `Map.get(x, :rank, 0)`

lib/teiserver/battle/libs/balance_lib.ex
```diff
@@ -94,11 +106,31 @@ defmodule Teiserver.Battle.BalanceLib do
...
+            end
+          end)
+
+        names =
+          members
+          |> Enum.map(fn {id, details} ->
+            cond do
+              Map.has_key?(details, :name) -> details.name
+              true -> "#{id}"
```
geekingfrog, line 125: Likewise `Map.get` supports a default value

lib/teiserver/battle/libs/balance_lib.ex
```diff
@@ -130,24 +162,43 @@ defmodule Teiserver.Battle.BalanceLib do
...
-      end)
+      cond do
+        Map.has_key?(balance_result, :team_groups) ->
+          balance_result.team_groups
+
+        true ->
+          balance_result.teams
+          |> Map.new(fn {team_id, groups} ->
+            {team_id, Enum.reverse(clean_groups((groups)))}
+          end)
```
geekingfrog, line 183: A more idiomatic way of checking that is to use pattern matching:
```elixir
case balance_result[:team_groups] do
  groups -> groups
  nil -> # do stuff
end
```

lib/teiserver/game/servers/balancer_server.ex
```diff
@@ -213,17 +213,14 @@ defmodule Teiserver.Game.BalancerServer do
           |> Enum.map(fn userid ->
             %{
               userid =>
-                BalanceLib.get_user_balance_rating_value(userid, rating_type)
-                |> fuzz_rating(opts[:fuzz_multiplier])
+                BalanceLib.get_user_rating_rank(userid, rating_type, opts[:fuzz_multiplier])
             }
```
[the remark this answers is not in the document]
geekingfrog: ah yes indeed, you're correct, it's not great. But I think this change is big enough already so let's leave it this way. Perhaps add a TODO mentioning that it would be better to grab everything in as few query as possible?

test/teiserver/battle/balance_lib_test.exs
```diff
@@ -1,6 +1,10 @@
 defmodule Teiserver.Battle.BalanceLibTest do
-  @moduledoc false
+  @moduledoc """
+  Can run tests in this file only by
+  mix test test/teiserver/battle/balance_lib_test.exs
+  """
```
geekingfrog, line 5: I wouldn't put this doc here as it's more something for `mix test` documentation.

geekingfrog (commented):
Some very minor details, more about using elixir than the actual logic itself.
If someone could test with a few local chobby and check that it does work I'd say it's good to be merged.

lib/teiserver/battle/balance/split_one_chevs.ex
```diff
@@ -0,0 +1,135 @@
...
+
+  defp get_picking_team(teams) do
+    default_picking_team = Enum.at(teams, 0)
+
+    Enum.reduce(teams, default_picking_team, fn x, acc ->
+      cond do
+        # Team is picker if it has least members
+        length(x.members) < length(acc.members) -> x
+        # Team is picker if it is tied for least and has lower team rating
+        length(x.members) == length(acc.members) && get_team_rating(x) < get_team_rating(acc) -> x
```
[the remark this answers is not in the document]
jauggy: Not sure I understand. Isn't members a list? I will do some research.

lib/teiserver/battle/balance/split_one_chevs_types.ex
```diff
@@ -0,0 +1,13 @@
...
+  @type member :: %{
+          rating: float(),
+          rank: non_neg_integer(),
+          member_id: any()
+        }
+  @type team :: %{
+    members: [member],
+    team_id: integer()
+  }
+end
```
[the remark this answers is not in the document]
jauggy: You can also see I've created a ticket in teiserver about types. Currently running dialyzer generates tons of errors.

lib/teiserver/battle/libs/balance_lib.ex
jauggy: Nice tip. Will use this one in the future.

lib/teiserver/battle/libs/balance_lib.ex
jauggy: Noted. Is there a difference between 
```
balance_result[:team_groups]
```
and
```
balance_result.team_groups
```
Can I use either?

test/teiserver/battle/balance_lib_test.exs
jauggy: I did this as a way to have something to copy and paste rather than typing it out by hand each time. Would it be acceptable to put as a `#` comment?

lib/teiserver/battle/libs/balance_lib.ex
```diff
@@ -162,6 +162,13 @@ defmodule Teiserver.Battle.BalanceLib do
     )
   end
 
+  # Only take keys we need
+  defp clean_groups(groups) do
+    groups |> Enum.map(fn x->
```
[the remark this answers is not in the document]
geekingfrog: Semantically it's the same. Performance wise, I'm not sure. I had a look at the generated beam asm for the following:

```elixir
defmodule DumpBytecode do
  def without_pipe(coll) do
    Enum.map(coll, &IO.inspect/1)
  end

  def with_pipe(coll) do
    coll |> Enum.map(&IO.inspect/1)
  end
end
```

Then, using [this project](https://github.com/michalmuskala/decompile) to decompile the bytecode:
```
elixirc test.ex
mix decompile DumpBytecode --to asm
```
This gives the file `Elixir.DumpBytecode.S`. This is the symbolic representation of the beam bytecode (assembly).
With a quick glance [at some doc](http://blog.stenmans.org/theBeamBook/#_beam_code_listings) about this format, the relevant bits are:

```
{function, with_pipe, 1, 11}.
  {label,10}.
    {line,[{location,"bytecode.ex",6}]}.
    {func_info,{atom,'Elixir.DumpBytecode'},{atom,with_pipe},1}.
  {label,11}.
    {move,{literal,fun 'Elixir.IO':inspect/1},{x,1}}.
    {line,[{location,"bytecode.ex",7}]}.
    {call_ext_only,2,{extfunc,'Elixir.Enum',map,2}}.


{function, without_pipe, 1, 13}.
  {label,12}.
    {line,[{location,"bytecode.ex",2}]}.
    {func_info,{atom,'Elixir.DumpBytecode'},{atom,without_pipe},1}.
  {label,13}.
    {move,{literal,fun 'Elixir.IO':inspect/1},{x,1}}.
    {line,[{location,"bytecode.ex",3}]}.
    {call_ext_only,2,{extfunc,'Elixir.Enum',map,2}}.
```

seems they compile to exactly the same.
Maybe more complex expression would result in different bytecode, but I don't think we should worry about that.

lib/teiserver/battle/balance/split_one_chevs.ex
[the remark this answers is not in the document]
geekingfrog: I'm not sure it's a list tbh. If you use `size` instead of `length` and it blows up (exception) you'll get your answer :D

jauggy:
@geekingfrog if you do end up getting multi chobby working, I have some sample instructions for testing in the [issue](https://github.com/beyond-all-reason/teiserver/issues/218#issue-2177058935) as well as a crude way of adding sample users.

geekingfrog:
After some unrelated struggles, I finally tested this branch as of 6422ac12d9b95597d9c59b6d1e30b61a208521df
and it looks good.

Using the default balance algorithm:
#
balance-normal

And then, switching using `$balancemode split_one_chevs` I get:
balance-new

I think it's probably good to be shipped.

PR #240, opened 2024-03-27
WIP: Fix cannot start match in dev environment when BAR in Windows and SPADS in WSL

### LINKED ISSUE #239: Cannot start match if SPADS is in WSL and BAR is in Windows (dev environment)
Let's say you have set up your dev environment as follows:
- TeiServer and SPADS on WSL
- BAR app in Windows

When you try and start a match, Chobby tries to connect to SPADS directly using 127.0.0.1. But it will fail. It can fixed if you get Chobby to connect to the WSL ip address.

jauggy:
### Context
Fixes #239 
If you've set up your dev environment as follows:
- BAR app in Windows
- Teiserver and SPADS in WSL

You will not be able to start a match

### Test Instruction
Inside config/dev.exs add your wsl ip address to this line:
```
  spads_ip_override: "Enter ip here"
```
You can find your wsl ip address via command prompt/powershell:
```
wsl hostname -I
```

Launch Teiserver and SPADS. Now launch BAR app. Try starting a match and it will work.

### DIFF
--- a/config/dev.exs
+++ b/config/dev.exs
@@ -63,7 +63,9 @@ config :teiserver, Teiserver,
   heartbeat_timeout: nil,
   enable_discord_bridge: false,
   enable_hailstorm: true,
-  accept_all_emails: true
+  accept_all_emails: true,
+  # Enter WSL ip address here if BAR is in Windows and SPADS is in WSL. Otherwise leave nil.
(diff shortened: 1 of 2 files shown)


jauggy:
This is not needed. Will update docs

PR #245, opened 2024-04-06
Moderation-adjacent fixes

StanczakDominik:
Pretty much just website-facing fixes. I tested all of this locally, can test on integration tomorrow, but I don't expect it to be at all noticeable game-side.

The TODO in e4753d86ad23e796cf9ccaaec7fe7e30a586ef97 I tested locally, the restored new moderation user profile page is visible for overwatchers.


lib/teiserver_web/templates/admin/user/tab_overview.html.heex
```diff
@@ -107,7 +107,15 @@
   <%= central_component("menu_card",
     icon: "fa-solid #{Teiserver.Moderation.ReportLib.icon()}",
     name: "moderation",
-    text: "Moderation",
+    text: "Reports",
+    icon_class: "fa-solid",
+    link: ~p"/moderation/report/user/#{@user.id}"
```
geekingfrog, line 112: You don't need to specify `user.id`, there's this [protocol param](https://hexdocs.pm/phoenix/Phoenix.Param.html) that will automatically take care of that for you. `~p"/moderation/report/user/#{@user}"` would do.

lib/teiserver_web/templates/moderation/report/index.html.heex
```diff
@@ -85,7 +77,7 @@
 
             <td>
               <a
-                href={Routes.profile_overview_path(@conn, :overview, report.target_id)}
+                href={Routes.moderation_report_path(@conn, :user, report.target_id)}
```
geekingfrog, line 80: Similar, I think using the `~p` sigil is preferred, because you get compilation warnings (errors?) if they don't match anything:
```
              <a href={~p"/moderation/report/user/#{report.reporter_id}"}>
```

lib/teiserver_web/templates/admin/user/tab_overview.html.heex
StanczakDominik: Didn't know, that's pretty cool, thank you!

lib/teiserver_web/templates/admin/user/smurf_list.html.heex
```diff
@@ -346,27 +346,30 @@ is_moderator = allow?(@conn, "Moderator") %>
...
-                  >
-                    <input type="hidden" name="_csrf_token" value={get_csrf_token()} />
-                    <td>
-                      <.input type="select" options={Map.keys(type_icons)} name="key" value="" />
-                    </td>
 
-                    <td>
-                      <.input type="text" name="value" value="" />
-                    </td>
+                <%= if allow?(@conn, "Server") do %>
```
geekingfrog, line 350: Tiny nitpick. At the top of the file there is already `is_moderator = allow?(@conn, "Moderator")`. You could do the same and reuse the value.
Although there are *also* other uses of `allow?(@conn, "Moderator")` elsewhere, so it's not super consistent to begin with.
I don't think there's a big impact either way since I believe the db calls are cached.

geekingfrog (commented):
For what it's worth, that looks alright to me.

PR #250, opened 2024-04-11
Better type support

### LINKED ISSUE #247: Add better type support
If we run `mix dialyzer` it will do type checking on the project and we will get MANY errors. Because of these errors it's difficult to add types and do type checking. However, following the instructions [here ](https://fly.io/phoenix-files/adding-dialyzer-without-the-pain/) we can run a command to create an ignore file. Basically that ignore file will be populated by all the current type errors. Then when we run the dialyzer command again, it will check this file for existing errors and ignore existing errors. It will only show new errors.

jauggy:
Fixes #247 

See issue for context. Adds ignore file so that you can run dialyzer without a million warnings.

## Test Steps
1. Change to master branch. Run `mix dialyzer` you will get a million errors and wall of red text
2. Change to this branch. Run `mix dialyzer`. You will get many errors skipped
3. Open 
```
teiserver/lib/teiserver/battle/balance/balance_types.ex
```
Delete the line 
```
          members: [T.userid()],
```

Now re run
```
mix dialyzer
```

You will be notified of a single type error


L-e-x-o-n:
It seems dialyzer isn't very reliable or accurate in considering the context. I noticed it in VS code as well (which displays these errors/warnings), it complains a about many things, function calls failing all over the place even when they pass and teiserver continuing to work without errors. I initially thought that some of these should be fixed instead of ignored, however, after looking at examples and considering the number of warnings/errors I realised it would take a lot of effort to go through all of them and find any actionable ones.
This is where this PR helps, by ignoring existing errors it might make dialyzer more useful in the future, making it easier to find potential future warnings.

L-e-x-o-n (approved):

PR #252, opened 2024-04-12
Render the section to edit user stats

geekingfrog:
Nothing was actually being rendered because of a typo.
Now we have this bit:
2024-04-12-616x169-scrot

### DIFF
--- a/lib/teiserver_web/templates/admin/user/tab_details.html.heex
+++ b/lib/teiserver_web/templates/admin/user/tab_details.html.heex
@@ -125,7 +125,7 @@
         </tbody>
       </table>
 
-      <% if allow?(@conn, "Server") do %>
+      <%= if allow?(@conn, "Server") do %>
       <hr />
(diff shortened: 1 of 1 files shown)


jauggy:
I tested with key `player_minutes`
and value `999`
and it goes to error page. 
```
ArithmeticError at GET /teiserver/admin/user/44
bad argument in arithmetic expression
```

geekingfrog:
Alright, I know why. Because through this interface, the only thing that goes through are string values. However the underlying database field is `jsonb`. That means this interface in this form isn't that useful since we cannot set anything but string values.

One quick fix, if not very ergonomic is to let people input arbitrary json values in this field, and the server then parse it before updating the record.

Beherith:
> Alright, I know why. Because through this interface, the only thing that goes through are string values. However the underlying database field is `jsonb`. That means this interface in this form isn't that useful since we cannot set anything but string values.
> 
> One quick fix, if not very ergonomic is to let people input arbitrary json values in this field, and the server then parse it before updating the record.

Indeed, the key-value thing is strings only!

PR #257, opened 2024-04-13
Can scroll big tables on narrow screens

geekingfrog:
A tiny fix, where I cannot scroll left-right.
scroll


Beherith:
Nice! Thank you :)

PR #258, opened 2024-04-14
Add guide for setting up on wsl

Jordan-Cottle:
## Summary

As part of getting introduced to this project, I set up teiserver on Ubuntu 22.0.4 from scratch. I did not have WSL set up on PC before, so this guide should cover everything needed from complete 0 to a working teiserver.

## Test Evidence

Server running locally in WSL:
image

Webpage after logging in with `root@localhost`
image


documents/guides/local_setup_ubunutu_wsl_from_scratch.md
```diff
@@ -0,0 +1,122 @@
...
+```
+
+Install postgres:
+```bash
+sudo apt install postgresql
+```
+
+Set password for postgres user (account should have been created automatically by the apt install)
+```bash
+sudo passwd postgres
```
jauggy, line 52: What happens if you skip this step? Is the default user/password just postgres postgres?

Jordan-Cottle: I'm not sure what the default password is. None of the few that I guessed were correct.

jauggy:
So in this document you created you already have lots of stuff that's already in the local_setup.md so I recommend we just update the original one but add some spoiler sections that you can fill in. See below for example. Update the below example with any extra comments and rearrange any sections if needed.

jauggy:
This is still a work in progress but loosely the sequence of steps you will need to take to setup a Teiserver is:

## Local/Dev
### Clone repo
```bash
git clone git@github.com:beyond-all-reason/teiserver.git
cd teiserver
```

### Windows-Specific Instructions
<details>
  <summary>Install WSL</summary>
  
 Add links to WSL installation and any other comments here
  
</details>

### Install Elixir and Erlang
You will need to install:
- [Elixir/Erlang installed](https://elixir-lang.org/install.html).

You must use the same versions defined in [.tools-versions](.tools-versions) or you will get errors. 


<details>
  <summary>Install Elixir and Erlang using asdf (Recommended)</summary>
 
 Add your asdf instructions here
</details>

### Install build tools (gcc, g++, make)
#### Ubuntu/Debian
```bash
sudo apt update
sudo apt install build-essential
```

### Compile dependencies
```bash
mix deps.get
mix deps.compile
```

### Postgres setup
<details>
  <summary>Install Postgres</summary>
 Add your postgres setup instructions here
</details>


#### Install database

Set up `teiserver_dev` and `teiserver_test` accounts:
```bash
sudo su postgres
psql postgres postgres <<EOF
CREATE USER teiserver_dev WITH PASSWORD '123456789';
CREATE DATABASE teiserver_dev;
GRANT ALL PRIVILEGES ON DATABASE teiserver_dev to teiserver_dev;
ALTER USER teiserver_dev WITH SUPERUSER;
CREATE USER teiserver_test WITH PASSWORD '123456789';
CREATE DATABASE teiserver_test;
GRANT ALL PRIVILEGES ON DATABASE teiserver_test to teiserver_test;
ALTER USER teiserver_test WITH SUPERUSER;
EOF
exit
```

Create the database schema for the teiserver application:
```bash
mix ecto.create
```



#### Localhost certs
To run the TLS server locally you will also need to create localhost certificates in `priv/certs` using the following commands

```bash
mkdir -p priv/certs
cd priv/certs
openssl dhparam -out dh-params.pem 2048
openssl req -x509 -out localhost.crt -keyout localhost.key \
  -newkey rsa:2048 -nodes -sha256 \
  -subj '/CN=localhost' -extensions EXT -config <( \
   printf "[dn]\nCN=localhost\n[req]\ndistinguished_name = dn\n[EXT]\nsubjectAltName=DNS:localhost\nkeyUsage=digitalSignature\nextendedKeyUsage=serverAuth")
cd ../..
```

### SASS
We use sass for our css generation and you'll need to run this to get it started.
```bash
mix sass.install
```

### Fake data
Running this:
```bash
mix teiserver.fakedata
```

Will generate a large amount of fake data and setup a root account for you. The account will have full access to everything and the database will be populated with false data as if generated over a period of time to make development and testing easier.

### Running it
Standard mode
```bash
mix phx.server
```

Interactive REPL mode
```
iex -S mix phx.server
```
If all goes to plan you should be able to access your site locally at [http://localhost:4000/](http://localhost:4000/).

### Libraries you need to get yourself
The site makes liberal use of [FontAwesome](https://fontawesome.com/) so if you are using the site you'll need to download it and do the following
```bash
fontawesome/css/all.css -> priv/static/css/fontawesome.css
fontawesome/webfonts -> priv/static/webfonts
```

If you want to use the blog you will also need to place [ace.js](https://ace.c9.io/) folder in the same place.

### Configuration
Most of the configuration takes place in [config/config.exs](config/config.exs) with the other config files overriding for specific environments. The first block of `config.exs` contains a bunch of keys and values, which you can update.

### Connecting to the spring party of your server locally
```bash
telnet localhost 8200
openssl s_client -connect localhost:8201
```

### config/dev.secret.exs
If you want to do things like have a discord bot in development you don't want these details going into git. It is advisable to create a file `config/dev.secret.exs` where you can put these config details. I would suggest a file like so:
```elixir
import Config

config :teiserver, Teiserver,
  enable_discord_bridge: true

config :teiserver, DiscordBridgeBot,
  token: "------",
  bot_name: "Teiserver Bridge DEV",
  bridges: [
    {"---", "main"},
    {"---", "promote"},
    {"---", "moderation-reports"},
    {"---", "moderation-actions"}
  ]

# Comment the below block to enable background jobs to take place locally
config :teiserver, Oban,
  queues: false,
  crontab: false

```

### Resetting your user password
When running locally it's likely you won't want to connect the server to an email account, as such password resets need to be done a little differently.

Run your server with `iex -S mix phx.server` and then once it has started up use the following code to update your password.

```elixir
user = Teiserver.Repo.get_by(Teiserver.Account.User, email: "root@localhost")
Teiserver.Account.update_user(user, %{"password" => "password"})
```

### Main 3rd party dependencies
The main dependencies of the project are:
- [Phoenix framework](https://www.phoenixframework.org/), a web framework with a role similar to Django or Rails.
- [Ecto](https://github.com/elixir-ecto/ecto), database ORM
- [Ranch](https://github.com/ninenines/ranch), a tcp server
- [Oban](https://github.com/sorentwo/oban), a backend job processing framework.

documents/guides/local_setup_ubunutu_wsl_from_scratch.md
L-e-x-o-n: Changing postgres password that way is not recommended, see https://serverfault.com/a/325596 for more details.

jauggy: > I'm not sure what the default password is. None of the few that I guessed were correct.

The reason I asked the question originally is because based on my own notes I didn't have to do this step. You can login to the db using `teiserver_dev WITH PASSWORD '123456789'`. So not sure what setting the postgres password was trying to accomplish.

Jordan-Cottle:
One reason I wanted a separate document is that this guide is very specific to ubuntu 22.0.4 (and with wsl, dunno if that changed anything though) while the other guide is generic and doesn't specify what OS to use, but is lacking specific instructions because of it.

I also wasn't wanting to disrupt the existing docs, but if we want to update that doc instead I can take a look at doing that

documents/guides/local_setup_ubunutu_wsl_from_scratch.md
jauggy: > I'm not sure what the default password is. None of the few that I guessed were correct.

> I'm not sure what the default password is. None of the few that I guessed were correct.

The reason I asked the question originally is because based on my own notes I didn't have to do this step. You can login to the db using `teiserver_dev WITH PASSWORD '123456789'`. So not sure what setting the postgres password was trying to accomplish.

Jordan-Cottle: Thanks for pointing this out. I don't think I ever actually needed to log in to the postgres user account directly, so leaving this off and making it to the user can't be logged into makes sense.

Although, since this is specific to a WSL setup, the risks of having the user account have a password and available for login is much less relevant than it would be for a live server or cloud instance.

jauggy:
People actually tried to setup Teiserver on pure Windows and ran into issues. So my recommendation for all future devs on Windows is to use Linux on Windows.

StanczakDominik (commented):
This is good stuff but I'm afraid it would diverge from local-setup.md if anything changes in the future. Think you could add references between the two so that someone interested can git log to figure out anything that changed in one and not the other?

Issue #255, opened 2024-04-12
Players with negative OS don't see changes as they win or lose games


jauggy:
See here https://discord.com/channels/549281623154229250/1228467265096384693

The match rating of that player is likely negative but they only see 0.0 on the website causing them to think the rating system is bugged. Similarly when they run $whoami they only see 0.

An improvement would be to show in $whoami what their skill and uncertainty is and some sort of message that it is using 0 instead.

For example:
"Your match rating would normally be skill - uncertainty (0-1.23), but since this is negative, your match rating (OS) will be 0 instead."

StanczakDominik:
Do you think we can do something about this chobby-side? If you query Teiserver and get match rating below 0, query the actual skill rating + uncertainty tuple for display?

Also quoting discord: *this then opens up the potential to systematically direct them to asking for feedback and checking the resources available in ⁠🎓｜academy-chat*

jauggy:
The simplest solution is just using negative match rating values and showing negative match rating values in Chobby. The screenshot I linked [here](https://discord.com/channels/549281623154229250/855772061095559179/1247315203469869100) is an endorsement of showing negative values as it gives teammates more accurate information and allows them to act appropriately.

jauggy:
Just adding relevant messages here from PowerfulPizza who had a significant negative rating (also linked above and I've asked their permission to share it). The assumption that negative rating would hurt someone's feelings is an assumption that is not necessarily correct.

![2](https://github.com/beyond-all-reason/teiserver/assets/1305569/e347ad40-26c0-4a38-bf24-024c1cd4fdce)
![1](https://github.com/beyond-all-reason/teiserver/assets/1305569/be6c9ded-5c90-45ce-be66-70551abf6528)

jauggy:
Another negative player who would be happy to show negative as it allows to differentiate between 0 OS players and allows number to grind: https://discord.com/channels/549281623154229250/549281623577722899/1260363051828379730

NortySpock:
Planning on exploring a test and fix for this in the next week or so

PR #261, opened 2024-04-14
Display uncertainy values for each match

Jordan-Cottle:
## Summary

This PR adds in the `uncertainty` and `uncertainty_change` values to the match rating table so that players can know what their uncertainty was and how much each match impacted their uncertainty values.

I've chosen to not highlight the change values red/green because uncertainty changes aren't inherently good/bad, and the value should only really go down over time.

relates to #255 (and might close it if this is deemed sufficient to resolve the confusion)

## Test Evidence

image

### DIFF
--- a/lib/teiserver_web/live/battles/match/ratings.html.heex
+++ b/lib/teiserver_web/live/battles/match/ratings.html.heex
@@ -77,14 +77,14 @@
         <%= for log <- @logs do %>
           <% {skill_text_class, skill_icon} =
             cond do
-              log.value["skill_change"] > 0 -> {"text-success", "up"}
-              log.value["skill_change"] < 0 -> {"text-danger", "down"}
+              log.value["skill_change"] > 0 -> {"text-success", "arrow-up"}
+              log.value["skill_change"] < 0 -> {"text-danger", "arrow-down"}
               true -> {"text-warning", "pause"}
             end %>
           <% {uncertainty_text_class, uncertainty_icon} =
             cond do
-              log.value["uncertainty_change"] > 0 -> {"text-success", "up"}
-              log.value["uncertainty_change"] < 0 -> {"text-danger", "down"}
+              log.value["uncertainty_change"] > 0 -> {"text-success", "arrow-up"}
+              log.value["uncertainty_change"] < 0 -> {"text-danger", "arrow-down"}
               true -> {"text-warning", "pause"}
             end %>
           <tr phx-click={"[[\"navigate\",{\"href\":\"/battle/#{log.match_id}\",\"replace\":false}]]"}>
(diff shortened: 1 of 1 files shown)


jauggy:
@Jordan-Cottle sorry I updated the ticket as you were working on it. Can you do it as a hover over tooltip? Show people the `skill-uncertainty`

Keep the columns the same. The main purpose is for people who have match rating 0.0 so it's not useful information except for those people.

jauggy:
The idea is so that this person here could hover over their 0.0 number and see what the skill-uncertainty is. Not sure exactly would the tooltip should say. It could show 3 numbers or just 1. I'll need to see.

Jordan-Cottle:
I'd need to figure out how to do a tooltip, but I'm guessing fontawesome might have a css class for it already?

Happy to make the change to a tooltip if needed, but I actually kind of like the additional column. 

I use this page myself a lot, and often felt like it didn't answer the "What is my uncertainty though?" question very well. So I know I'd personally be interested in having the uncertainty for each match displayed here and I'm not a player with a negative match rating.

Jordan-Cottle:
Also, I don't think negative match rating is a thing. It looks like everywhere match rating is calculated teiserver is pinning the min at 0.

Example: https://github.com/beyond-all-reason/teiserver/blob/324ba2b0a0e50c82feb3f86eeffbd7db626a2d37/lib/teiserver/battle/libs/balance_lib.ex#L505

jauggy:
Negative match rating is not a thing. So the tooltip is showing `skill-uncertainty`. Not their match rating. It's useful so that someone can see this value changing rather than staying at 0.

For the uncertainty column, if you really want it, can you change the diff column to be colour coded like the others?

Jordan-Cottle:
Yup, I had the diff column color coded at first. But since up/down of uncertainty isn't obviously good (green) or bad (red) I thought the colors might be misleading. If I make down red, then players might think they're doing something wrong if their uncertainty keeps decreasing.

Also, if uncertainty never goes up due to a match, the other color would never show up and the red color would effectively just be an additional highlight to distract from the more important numbers.

What if I just add the up/down icon and not the color coding?

jauggy:
Can I see what it looks like colored and with arrows? For me the colors just let me know that it's a delta not an absolute value. But maybe if you show me what it looks like I'll have a better idea.

Also for some reason the arrows disappeared on the other columns. If it's due to FontAwesome, try and find a free arrow since Marek wanted us to not depend on paid version.

Jordan-Cottle:
Sure!

I've added in the color code and arrows to the uncertainty change column. I also updated the icon to use a free version of the up/down arrows.

image


If you like it, I'll update the test evidence in the initial PR description to show this version

Jordan-Cottle:
And this is what it looks like with the arrows, but not colored

image

PR #263, opened 2024-04-14
Replace Parallel with ParallelStream

L-e-x-o-n:
The original Parallel has not been updated since 2015 and is causing an error during compilation on newer Elixir versions.
This PR replaces Parallel with ParallelStream, a newer version with same functionality that compiles without errors on newer Elixir versions.

I recommend running 
`mix deps.clean --unused` or `mix deps.clean parallel`
`mix deps.get`
`mix deps.compile`
after merge or for testing

### DIFF
--- a/lib/teiserver/account/libs/accolade_lib.ex
+++ b/lib/teiserver/account/libs/accolade_lib.ex
@@ -384,10 +384,10 @@ defmodule Teiserver.Account.AccoladeLib do
 
           pings =
             children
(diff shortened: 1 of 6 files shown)


joshua-revolutions:
So in theory you could update .tool-versions ? Is that how you tested this?

Beherith:
The dockerfile also references elixir 1.14.3, could you PR that too?

PR #264, opened 2024-04-14
Add Github Actions for Elixir tests

L-e-x-o-n:
Adds a Github action that runs Teiserver tests.

Many tests are failing at the moment and as a result it is expected that this action will currently fail as well, however, it is still useful for future use and for checking the effect of new commits on tests without having to run them locally each time.


StanczakDominik (approved):
This is the kind of change that cannot hurt anything and will be much easier to poke holes in once merged :)

Beherith:
Thank you both! <3

PR #269, opened 2024-04-15
Disocrd channels improvements

L-e-x-o-n:
- Discord channels configs were removed from Discord site configs at teiserver/admin/site, they are now set at http://localhost:4000/admin/discord_channels (which is not a new thing but it was only used for a few channels before)
 Make sure to transfer any missing channel IDs from the old to new configs
- All channel IDs are now read from the the new config when sending messages
- Replaced usage of Nostrum.Api with Teiserver.Communication where possible Teiserver.Communication uses Nostrum.Api in the background. By using Teiserver.Communication everywhere it is  easier to replace Nostrum.APi if it breaks in the future


lib/teiserver/bridge/discord_bridge_bot.ex
```diff
@@ -127,13 +127,13 @@ defmodule Teiserver.Bridge.DiscordBridgeBot do
...
+    response = case data.name do
+      "textcb" ->
+        Teiserver.Bridge.TextcbCommand.execute(interaction, options_map)
 
-    #   _ ->
-    #     nil
-    # end
+      _ ->
+        nil
+    end
```
geekingfrog, line 136: why the uncomment here? It seems this is going to override the response.
I'm not sure what is this `textcb` command, but at least it should be

```elixir
    response = case data.name do
      "textcb" ->
        Teiserver.Bridge.TextcbCommand.execute(interaction, options_map)
      _ ->
        CommandLib.handle_command(interaction, options_map)
    end

```

lib/teiserver/libs/teiserver_configs.ex
```diff
@@ -431,86 +431,6 @@ defmodule Teiserver.TeiserverConfigs do
...
-      description: "The discord ID for the forum for starting GDT discussions"
-    })
-
-    add_site_config_type(%{
-      key: "teiserver.Discord forum #gdt-voting",
-      section: "Discord",
-      type: "integer",
-      permissions: ["Admin"],
-      description: "The discord ID for the forum for starting GDT voting"
-    })
```
geekingfrog, line 512: Nice, that got me confused for a while trying to figure out where these things lived (or not actually).

geekingfrog (commented):
When I'm running the server with my discord config, I can send messages, but it seems there's one problem:

```
[error] Task #PID<0.2860.0> started from Teiserver.Bridge.DiscordBridgeBot terminating
** (ArgumentError) the Access calls for keywords expect the key to be an atom, got: 1238756874313011224
    (elixir 1.14.3) lib/access.ex:313: Access.get/3
    (teiserver 0.1.0) lib/teiserver/bridge/discord_bridge_bot.ex:47: Teiserver.Bridge.DiscordBridgeBot.handle_event/1
    (elixir 1.14.3) lib/task/supervised.ex:89: Task.Supervised.invoke_mfa/2
    (stdlib 4.2) proc_lib.erl:240: :proc_lib.init_p_do_apply/3
Function: #Function<5.64421518/0 in Teiserver.Bridge.DiscordBridgeBot.handle_info/2>
    Args: []
```

It seems the function `bridge_channel_to_room` does need some change as well.

lib/teiserver/bridge/discord_bridge_bot.ex
L-e-x-o-n: Text callbacks are the  discord commands tha start with / and have autocomplete, for teiserver they return some premade responses on commonly asked questions. This isn't working for some reason so I was trying out different things.

PR #271, opened 2024-04-16
Make minimal game duration to treat game as ranked configurable, lower default 180 -> 90

### LINKED ISSUE #214: 3 minute timer for ranked to count is way too long for 1v1 format.
I realise the system is in place for 8v8s and to protect the other 15 players in a lobby, However the reasoning badly falls apart in a 1v1 setting for reason I will outline.

1) 1v1 games can often straight up end before the 3 minute mark, I killed someones commander in 1:55 in a game earlier, that might sound extreme but it's genuinely not that uncommon.

2) It can create some weird anti patterns not only in performing cheese builds, but also in DEFENDING cheese builds. If someone comm drops me and I manage to stop it from working, the game /strongly/ incentivises me not to finish him off, it somehow makes sense for me to risk a resurgent enemy, that's bananas!

3) I can't remember the last time I played a 1v1 where by 3 minutes you don't already have a strong sense of if you're ahead or behind. If someone scout rushes me and I lose my first con, why would I not just GG before 3 mins and not lose OS?**

** You may say something like, that's against the spirit of the rules and could be moderatable, and I understand that point of view, but please consider if that's your line of thought you are actually saying someone can be punished for not intentionally staying alive in a game they already know they lost just for someone elses benefit, I have to just sit around and play for another 10 mins waiting to die? and yet if I am straight up killed before 3 mins my opponent goes away empty handed?

I think it would really make sense to add some sort of conditional logic to the 3 minute check, like maybe it's only 1 minute in a 1v1, or doesn't check at all in a 1v1.

StanczakDominik:
Fixes #214.

Relevant discussion in
https://discord.com/channels/549281623154229250/855772061095559179/1210540660604538951

Basically, some 1v1 games do legitimately conclude before 3:00, notably 2:33 is a reported today timing to shut down a comm drop. Also tick/rover rushes. This keeps coming up on the discord. 90s should be pretty optimal. And if it isn't, well, it's now configurable so we can just change the duration :)

Tested to not blow up the site, not tested in-game yet.

### DIFF
--- a/lib/teiserver/game/libs/match_rating_lib.ex
+++ b/lib/teiserver/game/libs/match_rating_lib.ex
@@ -4,7 +4,7 @@ defmodule Teiserver.Game.MatchRatingLib do
   to balance matches. For that use Teiserver.Battle.BalanceLib.
   """
 
-  alias Teiserver.{Account, Coordinator, Game, Battle}
+  alias Teiserver.{Account, Coordinator, Config, Game, Battle,}
   alias Teiserver.Data.Types, as: T
   alias Teiserver.Repo
   alias Teiserver.Battle.{BalanceLib, MatchLib}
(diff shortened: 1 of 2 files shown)


StanczakDominik:
Tested to work. I played two matches on the integration server, resigned at 60s in one of them, resigned at 120s in the other. Both show up in https://server5.beyondallreason.info/teiserver/admin/matches/user/44674 (which does not take ratings into account), only the latter shows up in https://server5.beyondallreason.info/teiserver/admin/users/ratings/44674 (which does).

PR #273, opened 2024-04-16
WIP: Add rankicon parameter to CLIENTSTATUS command

jauggy:
# Context
Contributors use a high rank to determine their rank icon. The problem is that this makes contributors look like pro gamers and they might not be. In order to create more flexibility to rank icons, we should separate them from rank. If we can separate it from rank we could use [rank icons made by Ice](https://discord.com/channels/549281623154229250/564591092360675328/1212462364960497716)

The current protocol is this:
```
CLIENTSTATUS userName status
```

status contains bits of information including rank.

# Proposed Changes
I propose to send this instead from TeiServer
```
CLIENTSTATUS userName status rankIcon
```
where rankIcon is a string that can take these values:

- 1Chev, 2Chev, ...8Chev
- 1Moderator, 2Moderator, ...8Moderator
- 1Contributor, 2Contributor, ...8Contributor
- 1Streamer, 2Streamer, ...8Streamer
- 1Mentor, 2Mentor, ...8Mentor
- Bot

But it's very flexible and can really be anything. The numeric prefix represents their chev level purely based on playtime. After that the postfix depends on if they have a special role or not. If a user has multiple roles it will pick one in this order: Moderator > Contributor > Streamer > Mentor.

# Required SPADS modification
SPADS will error due to the extra argument. The var/plugins/barmanager.py needs the `hCLIENTSTATUS_pre` signature to change (see below)

# Required Chobby Modifications
No change is required to Chobby to prevent errors. Chobby is using regex that will simply ignore the extra argument.

# Test Steps
### 1. Fix SPADS Plugin
Before launching SPADS you will need to modify `var/plugins/barmanager.py`
Change this line
```
def hCLIENTSTATUS_pre(command, userName, status):
```
to this
```
def hCLIENTSTATUS_pre(command, userName, status, rankIcon=None):
```
Also at the top of this function add a log:
```
	spads.slog("Rank icon: " + str(rankIcon), 0)
```
Now launch SPADS as normal
```
perl spads.pl etc/spads.conf
```

### 2. Add TeiServer Test Users
Launch teiserver with this command (that allows running Elixir commands)
```
iex -S mix phx.server
```
In the same terminal run this to generate test users
```
TestScript.run()
```

This will generate the following users:
```
1Chev, 2Chev, 3Chev, 4Chev, 5Chev, 6Chev, 7Chev, 8Chev
```
Each of them will have a password of `password`. Their chev level is defined by their name.

### 3. Login with test users
Launch Chobby and login with any of the test users. In the SPADS logs you will see what rank icon is being sent. Now go to Teiserver website and login with `root@localhost` with password `password`
Go to Users > Find the user > Edit > Change their role to include any of Moderator/Contributor/Streamer/Mentor
Relogin with that user in Chobby. Check the SPADS logs and it should be different.

### DIFF
--- a/config/dev.exs
+++ b/config/dev.exs
@@ -88,7 +88,7 @@ config :teiserver, Oban,
   crontab: false
 
 # Do not include metadata nor timestamps in development logs
-config :logger, :console, format: "[$level] $message\n"
+config :logger, :console, format: "[$level] $message\n", level: :info
 
 config :logger,
   backends: [
--- a/lib/teiserver/account.ex
+++ b/lib/teiserver/account.ex
@@ -137,7 +137,7 @@ defmodule Teiserver.Account do
     |> Repo.one()
   end
 
-  @spec get_user_stat_data(integer()) :: Map.t()
+  @spec get_user_stat_data(integer()) :: map()
   def get_user_stat_data(userid) do
     Teiserver.cache_get_or_store(:teiserver_user_stat_cache, userid, fn ->
       case get_user_stat(userid) do
--- a/lib/teiserver/account/libs/role_lib.ex
+++ b/lib/teiserver/account/libs/role_lib.ex
@@ -263,7 +263,7 @@ defmodule Teiserver.Account.RoleLib do
   end
 
   def allowed_role_management("Moderator") do
-    global_roles() ++ moderation_roles() ++ property_roles()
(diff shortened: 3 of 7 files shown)


jauggy:
The code works but requires a DB call whenever we update the client's status. Not good

PR #275, opened 2024-04-16
(Blocked) Add rank_icon to ADDUSER command

jauggy:
# Context
Contributors use a high rank to determine their rank icon. The problem is that this makes contributors look like pro gamers and they might not be. In order to create more flexibility to rank icons, we should separate them from rank. If we can separate it from rank we could use [rank icons made by Ice](https://discord.com/channels/549281623154229250/564591092360675328/1212462364960497716)

Teiserver currently sends this for [ADDUSER ](https://springrts.com/dl/LobbyProtocol/ProtocolDescription.html#ADDUSER:server):
```
    "ADDUSER #{client.name} #{client.country} #{client.userid} #{client.lobby_client}\n"

```

# Proposed Changes
I propose to send this instead from TeiServer
```
    "ADDUSER #{client.name} #{client.country} #{client.userid} #{client.lobby_client} #{client.rank_icon}\n"
```
where `rank_icon` is a string that can take these values:

- 1Chev, 2Chev, ...8Chev
- 1Moderator, 2Moderator, ...8Moderator
- 1Contributor, 2Contributor, ...8Contributor
- 1Streamer, 2Streamer, ...8Streamer
- 1Mentor, 2Mentor, ...8Mentor
- Bot

But it's very flexible and can really be anything. The numeric prefix represents their chev level purely based on playtime. After that the postfix depends on if they have a special role or not. If a user has multiple roles it will pick one in this order: Moderator > Contributor > Streamer > Mentor.

# Required SPADS modification
SPADS doesn't error with this modification. However I have no idea why it doesn't error.

# Required Chobby Modifications
No change is required to Chobby to prevent errors. Chobby is using regex that will simply ignore the extra argument.

# Test Steps
### 1. Launch SPADS
```
perl spads.pl etc/spads.conf
```
This is the normal command to launch spads.

### 2. Add TeiServer Test Users
Launch teiserver with this command (that allows running Elixir commands)
```
iex -S mix phx.server
```
In the same terminal run this to generate test users
```
TestScript.run()
```

This will generate the following users:
```
1Chev, 2Chev, 3Chev, 4Chev, 5Chev, 6Chev, 7Chev, 8Chev
```
Each of them will have a password of `password`. Their chev level is defined by their name.

### 3. Login with test users
Launch Chobby and login with any of the test users. Check via logs what rank_icon is being send (Not sure how to log this)

Now go to Teiserver website and login with `root@localhost` with password `password`
Go to Users > Find the user > Edit > Change their role to include any of Moderator/Contributor/Streamer/Mentor

Relogin with that user in Chobby. Check the logs and it should be different.

### DIFF
--- a/lib/teiserver/account/libs/role_lib.ex
+++ b/lib/teiserver/account/libs/role_lib.ex
@@ -263,7 +263,7 @@ defmodule Teiserver.Account.RoleLib do
   end
 
   def allowed_role_management("Moderator") do
-    global_roles() ++ moderation_roles() ++ property_roles()
+    global_roles() ++ moderation_roles() ++ property_roles() ++ community_roles()
   end
 
   def allowed_role_management(_) do
--- a/lib/teiserver/account/libs/role_lib.ex
+++ b/lib/teiserver/account/libs/role_lib.ex
@@ -259,11 +259,12 @@ defmodule Teiserver.Account.RoleLib do
   end
 
   def allowed_role_management("Admin") do
-    staff_roles() ++ privileged_roles() ++ allowed_role_management("Moderator")
+    # Remove this code later; it has been fixed in another PR
+    staff_roles() ++ privileged_roles() ++ allowed_role_management("Moderator") ++ community_roles()
   end
 
   def allowed_role_management("Moderator") do
-    global_roles() ++ moderation_roles() ++ property_roles() ++ community_roles()
+    global_roles() ++ moderation_roles() ++ property_roles()
   end
 
   def allowed_role_management(_) do
--- a/config/dev.exs
+++ b/config/dev.exs
@@ -88,7 +88,7 @@ config :teiserver, Oban,
   crontab: false
 
 # Do not include metadata nor timestamps in development logs
-config :logger, :console, format: "[$level] $message\n"
+config :logger, :console, format: "[$level] $message\n", level: :info
 
 config :logger,
   backends: [
--- a/lib/teiserver/account.ex
+++ b/lib/teiserver/account.ex
@@ -137,7 +137,7 @@ defmodule Teiserver.Account do
     |> Repo.one()
   end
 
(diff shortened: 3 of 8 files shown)


lib/teiserver/account/libs/role_lib.ex
```diff
@@ -263,7 +263,7 @@ defmodule Teiserver.Account.RoleLib do
   end
 
   def allowed_role_management("Moderator") do
-    global_roles() ++ moderation_roles() ++ property_roles()
+    global_roles() ++ moderation_roles() ++ property_roles() ++ community_roles()
```
jauggy, line 266: Might have to undo this if it's been fixed in another ticket

jauggy:
While the code in this PR works, it is currently blocked due to not knowing how to get this information into the in-game player list. Getting the data into chobby should be working fine.

Beherith:
I would honestly expect SPADS to panic at this, ill have to test it on integration.

Beherith:
I think I would much prefer to make this an extensible field, likely via base64'd json. There have been multiple requests to transfer additional metadata in ADDUSER.

jauggy:
@Beherith just to clarify you want the last argument to be base 64 json string instead of just rank icon?  Example:
```
    "ADDUSER #{client.name} #{client.country} #{client.userid} #{client.lobby_client} #{base64json_extradata}\n"
```

jauggy:
> I would honestly expect SPADS to panic at this, ill have to test it on integration.
I later figured out why it doesn't error. It's because there's an existing bug. Teifion does not send cpu in the ADDUSER command. BarManager.py is expecting it 

```
def hADDUSER(command, userName, country = "nil", cpu = "nil", userID = "nil", lobbyID = "nil"):
```

So everything is off by one.

jauggy:
Not working on this so closing

Issue #236, opened 2024-03-27
Make the server configurable at runtime, not compile time.


p2004a:
Currently the server configuration is part of the **build process**, including the `prod.secret.exs`, this is very cursed for many reason:
 - Any change to configuration requires recompile of application
 - Because config can impact which dependencies are compiled, when building in reproducible way, changing config invalidates a lot of caching
 - We can't prebuild anything, or build the software outside of the production machine which should be the only place where secrets are stored unencrypted
 - We can't use the same exact binary version in the integration and production environment
 - This highly increases the production server deployment complexity as now it also needs to contain the CI to build the software

This needs to be refactored to load configuration at **runtime**, when server starts, potentially also allowing to reload part of the configuration without having to restart the whole server just like it's done in many other servers that are supposed to be long running.

p2004a:
Most likely need to steal by how it's done in https://github.com/Teifion/angen/blob/master/config/runtime.exs

geekingfrog:
Other useful link:
* elixir guide about [configuration and release](https://hexdocs.pm/elixir/config-and-releases.html)
* the [doc](https://hexdocs.pm/elixir/main/Config.html) for the config module

geekingfrog:
FYI: I started working on that

StanczakDominik:
Closed by #278

PR #278, opened 2024-04-21
Move config to runtime

geekingfrog:
Remove `config/prod.secret.exs` and instead use `config.runtime.exs`. This address https://github.com/beyond-all-reason/teiserver/issues/236
Now, most of the server can be configured through this file that is loaded at runtime. Changing it and restarting the server is enough.
Sensitive values like secrets are loaded through the environment. I provided an example for an EnvironmentFile.

I split up the changes in many small commits so hopefully it should be easy to review and make sure I haven't missed anything. I based my work on the provided example production secret file.

There are still some things like cipher config in `config/prod.exs` that are baked into the compiled bytecode since they are unlikely to change.

@p2004a This is difficult to test that locally since I have no idea what the actual production secret file looks like. Feel free to amend this new runtime file as needed.


p2004a (commented):
Thank you! Now, my review is purely from the configurability for the deployment, Not elixir or how things are supposed to be in elixir as I'm not familiar with the language.

First, so see how sane, more cleaned up `prod.secret.exs` looks like, take a look at https://github.com/beyond-all-reason/ansible-teiserver/blob/main/roles/teiserver/templates/prod.secret.exs.j2 It is inferred from real one and also some settings tuned slightly.

You can see there how much and where I have to be able to template in Jinja dynamically. And there are still some variables there that I should template and cleanup, e.g:
- TLS certificates location
- More things in the mailer configuration

To be honest the main major PITA is Discord, because I have to change how it's build, change `mix.exs` depending If I want to have it enabled or not: https://github.com/beyond-all-reason/ansible-teiserver/blob/main/roles/teiserver/templates/teiserver.Dockerfile.j2#L45-L51. Per https://github.com/Kraigie/nostrum/blob/3d8d5483353a25a09fc370c57a995af879bafbaa/guides/advanced/multi_node.md#bundling-nostrum-with-our-app it looks like it's possible to do it like... there? Not sure if it's appropriate to also bundle in this PR but that's another "configuration" bit outside of prod.secret.exs I had to deal with.

geekingfrog:
I've added some commits so that the configuration now matches what you linked me.
You shouldn't need to template the configuration anymore and should be able to achieve the same result purely through environment variables that you can pass to the docker container now.

AdamChlupacek:
@geekingfrog Do you think it would be possible to make ports, on which the server listens (tcp, tls, tachyon), configurable as well? I am currently running into an issue where I have to change the values in config locally since I have some other services running on these ports. It would be nice to be able to change the ports via config/environment and not having to have to change code every time.

geekingfrog:
@AdamChlupacek Done in f7be2eec3a78d6fe506bdd458b7a6daa777f8958 (also rebased https://github.com/beyond-all-reason/teiserver/pull/281)

config/runtime.exs
```diff
@@ -0,0 +1,170 @@
...
+# files are automatically recompiled on the fly and thus, config/{dev,test}.exs
+# are just fine
+if config_env() == :prod do
+  # used for mailing, checking origins, finding tls certs…
+  domain_name = Teiserver.ConfigHelpers.get_env("DOMAIN_NAME", "beyondallreason.info")
+
+  certificates = [
+    keyfile: "/etc/letsencrypt/live/#{domain_name}/privkey.pem",
+    certfile: "/etc/letsencrypt/live/#{domain_name}/cert.pem",
+    cacertfile: "/etc/letsencrypt/live/#{domain_name}/fullchain.pem"
```
p2004a, line 37: All those locations need to be configurable, those paths effectively assume certbot usage which is invalid.

config/runtime.exs
```diff
@@ -0,0 +1,170 @@
...
+  config :teiserver, TeiserverWeb.Endpoint,
+    url: [host: domain_name],
+    check_origin: check_origin,
+    https:
+      certificates ++
+        [
+          versions: [:"tlsv1.2"],
+          # dhfile is not supported for tls 1.3
+          # https://www.erlang.org/doc/man/ssl.html#type-dh_file
+          dhfile: Teiserver.ConfigHelpers.get_env("HTTPS_DH_FILE_PATH", "/etc/ssl/dhparam.pem")
```
p2004a, line 95: why this isn's together with `certificates` ?

config/runtime.exs
```diff
@@ -0,0 +1,170 @@
...
+    secret_key: Teiserver.ConfigHelpers.get_env("GUARDIAN_SECRET_KEY")
+
+  if Teiserver.ConfigHelpers.get_env("ENABLE_EMAIL_INTEGRATION", :bool) do
+    smtp_server = Teiserver.ConfigHelpers.get_env("SMTP_HOSTNAME")
+
+    config :teiserver, Teiserver.Mailer,
+      adapter: Bamboo.SMTPAdapter,
+      noreply_name: "Beyond All Reason",
+      noreply_address: "noreply@#{domain_name}",
+      contact_address: "info@#{domain_name}",
```
p2004a, line 111: I'm not confident about this assumption that noreply and info are like this, and based on the domain. Maybe ok default, but should be configurable.

config/runtime.exs
```diff
@@ -0,0 +1,170 @@
...
+          do: :verify_peer,
+          else: :verify_none
+        ),
+      # or {":system", ALLOWED_TLS_VERSIONS"} w/ comma seprated values (e.g. "tlsv1.1,tlsv1.2")
+      allowed_tls_versions: [:"tlsv1.2"],
+      # can be `true`
+      ssl: false,
+      retries: 1,
+      # can be `true`
+      no_mx_lookups: false,
```
p2004a, line 132: The 3 above variables are basically defaults, I would skip setting them entirely to reduce clutter.

config/runtime.exs
```diff
@@ -0,0 +1,170 @@
...
+      # auth: :if_available # can be `always`. If your smtp relay requires authentication set it to `always`.
+      auth: :always
+  end
+
+  log_root_path = "/var/log/teiserver/"
```
p2004a, line 137: Should be configurable

config/runtime.exs
```diff
@@ -0,0 +1,170 @@
...
+# ## Using releases
+#
+# If you use `mix release`, you need to explicitly enable the server
+# by passing the PHX_SERVER=true when you start it:
+#
+#     PHX_SERVER=true bin/teiserver start
+#
+# Alternatively, you can use `mix phx.gen.release` to generate a `bin/server`
+# script that automatically sets the env var above.
+if Teiserver.ConfigHelpers.get_env("PHX_SERVER", nil) do
```
p2004a, line 23: Why is this needed? I don't follow

config/runtime.exs
```diff
@@ -0,0 +1,170 @@
...
+  config :teiserver, Teiserver.Repo,
+    username: Teiserver.ConfigHelpers.get_env("DB_USERNAME"),
+    password: Teiserver.ConfigHelpers.get_env("DB_PASSWORD"),
+    database: Teiserver.ConfigHelpers.get_env("DB_NAME"),
+    pool_size: 40,
+    timeout: 120_000,
+    queue_interval: 2000
+
+  check_origin =
+    if Teiserver.ConfigHelpers.get_env("ACQUIRE_TLS_CERTIFICATE", false, :bool) do
```
p2004a, line 80: The name of this variable suggests some sort of automatic TLS acquiring. This should be more descriptive, more related to "check_origin" functionality.

p2004a (changes requested):
I would also likely prefix all environment variables with `TEI_` to make it obvious what they are for. It's useful for debugging when dumping the process environment and also it prevents conflicts.

I've started integrating those PRs into the ansible teiserver but quickly hit some stuff that can't configure and have to.

config/runtime.exs
geekingfrog: I amended the commit, the 3 paths are now expected to be in an environment variable.

config/runtime.exs
geekingfrog: I used these as default. In the amended commit, the config now look for `TEI_NOREPLY_EMAIL_ADDRESS` and `TEI_CONTACT_EMAIL_ADDRESS`.

config/runtime.exs
geekingfrog: I've removed `ssl` and `retries`. It's not clear that `no_mx_lookups` defaults to false though so I left it.

config/runtime.exs
geekingfrog: Added `TEI_LOG_ROOT_PATH`, if not present, will default to this value.

config/runtime.exs
geekingfrog: This is a standard env var expected by phoenix, the framework used by teiserver. If it's not set, there won't be any actual http server start up.
I don't think we should modify that one, or prefix it with `TEI_` since the tooling like `mix release` will generate scripts that expects or set this specific env variable.
That should be transparent from your side though, if you start with `teiserver/bin/teiserver start` this handled automatically.

config/runtime.exs
geekingfrog: Renamed this one into `TEI_SHOULD_CHECK_ORIGIN`

config/runtime.exs
geekingfrog: The certificates are required in two places, under `TeiserverWeb.Endpoint` and also under `Teiserver`. I'm not yet 100% sure why, I suspect it's because the `TeiserverWeb` is the html server for the website, and `Teiserver` is the thing handling lobbies and autohosts, but they both require the similar tls setup.

p2004a (changes requested):
I have draft integration for this in the playbook now https://github.com/beyond-all-reason/ansible-teiserver/pull/1 the server starts with a few fixes as mentioned in the comments but will also some testing if all switches work as intended.

config/runtime.exs
p2004a: Yes, I understand that, I'm just not clear why the dhparams is only for one of them, in all configs I've seen they are set in both.

geekingfrog: ah, I found where and how these are used. Here's [an example](https://github.com/geekingfrog/teiserver/blob/3b72d871c2b56101646b33418d7af5cd189c7e3b/lib/teiserver/tcp/spring/spring_tcp_server.ex#L19-L23).

I checked the available options, and I could indeed pull dhfile alongside.
I'd rather do that in another PR if possible, that one is already quite big and changing how we setup existing tcp/ssl pools seems out of scope for this one which purely aim at making the configuration more env-var driven.

If that's alright with you, I'll do these change later.

p2004a: Sounds good

config/runtime.exs
```diff
@@ -0,0 +1,169 @@
...
+    keyfile: Teiserver.ConfigHelpers.get_env("TEI_TLS_PRIVATE_KEY_PATH"),
+    certfile: Teiserver.ConfigHelpers.get_env("TEI_TLS_CERT_PATH"),
+    cacertfile: Teiserver.ConfigHelpers.get_env("TEI_TLS_CA_CERT_PATH")
+  ]
+
+  # this is used in lib/teiserver_web/controllers/account/setup_controller.ex
+  # as a special endpoint to create the root user. Setting it to empty or nil
+  # will disable the functionality completely.
+  # There is already a root user, so disable it
+  config :teiserver, Teiserver.Setup, key: nil
```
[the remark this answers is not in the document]
geekingfrog: done in 6a9490b9212d3eeacd5bbea1edc44184d7a40433

config/runtime.exs
```diff
@@ -0,0 +1,169 @@
...
+        ],
+    http: [:inet6, port: Teiserver.ConfigHelpers.get_env("TEI_PORT", "4000", :int)],
+    secret_key_base: Teiserver.ConfigHelpers.get_env("TEI_HTTP_SECRET_KEY_BASE")
+
+  config :teiserver, Teiserver.Account.Guardian,
+    issuer: Teiserver.ConfigHelpers.get_env("TEI_GUARDIAN_ISSUER", "teiserver"),
+    secret_key: Teiserver.ConfigHelpers.get_env("TEI_GUARDIAN_SECRET_KEY")
+
+  if Teiserver.ConfigHelpers.get_env("TEI_ENABLE_EMAIL_INTEGRATION", :bool) do
+    smtp_server = Teiserver.ConfigHelpers.get_env("TEI_SMTP_HOSTNAME")
```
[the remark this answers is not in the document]
geekingfrog: indeed, and below it should be `TEI_SMTP_HOSTNAME` with no fallback on the server.

geekingfrog:
I've pushed a few more changes. Thanks for the pointer to the discord (and smtp) guide.
I managed to build the application and run it locally with the updated example env file and everything seems to work, I even got the email setup working.

```
MIX_ENV=prod mix release --overwrite
set -o allexport; source documents/prod_files/example-environment-file; set +o allexport
./_build/prod/rel/teiserver/bin/teiserver start
```

config/runtime.exs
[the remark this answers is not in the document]
p2004a: You can drop this line entirely, this variable is no longer use below

config/runtime.exs
```diff
@@ -0,0 +1,170 @@
...
+    keyfile: Teiserver.ConfigHelpers.get_env("TEI_TLS_PRIVATE_KEY_PATH"),
+    certfile: Teiserver.ConfigHelpers.get_env("TEI_TLS_CERT_PATH"),
+    cacertfile: Teiserver.ConfigHelpers.get_env("TEI_TLS_CA_CERT_PATH")
+  ]
+
+  # this is used in lib/teiserver_web/controllers/account/setup_controller.ex
+  # as a special endpoint to create the root user. Setting it to empty or nil
+  # will disable the functionality completely.
+  config :teiserver, Teiserver.Setup,
+    key: "---- random string of alpha numeric characters ----"
```
p2004a, line 44: The point of my previous comment was to also allow overriding it via environment variable.

It should be settable easily from e.g. my Ansible playbook, right now in my playbook I would have to modify this file. It's helpful primarily for local testing setup of the playbook.

Default, per comment, should be also empty string, not what's set now.

config/runtime.exs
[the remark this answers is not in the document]
geekingfrog: good catch, dropped it.

config/runtime.exs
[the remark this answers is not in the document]
geekingfrog: I amended the commit, it's now looking for `TEI_SETUP_ROOT_KEY`, if not set, defaults to `nil`, which effectively disable the feature, which is what we want when the root user is setup.

config/runtime.exs
```diff
@@ -0,0 +1,168 @@
...
+
+  if Teiserver.ConfigHelpers.get_env("TEI_ENABLE_EMAIL_INTEGRATION", :bool) do
+    config :teiserver, Teiserver.Mailer,
+      adapter: Bamboo.SMTPAdapter,
+      contact_address:
+        Teiserver.ConfigHelpers.get_env("TEI_CONTACT_EMAIL_ADDRESS", "info@#{domain_name}"),
+      noreply_name: "Beyond All Reason",
+      noreply_address:
+        Teiserver.ConfigHelpers.get_env("NOREPLY_EMAIL_ADDRESS", "noreply@#{domain_name}"),
+      smtp_server: Teiserver.ConfigHelpers.get_env("TEI_SMTP_SERVER"),
```
p2004a, line 113: The key name is `server`, not `smtp_server`, I get

```
May 13 23:05:39 teiserver-integration teiserver[1513468]: Request: POST /send_password_reset
May 13 23:05:39 teiserver-integration teiserver[1513468]: ** (exit) an exception was raised:
May 13 23:05:39 teiserver-integration teiserver[1513468]:     ** (ArgumentError) The following settings have not been found in yo>
May 13 23:05:39 teiserver-integration teiserver[1513468]: * Key server is required for SMTP Adapter
```

geekingfrog: fixed

config/runtime.exs
```diff
@@ -0,0 +1,168 @@
...
+    secret_key: Teiserver.ConfigHelpers.get_env("TEI_GUARDIAN_SECRET_KEY")
+
+  if Teiserver.ConfigHelpers.get_env("TEI_ENABLE_EMAIL_INTEGRATION", :bool) do
+    config :teiserver, Teiserver.Mailer,
+      adapter: Bamboo.SMTPAdapter,
+      contact_address:
+        Teiserver.ConfigHelpers.get_env("TEI_CONTACT_EMAIL_ADDRESS", "info@#{domain_name}"),
+      noreply_name: "Beyond All Reason",
+      noreply_address:
+        Teiserver.ConfigHelpers.get_env("NOREPLY_EMAIL_ADDRESS", "noreply@#{domain_name}"),
```
p2004a, line 112: Missing `TEI_` prefix... took me embarrassingly long time to find because the email wasn't delivered only between the SMTP server and my real email...

geekingfrog: Fixed

p2004a (approved):
It works, can be deployed to prod once prod server is rebuild with Ansible playbook.

PR #279, opened 2024-04-21
Fix ratings columns on smaller screens

jauggy:
Currently in production it looks like this for small screens:

This ticket gives a better layout for smaller screens by adjusting the column widths and padding. Also I changed the uncertainty delta color to grey because it will always go down and users might be confused why that column is always red (and it's not bad to go down).

# Test Steps
Go here http://localhost:4000/battle/ratings

Test on resolutions such as 
1360 x 768
1280 x 1024
(or similar if you don't have any of these screen sizes)

### DIFF
--- a/assets/scss/custom/_styles.scss
+++ b/assets/scss/custom/_styles.scss
@@ -248,7 +248,7 @@ body {
 
 //Map column for Ratings table
 .table-ratings tr th:nth-child(1) {
-  width: 200px;
+  width: 12.5rem;
 }
 
 .table-ratings th {
--- a/assets/scss/custom/_styles.scss
+++ b/assets/scss/custom/_styles.scss
@@ -240,6 +240,29 @@ body {
   padding: 0.25rem 0.6rem;
 }
 
+//Ratings table
+.table-ratings {
+  table-layout: fixed;
+  font-size: 0.9rem;
+}
+
+//Map column for Ratings table
+.table-ratings tr th:nth-child(1) {
+  width: 200px;
+}
+
+.table-ratings th {
+  padding: 0.5rem
+}
+
+.table-ratings td {
+  padding: 0.5rem
+}
+
+// Make the delta column closer to the associated value
+.table-ratings .column-delta {
+  transform: translate(-1.6rem, 0rem)
+}
 
 
 // We want out outline buttons to be filled in when you mouseover or activate them
--- a/lib/teiserver_web/live/battles/match/ratings.html.heex
+++ b/lib/teiserver_web/live/battles/match/ratings.html.heex
@@ -54,7 +54,7 @@
       </div>
     <% end %>
 
-    <table class="table table-sm table-hover mt-3 cursor-pointer">
+    <table class="table table-sm table-hover table-ratings mt-3 cursor-pointer">
       <thead>
         <tr class="">
           <th>Map name</th>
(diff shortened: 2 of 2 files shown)


jauggy:
@L-e-x-o-n can you please review if you have spare time thanks.

L-e-x-o-n:
# 1360x768 and 1280 x 1024
Should uncertainty arrow should always be next to number (not in next row)?
## PR
image

## Original 1360x768 and 1280 x 1024
image

# 1920x1080
For reference,  looks the same PR or not except grey uncertainty
image

jauggy:
So previously the arrow was always next to number in same row. Then the uncertainty columns were added causing some columns to get squashed on smaller screens.

jauggy:
@L-e-x-o-n can you try in incognito mode. Maybe css is getting cached?

L-e-x-o-n:
Tried, on latest PR, same thing :( Firefox

jauggy:
I tested with Firefox again (latest version 125.0.2 and it works). can you do an experiment for me? 

Open up 
`assets/scss/custom/_styles.scss` and modify to add background-color to the Map column:
```
//Map column for Ratings table
.table-ratings tr th:nth-child(1) {
  width: 200px;
  background-color: red;
}
```

Do you get the new color for the map heading?

L-e-x-o-n:
> I tested with Firefox again (latest version 125.0.2 and it works). can you do an experiment for me?
> 
> Open up `assets/scss/custom/_styles.scss` and modify to add background-color to the Map column: Find this section
> 
> ```
> //Map column for Ratings table
> .table-ratings tr th:nth-child(1) {
>   width: 200px;
> }
> ```
> 
> Then modify to this:
> 
> ```
> //Map column for Ratings table
> .table-ratings tr th:nth-child(1) {
>   width: 200px;
>   background-color: red;
> }
> ```
> 
> Do you get the new color for the map heading?

Doesn't work, even with cleared cache...

jauggy:
Part of the local setup is sass install
```
mix sass.install
```
@L-e-x-o-n Can you run that and try again?

L-e-x-o-n:
> Part of the local setup is sass install
> 
> ```
> mix sass.install
> ```
> 
> @L-e-x-o-n Can you run that and try again?

No changed :(
Would be nice if someone else can try, maybe it's just me :(

jauggy:
@geekingfrog can you help test this PR? If the css is working correctly you should be able to go here:
http://localhost:4000/battle/ratings
and the columns will all be same width except the first one. For some reason it's not working for Lexon.

AdamChlupacek:
I can confirm the PR works as expected: 

On master: 
Screenshot 2024-04-23 at 15 27 09

On your branch: 
Screenshot 2024-04-23 at 15 27 49

This is working as expected.

L-e-x-o-n (approved):

jauggy:
Thanks Adam. I wonder why css is not working on Lexon's machine.

assets/scss/custom/_styles.scss
```diff
@@ -240,6 +240,29 @@ body {
...
 
+//Ratings table
+.table-ratings {
+  table-layout: fixed;
+  font-size: 0.9rem;
+}
+
+//Map column for Ratings table
+.table-ratings tr th:nth-child(1) {
+  width: 200px;
```
geekingfrog, line 251: not a fan of putting dimensions in pixels, since they don't adapt with font size. So if the user changes its font size, it gets out of whack easily. The default font size is 16px, that means 200px is `12.5rem`.

geekingfrog:
It breaks down under 1080px for viewport width, but otherwise it's a net improvement, can fix that (mobile view?) later.

assets/scss/custom/_styles.scss
jauggy: Have updated to use REM and confirmed it looks the same.

geekingfrog (approved):

PR #281, opened 2024-04-22
Dynamic discord integration

geekingfrog:
# ONLY LOOK AT THE LAST 4 COMMITS

This enable or disable the discord bridge through environment variable. This way, we can have one release bundle of teiserver that can be used for staging with the integration disabled, and the same bundle in prod.
This was done following the [nostrum doc about bundling it with the app](https://hexdocs.pm/nostrum/multi_node.html#bundling-nostrum-with-our-app).

This assume https://github.com/beyond-all-reason/teiserver/pull/278 is merged first (hence the warning about only looking at the last 4 commits).

@p2004a I think that's going to be useful for your work on ansible+docker

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -24,11 +24,32 @@ end
 # files are automatically recompiled on the fly and thus, config/{dev,test}.exs
 # are just fine
 if config_env() == :prod do
+  # used for mailing, checking origins, finding tls certs…
+  domain_name = System.get_env("DOMAIN_NAME", "beyondallreason.info")
+
+  certificates = [
+    keyfile: System.fetch_env!("TLS_PRIVATE_KEY_PATH"),
+    certfile: System.fetch_env!("TLS_CERT_PATH"),
+    cacertfile: System.fetch_env!("TLS_CA_CERT_PATH")
+  ]
 
   # this is used in lib/teiserver_web/controllers/account/setup_controller.ex
   # as a special endpoint to create the root user. Setting it to empty or nil
   # will disable the functionality completely.
   # There is already a root user, so disable it
   config :teiserver, Teiserver.Setup, key: nil
+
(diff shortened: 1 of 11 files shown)


config/runtime.exs
```diff
@@ -0,0 +1,188 @@
...
+    metadata: [:request_id, :user_id],
+    level: :notice
+
+  config :logger, :info_log,
+    path: "#{log_root_path}info.log",
+    format: "$date $time [$level] $metadata $message\n",
+    metadata: [:request_id, :user_id],
+    level: :info
+
+  if enable_discord_bridge do
```
p2004a, line 174: This seems to miss the `config :teiserver, DiscordBridgeBot,` part I have https://github.com/beyond-all-reason/ansible-teiserver/blob/main/roles/teiserver/templates/prod.secret.exs.j2#L73-L77

I'm not sure how without that it would know to what server to connect, how have you tested it?

geekingfrog: I can't really test the actual integration since I don't have access to any token. Are there some test credentials somewhere I could use?
The only test I've done is that, when the discord integration is disabled, I can start the server normally and it won't complain that it's missing stuff for discord.

I've added the configuration for the bridge in 3b72d871c2b56101646b33418d7af5cd189c7e3b

p2004a: You can follow https://github.com/beyond-all-reason/ansible-teiserver?tab=readme-ov-file#discord to create discord for testing, there are also instructions there for email testing etc.

geekingfrog: I've amended the commit for the bot config, followed your steps and now it seems to work.
2024-05-11-475x166-scrot

p2004a (approved):
It works, can be deployed to prod once prod server is rebuild with Ansible playbook.

PR #283, opened 2024-04-23
FixChange fake data accounts to generate a list. 

AdamChlupacek:
Fake data could not be loaded since accounts did not generated into a list.

### DIFF
--- a/lib/teiserver/mix_tasks/fake_data.ex
+++ b/lib/teiserver/mix_tasks/fake_data.ex
@@ -109,8 +109,8 @@ defmodule Mix.Tasks.Teiserver.Fakedata do
             inserted_at: Timex.shift(Timex.now(), days: -day, minutes: -minutes) |> time_convert,
             updated_at: Timex.shift(Timex.now(), days: -day, minutes: -minutes) |> time_convert
           }
-        end) |> Enum.to_list
-      end) |> Enum.to_list
+        end) |> Enum.to_list()
+      end) |> Enum.to_list()
       |> List.flatten()
 
     Ecto.Multi.new()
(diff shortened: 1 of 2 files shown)


jauggy:
I was able to get Teifion's fake data script to run before. What does this PR actually do? How to test?

AdamChlupacek:
If you run the fake data script which is now on master, it will fail at generating accounts. There was a breaking change introduced [here](https://github.com/beyond-all-reason/teiserver/commit/08e45caa64276c4564edd2d36da7869193135e9b), this just fixes the issue converting the `ParallelStream` back into a `List`, so that `List.flatten()` can actually be called and we get a list of accounts to be inserted to the DB.

jauggy:
Ah I see. When I used the fake data script that was before @L-e-x-o-n made that merge.

L-e-x-o-n:
Good find and fix.

I checked other places where ParallelStream is used again, this should be applied there as well.
This is  my mistake, I did some tests, but with unit tests being broken and parallel being used in currently mostly unused parts of Teiserver, I missed it.

Would you be interested in adding `|> Enum.to_list` to those other places (6 ParallelStream usages, excluding this one) as part of this PR or should I fix it in another PR?

L-e-x-o-n (approved):

AdamChlupacek:
@L-e-x-o-n sure I can add it to the PR over the weekend.

AdamChlupacek:
@L-e-x-o-n I went through the usecases and in most cases there is `Enum.count` already called, so it did not needed to have anything added. I only identified one spot where it needed to be added.

L-e-x-o-n:
> @L-e-x-o-n I went through the usecases and in most cases there is `Enum.count` already called, so it did not needed to have anything added. I only identified one spot where it needed to be added.

Nice, thanks.

jauggy (approved):
This PR is important to be merged because in master branch now you cannot run the test data. Therefore new contributors will get stuck. This PR fixes this and allows test data creation.

jauggy:
@StanczakDominik @Beherith 
I've tested and this ticket works and is important to merge. Basically if new contributors follow the setup instructions for teiserver, they will eventually need to create test data. Without this merged, they won't be able to create test data and therefore cannot complete the setup instructions.

The local setup instructions have this command for adding fake data:
```
mix teiserver.fakedata
```
that will fail on master branch currently.

Beherith:
Thx!

PR #285, opened 2024-05-01
Allow friends to join private battles without a password

jere0500:
I want to allow friends to join private lobbies without having to enter a password.
This should be easy by bypassing the password requirement when the gatekeeper of a lobby is set to friends.

General Idea:
https://github.com/beyond-all-reason/BYAR-Chobby/issues/654

### DIFF
--- a/lib/teiserver/lobby.ex
+++ b/lib/teiserver/lobby.ex
@@ -565,7 +565,7 @@ defmodule Teiserver.Lobby do
       Enum.any?([
         CacheUser.is_moderator?(user),
         Enum.member?(user.roles, "Caster"),
-        consul_reason == :override_approve
+        consul_reason in [:override_approve, :allow_friends]
       ])
 
     ignore_locked =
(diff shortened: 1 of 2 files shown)


lib/teiserver/lobby.ex
```diff
@@ -565,7 +565,8 @@ defmodule Teiserver.Lobby do
       Enum.any?([
         CacheUser.is_moderator?(user),
         Enum.member?(user.roles, "Caster"),
-        consul_reason == :override_approve
+        consul_reason == :override_approve,
+        consul_reason == :allow_friends
```
L-e-x-o-n, line 569: You can combine these 2 conditions with `consul_reason in [:override_approve, :allow_friends]`

jere0500: alright, done

L-e-x-o-n (approved):

StanczakDominik:
All right, thanks for doing this! Happy to unblock chobby on it, it should be deployed within the week :)

PR #286, opened 2024-05-13
Add indexes for foreign key constraints

geekingfrog:
Related to the discussion on discord.
Once merged, the migration will be applied on the next server startup.

This can also be applied manually before:
```bash
mix ecto.migrations # display migrations status
mix ecto.migrate --log-migrations-sql # apply pending migrations and print the SQL used
mix ecto.rollback # rollback the last applied migration
```

The indexes are coming from `\d+ teiserver_battle_matches` and creating any index in the `Referenced by` section that don't already exist.


Beherith:
Thank you, made this job literally 1000x faster!

Issue #289, opened 2024-05-15
Teiserver swallows all email sending errors


p2004a:
If sending emails fails, there isn't any error returned to user or log message created.

Calls to `Teiserver.Mailer.deliver_now` aren't checking the return tuple, and code isn't written to handle failures in sending of the emails: e.g. reset code is created before sending email and not cleaned up if it fails, so if email sending fails it's not logged, user code is still created blocking new recovery codes creation, and user has have to wait 24h to retry sending password link.

StanczakDominik:
Closed by #290

p2004a:
@StanczakDominik Is that also handling registration?

StanczakDominik:
I'll double check

geekingfrog:
I did a check locally, and it only logs an error, there's no feedback for chobby.
The server logs the error:
```
[error] Error sending new user email - <THE-EMAIL> - {:network_failure, '127.0.0.1', {:error, :econnrefused}}
```
but the code is still in the database.

I created: https://github.com/beyond-all-reason/teiserver/issues/310

PR #290, opened 2024-05-18
Delete reset token if email failure

geekingfrog:
Address https://github.com/beyond-all-reason/teiserver/issues/289

If, for whatever reason, sending the password reset email fails, then attempt to delete the reset code so that the operation can be retried.

Also fixed a template so that flashes are shown on the login page as well.

This is what the admin UI looks like:
2024-05-18-971x497-scrot
2024-05-18-1010x577-scrot

and this is what the user facing interface looks like:
2024-05-18-1937x473-scrot
2024-05-18-1937x386-scrot

### DIFF
--- a/lib/teiserver/account/emails.ex
+++ b/lib/teiserver/account/emails.ex
@@ -1,7 +1,34 @@
 defmodule Teiserver.Account.Emails do
   @moduledoc false
+  require Logger
   alias Bamboo.Email
   alias Teiserver.Helper.TimexHelper
+
+  def send_password_reset(user, code \\ nil) do
+    {code, email} = password_reset(user, code)
+    response = Teiserver.Mailer.deliver_now(email, response: true)
+
+    case response do
+      {:ok, _email, _response} ->
+        :ok
+
+      {:error, error} ->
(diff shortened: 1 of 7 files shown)


geekingfrog:
Also tested the flow when registering a new user through chobby, and got the email fine:

2024-05-19-734x170-scrot

L-e-x-o-n (approved):
Nice!

StanczakDominik:
For the sake of testing, how would you force that password reset email send to fail?

geekingfrog:
> For the sake of testing, how would you force that password reset email send to fail?

I rebased this branch on top of latest master, this way it's using the new config system. In the integration server, the easiest would be to change the host/password for the mailer configuration to something that will fail.
This way, any attempt to send an email would fail and trigger this codepath.

Locally I force the email integration and don't start the local docker container that simulate the smtp server (same thing as putting a wrong hostname).

I don't think there's a way to test that without config/code change otherwise.

StanczakDominik (approved):
Tested it by temporarily messing up the mailer password on integration; task failed successfully 😁 Thank you!

Issue #272, opened 2024-04-16
Cannot save community roles in dev environment


jauggy:
Steps
1. Run teiserver locally
2. Launch website and login as `root@localhost`      `password`
3. Find a user and edit them
4. Tick the community roles
5. Save changes
6. Go to edit user again. The checkboxes should be ticked but aren't.
![1](https://github.com/beyond-all-reason/teiserver/assets/1305569/cff15fbf-9d0f-4826-88be-5230fc585a83)

jauggy:
@L-e-x-o-n Can you please verify if you get this bug on your local machine or if it's just me?

AdamChlupacek:
I can fix this, please assign to me.

StanczakDominik:
I saw something like this while following https://github.com/beyond-all-reason/teiserver/pull/256, but the difference is that I wasn't able to give my spads bot the Bot role, it wouldn't save it - like you're seeing here. I ended up completely messing up my root permissions and doing `mix teiserver.fakedata` to sidestep it.

jauggy:
> I saw something like this while following #256, but the difference is that I wasn't able to give my spads bot the Bot role, it wouldn't save it - like you're seeing here. I ended up completely messing up my root permissions and doing `mix teiserver.fakedata` to sidestep it.

Just tested and it I can set someone to bot if I use `root@localhost` . Make sure you test with Administrator login because they are the only one who can give bot role.

jere0500:
Tldr: I still have the problem, because booleans from the checkboxes are not transferred into the role field.

Unfortunately this issue still persisted for me, maybe it is due to the elixir version (I used 1.14.5) or some dependency behaving differently. (I logged in as root@localhost so that shouldn't be an issue)
Thats why I decided to dig into it more.

In a nutshell: 3 files are relevant here:
1. The html form page for editing the user (where the checkboxes get loaded): https://github.com/beyond-all-reason/teiserver/blob/master/lib/teiserver_web/templates/admin/user/form.html.heex
2. The change request processing from the form: https://github.com/beyond-all-reason/teiserver/blob/3e13abb6cb79f17c684db2aa8b38fb581724c7e2/lib/teiserver/account/schemas/user.ex#L52
3. The part where the update gets written into the database: https://github.com/beyond-all-reason/teiserver/blob/3e13abb6cb79f17c684db2aa8b38fb581724c7e2/lib/teiserver/account/libs/user_lib.ex#L169

### The Problem
[...]

PR #294, opened 2024-05-25
Fix server role updates

jere0500:
This my hacky draft, mentioned in #272. Which works for my setup. 
There is likely a better way to do it. It might even be obsolete due to using the wrong version.


jere0500:
Issue transferred to #295

geekingfrog:
The issue you linked #295 seems to be closed now. Does that mean this PR is effectively obsolete and we can close it?

jere0500:
Yes, it is obsolete.
Closing

PR #304, opened 2024-06-03
add mix format and dialyzer to CI

### LINKED ISSUE #298: Mix dialyzer fails
## Test Steps
Run
```
mix dialyzer
```
This does typechecking. You will get this error:
```
lib/teiserver/helpers/oban_logger.ex:43:call
The function call will not succeed.

System.convert_time_unit(any(), :native, :milliseconds)

breaks the contract
(integer(), time_unit() | :native, time_unit() | :native) :: integer()
```

Looking at the valid units
https://hexdocs.pm/elixir/1.12/System.html#t:time_unit/0

I think it should be `:millisecond` not `:milliseconds`

@Beherith this is a file you have worked on recently so it would be good if you can investigate.

StanczakDominik:
Closes #298


geekingfrog:
Very nice, big fan of checking these tedious things in CI.
It's probably worth putting some doc in the README, or the [documents/guides/local_setup.md](local setup guide) to also [setup git to ignore some commits when blaming](https://www.stefanjudis.com/today-i-learned/how-to-exclude-commits-from-git-blame/).

PR #307, opened 2024-06-04
Add the ability to add and remove users from the webui

jere0500:
## Idea
In Addition to managing user by script for testing purposes. I thought it would be easier to also allow creation and deletion of users directly in the webui. 

I therefore added a button to the list menu.
AddButton

The button opens a form where users can be simply generated. Either create a random user or completely/ partially filling form.
CreationForm

In order to delete users again I also added delete buttons to the end of the users in the list.
Deletion

Deletion can also be done in the user action dropdown menu.
ActionMenu

Addition/ Deletion is only visible and usable by accounts with system permission.

(Unfortunately the icons of Fontawesome do only partially work, probably due to the wrong version)

### Current problems:
I have problems with the delete query raising errors as some tables didn't exist in my setup (e.g. telemetry). I therefore switched to ```Ecto.Adapters.SQL.query``` instead of ```Ecto.Adapters.SQL.query!``` to not raise errors. 
I also split up the deletion from moderation_reports, because the combined query would fail due to certain columns not existing.
This might not be a problem on other setups.

Also deletion of user like the *spadsbot* user does not work, because the spadsbot creates lobbies and more foreign key dependencies.

There might be some better ways to route to the correct form for user creation.

I would be happy to get some feedback.

### DIFF
--- a/lib/teiserver/account/tasks/delete_user_task.ex
+++ b/lib/teiserver/account/tasks/delete_user_task.ex
@@ -11,7 +11,6 @@ defmodule Teiserver.Admin.DeleteUserTask do
     id_list
     |> Enum.each(&Account.decache_user/1)
 
-
     int_id_list = Enum.map(id_list, fn x -> String.to_integer(x) end)
 
     [
@@ -27,17 +26,15 @@ defmodule Teiserver.Admin.DeleteUserTask do
       "DELETE FROM account_friends WHERE user1_id = ANY($1) OR user2_id = ANY($1)",
 
       # Telemetry
-      "DELETE FROM telemetry_complex_client_event_types WHERE user_id = ANY($1)",
-      "DELETE FROM telemetry_simple_client_event_types WHERE user_id = ANY($1)",
-      "DELETE FROM telemetry_complex_match_event_types WHERE user_id = ANY($1)",
-      "DELETE FROM telemetry_simple_match_event_types WHERE user_id = ANY($1)",
-      "DELETE FROM telemetry_complex_lobby_event_types WHERE user_id = ANY($1)",
-      "DELETE FROM telemetry_simple_lobby_event_types WHERE user_id = ANY($1)",
-      "DELETE FROM telemetry_complex_server_event_types WHERE user_id = ANY($1)",
-      "DELETE FROM telemetry_simple_server_event_types WHERE user_id = ANY($1)",
-      "DELETE FROM telemetry_user_properties WHERE user_id = ANY($1)",
-      "DELETE FROM telemetry_simple_server_events WHERE user_id = ANY($1)",
+      "DELETE FROM telemetry_complex_client_events WHERE user_id = ANY($1)",
+      "DELETE FROM telemetry_simple_client_events WHERE user_id = ANY($1)",
+      "DELETE FROM telemetry_complex_match_events WHERE user_id = ANY($1)",
+      "DELETE FROM telemetry_simple_match_events WHERE user_id = ANY($1)",
+      "DELETE FROM telemetry_complex_lobby_events WHERE user_id = ANY($1)",
       "DELETE FROM telemetry_simple_lobby_events WHERE user_id = ANY($1)",
+      "DELETE FROM telemetry_complex_server_events WHERE user_id = ANY($1)",
+      "DELETE FROM telemetry_simple_server_events WHERE user_id = ANY($1)",
+      "DELETE FROM telemetry_user_properties WHERE user_id = ANY($1)",
 
       # User table extensions/stats
       "DELETE FROM teiserver_account_user_stats WHERE user_id = ANY($1)",
@@ -52,7 +49,7 @@ defmodule Teiserver.Admin.DeleteUserTask do
       # Chat
       "DELETE FROM teiserver_lobby_messages WHERE user_id = ANY($1)",
       "DELETE FROM teiserver_room_messages WHERE user_id = ANY($1)",
-      "DELETE FROM direct_messages WHERE from_user_id = ANY($1) OR to_user_id = ANY($1)",
+      "DELETE FROM direct_messages WHERE from_id = ANY($1) OR to_id = ANY($1)",
 
       # Match stuff
       "DELETE FROM teiserver_battle_match_memberships WHERE user_id = ANY($1)",
@@ -60,12 +57,11 @@ defmodule Teiserver.Admin.DeleteUserTask do
       "DELETE FROM teiserver_game_rating_logs WHERE user_id = ANY($1)",
 
       # Moderation
-      "DELETE FROM moderation_reports WHERE reporter_id = ANY($1)",
-      "DELETE FROM moderation_reports WHERE target_id = ANY($1)",
-      "DELETE FROM moderation_actions WHERE responder_id = ANY($1)"
+      "DELETE FROM moderation_reports WHERE reporter_id = ANY($1) OR target_id = ANY($1)",
+      "DELETE FROM moderation_actions WHERE target_id = ANY($1)"
     ]
     |> Enum.each(fn query ->
-      Ecto.Adapters.SQL.query(Repo, query, [int_id_list])
+      Ecto.Adapters.SQL.query!(Repo, query, [int_id_list])
     end)
 
     # And now the users
--- a/lib/teiserver/account/tasks/delete_user_task.ex
+++ b/lib/teiserver/account/tasks/delete_user_task.ex
@@ -11,6 +11,9 @@ defmodule Teiserver.Admin.DeleteUserTask do
     id_list
     |> Enum.each(&Account.decache_user/1)
 
+
+    int_id_list = Enum.map(id_list, fn x -> String.to_integer(x) end)
+
     [
       # Clan memberships
       "DELETE FROM teiserver_clan_memberships WHERE user_id = ANY($1)",
@@ -33,6 +36,8 @@ defmodule Teiserver.Admin.DeleteUserTask do
       "DELETE FROM telemetry_complex_server_event_types WHERE user_id = ANY($1)",
       "DELETE FROM telemetry_simple_server_event_types WHERE user_id = ANY($1)",
       "DELETE FROM telemetry_user_properties WHERE user_id = ANY($1)",
+      "DELETE FROM telemetry_simple_server_events WHERE user_id = ANY($1)",
+      "DELETE FROM telemetry_simple_lobby_events WHERE user_id = ANY($1)",
 
       # User table extensions/stats
       "DELETE FROM teiserver_account_user_stats WHERE user_id = ANY($1)",
@@ -55,16 +60,17 @@ defmodule Teiserver.Admin.DeleteUserTask do
       "DELETE FROM teiserver_game_rating_logs WHERE user_id = ANY($1)",
 
       # Moderation
-      "DELETE FROM moderation_reports WHERE reporter_id = ANY($1) OR target_id = ANY($1) OR responder_id = ANY($1)",
-      "DELETE FROM moderation_actions WHERE target_id = ANY($1)"
+      "DELETE FROM moderation_reports WHERE reporter_id = ANY($1)",
+      "DELETE FROM moderation_reports WHERE target_id = ANY($1)",
+      "DELETE FROM moderation_actions WHERE responder_id = ANY($1)"
     ]
     |> Enum.each(fn query ->
-      Ecto.Adapters.SQL.query!(Repo, query, [id_list])
+      Ecto.Adapters.SQL.query(Repo, query, [int_id_list])
     end)
 
     # And now the users
     query = "DELETE FROM account_users WHERE id = ANY($1)"
-    Ecto.Adapters.SQL.query!(Repo, query, [id_list])
+    Ecto.Adapters.SQL.query!(Repo, query, [int_id_list])
 
     # Delete our cache of them
     id_list
--- a/lib/teiserver/data/cache_user.ex
+++ b/lib/teiserver/data/cache_user.ex
@@ -263,14 +263,16 @@ defmodule Teiserver.CacheUser do
 
     case Account.script_create_user(params) do
       {:ok, user} ->
+        verification_code = (:rand.uniform(899_999) + 100_000) |> to_string
+
         Account.update_user_stat(user.id, %{
-          "verification_code" => (:rand.uniform(899_999) + 100_000) |> to_string
+          "verification_code" => verification_code
         })
 
         # Now add them to the cache
         user
         |> convert_user
-        |> Map.put(:password_hash, spring_md5_password(password))
+        |> Map.put(:password_hash, encrypt_password(spring_md5_password(password)))
         |> Map.put(:spring_password, false)
         |> add_user
         |> update_user(persist: true)
@@ -289,15 +291,24 @@ defmodule Teiserver.CacheUser do
             :ok
 
           true ->
-            case EmailHelper.new_user(user) do
-              {:error, error} ->
-                Logger.error("Error sending new user email - #{user.email} - #{error}")
+            case Application.fetch_env(:teiserver, Teiserver.Mailer) do
+              {:ok, _} ->
+                case EmailHelper.new_user(user) do
+                  {:error, error} ->
+                    Logger.error("Error sending new user email - #{user.email} - #{error}")
 
-              :no_verify ->
-                verify_user(get_user_by_id(user.id))
+                  :no_verify ->
+                    verify_user(get_user_by_id(user.id))
 
-              {:ok, _, _} ->
-                :ok
+                  {:ok, _, _} ->
+                    :ok
+                end
+
+              :error ->
+                # case, where no mailer is defined, skipping the verification
+                Account.delete_user_stat_keys(user.id, ~w(verification_code))
+
+                Logger.warn("Skipped sending mail, mailer is not defined")
             end
         end
 
--- a/lib/teiserver/data/cache_user.ex
+++ b/lib/teiserver/data/cache_user.ex
@@ -291,24 +291,15 @@ defmodule Teiserver.CacheUser do
             :ok
 
           true ->
-            case Application.fetch_env(:teiserver, Teiserver.Mailer) do
-              {:ok, _} ->
-                case EmailHelper.new_user(user) do
-                  {:error, error} ->
-                    Logger.error("Error sending new user email - #{user.email} - #{error}")
+            case EmailHelper.new_user(user) do
+              {:error, error} ->
+                Logger.error("Error sending new user email - #{user.email} - #{error}")
 
-                  :no_verify ->
-                    verify_user(get_user_by_id(user.id))
+              :no_verify ->
+                verify_user(get_user_by_id(user.id))
 
-                  {:ok, _, _} ->
-                    :ok
-                end
-
-              :error ->
-                # case, where no mailer is defined, skipping the verification
-                Account.delete_user_stat_keys(user.id, ~w(verification_code))
-
-                Logger.warn("Skipped sending mail, mailer is not defined")
+              {:ok, _, _} ->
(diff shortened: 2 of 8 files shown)


lib/teiserver/account/tasks/delete_user_task.ex
```diff
@@ -55,16 +60,17 @@ defmodule Teiserver.Admin.DeleteUserTask do
...
     ]
     |> Enum.each(fn query ->
-      Ecto.Adapters.SQL.query!(Repo, query, [id_list])
+      Ecto.Adapters.SQL.query(Repo, query, [int_id_list])
     end)
 
     # And now the users
     query = "DELETE FROM account_users WHERE id = ANY($1)"
-    Ecto.Adapters.SQL.query!(Repo, query, [id_list])
+    Ecto.Adapters.SQL.query!(Repo, query, [int_id_list])
```
geekingfrog, line 73: >I have problems with the delete query raising errors as some tables didn't exist in my setup (e.g. telemetry). I therefore switched to Ecto.Adapters.SQL.query instead of Ecto.Adapters.SQL.query! to not raise errors.

You should ensure your setup is correct first. There are some guide in the repo, and you might be interested in [postgres setup](https://github.com/beyond-all-reason/teiserver/blob/master/documents/guides/local_setup.md)

I have the telemetry tables locally and I didn't recall doing anything fancy to get them.

The main problem with swapping `query!/3` for `query/3` is that it doesn't raise an error and returns the exception as a value, but that doesn't fix the underlying problem.

If anything, one thing to fix here would be to wrap all the deletes inside a transaction, but that's irrelevant for your changes.

lib/teiserver/account/tasks/delete_user_task.ex
```diff
@@ -55,16 +60,17 @@ defmodule Teiserver.Admin.DeleteUserTask do
       "DELETE FROM teiserver_game_rating_logs WHERE user_id = ANY($1)",
 
       # Moderation
-      "DELETE FROM moderation_reports WHERE reporter_id = ANY($1) OR target_id = ANY($1) OR responder_id = ANY($1)",
-      "DELETE FROM moderation_actions WHERE target_id = ANY($1)"
+      "DELETE FROM moderation_reports WHERE reporter_id = ANY($1)",
+      "DELETE FROM moderation_reports WHERE target_id = ANY($1)",
+      "DELETE FROM moderation_actions WHERE responder_id = ANY($1)"
```
geekingfrog, line 65: :+1: good catch updating these missing relations.

lib/teiserver_web/controllers/admin/user_controller.ex
```diff
@@ -208,6 +208,54 @@ defmodule TeiserverWeb.Admin.UserController do
...
+    conn
+    |> render("create_form.html")
+  end
+
+  @spec create_post(Plug.Conn.t(), map) :: Plug.Conn.t()
+  def create_post(conn, params \\ %{}) do
+    (params["name"] || "" == "")
+
+    passwordfn = fn -> if is_nil(params["password"]) or String.trim(params["password"]) == "" do "password" else params["password"] end end
+    emailfn = fn -> if is_nil(params["email"]) or String.trim(params["email"]) == "" do UUID.uuid1() else params["email"] end end
```
geekingfrog, line 222: you're going to want to setup the formatting tool, either `mix format` or your IDE.

lib/teiserver_web/controllers/admin/user_controller.ex
```diff
@@ -208,6 +208,54 @@ defmodule TeiserverWeb.Admin.UserController do
...
+
+
+    user_params = %{
+      "name" => params["name"],
+      "password" => passwordfn.(),
+      "email" => emailfn.(),
```
geekingfrog, line 228: Is there any reason you're invoking the lambdas you just created above instead of getting values directly?
I'm thinking of

```elxir
password = if ... do: "password", else: ... end
```

lib/teiserver_web/router.ex
```diff
@@ -561,6 +561,8 @@ defmodule TeiserverWeb.Router do
 
     # User stuff
     put("/users/gdpr_clean/:id", UserController, :gdpr_clean)
+    get("/users/delete_user/:id", UserController, :delete_user)
+    put("/users/delete_user/:id", UserController, :delete_user)
```
geekingfrog, line 565: The routing is roughly REST-ish, so it would be better to use the `delete` function (and associated http verb) to delete the user.

lib/teiserver_web/router.ex
```diff
@@ -617,6 +619,8 @@ defmodule TeiserverWeb.Router do
     get("/tools/falist", ToolController, :falist)
     get("/tools/test_page", ToolController, :test_page)
 
+    get("/users/create_form", UserController, :create_form)
+    put("/users/create_post", UserController, :create_post)
```
geekingfrog, line 623: Similar to the delete comment, this should be `post` for resource creation.

lib/teiserver_web/controllers/admin/user_controller.ex
```diff
@@ -208,6 +208,54 @@ defmodule TeiserverWeb.Admin.UserController do
...
   end
 
+  @spec create_form(Plug.Conn.t(), map) :: Plug.Conn.t()
+  def create_form(conn, _) do
+    conn
+    |> render("create_form.html")
+  end
+
+  @spec create_post(Plug.Conn.t(), map) :: Plug.Conn.t()
+  def create_post(conn, params \\ %{}) do
```
geekingfrog, line 218: You should add a guard to these functions first, that checks the user has indeed the correct permissions to perform that. Currently, the only check is done in the template rendering, but that is trivially bypassable.
Something along the lines of:

```elixir
def create_post(conn, _) when not Teiserver.Account.Authlib.allow?(conn, "Server") do
    # flash + redirect elsewhere
end

def create_post(conn, params \\ %{}) do
  # your function
end
```

lib/teiserver_web/controllers/admin/user_controller.ex
```diff
@@ -1104,4 +1152,24 @@ defmodule TeiserverWeb.Admin.UserController do
         |> redirect(to: ~p"/teiserver/admin/user")
     end
   end
+
+
+  @spec delete_user(Plug.Conn.t(), map()) :: Plug.Conn.t()
+  def delete_user(conn, %{"id" => id}) do
```
geekingfrog, line 1158: Similar comment regarding authentication.

Also, if the given parameters don't have an `id` key, the pattern match will fail and you're going to get a server error.
It would be better to add another catch all clause after this one to return back a 400 with a pre-rendered form (optional but nice to have)

lib/teiserver_web/controllers/admin/user_controller.ex
```diff
@@ -1104,4 +1152,24 @@ defmodule TeiserverWeb.Admin.UserController do
...
+    user = Account.get_user_by_id(id)
+
+    case Teiserver.Account.UserLib.has_access(user, conn) do
+      {true, _} ->
+        Teiserver.manually_delete_user(id)
```
geekingfrog, line 1163: That's my biggest problem in this change. `manually_delete_user` is a mix task, and is supposed to be invoked through the terminal with the mix command, or some script, but definitely not from application code.

There is `Teiserver.Account.delete_user` that is more suitable, although it doesn't remove all the linked rows.
A quick grep shows that it's likely a generated stub and isn't really used. I'd rather have this function taking care of deleting the various entities and then you can call it from there.

lib/teiserver_web/controllers/admin/user_controller.ex
```diff
@@ -1104,4 +1152,24 @@ defmodule TeiserverWeb.Admin.UserController do
         |> redirect(to: ~p"/teiserver/admin/user")
     end
   end
+
+
+  @spec delete_user(Plug.Conn.t(), map()) :: Plug.Conn.t()
+  def delete_user(conn, %{"id" => id}) do
+    user = Account.get_user_by_id(id)
+
+    case Teiserver.Account.UserLib.has_access(user, conn) do
```
geekingfrog, line 1161: :+1: for the access control here. I'd recommend using that approach or the one I suggested in both functions.

lib/teiserver_web/templates/admin/user/index.html.heex
```diff
@@ -1,5 +1,6 @@
 <% bsname = view_colour()
-is_moderator = allow?(@conn, "Moderator") %>
+is_moderator = allow?(@conn, "Moderator")
+is_system = allow?(@conn, "System") %>
```
geekingfrog, line 3: Why is this permission different than the `Server` you used elsewhere? I'm not on top of the permission system, so I may very well have missed something.

geekingfrog (commented):
Very nice to have an API to create users. This may get extended later, for testing purposes (both integration and load testing).
Perhaps this would be worth mentioning quickly in [the testing guide](https://github.com/beyond-all-reason/teiserver/blob/master/documents/guides/testing.md) as well.

I'm more concerned about the delete user button. There's a confirmation popup, but that's still a big irrevocable action and I'm not super sure we want that.
Perhaps someone with more experience using the admin website can comment on this @StanczakDominik (or ping someone else)?

I think without the delete and my suggested changes that would be a nice addition. Not even two days ago I ran into trouble trying to create some random local accounts to test something, and that would have helped.

L-e-x-o-n:
In my opinion, this shouldn't be creating users directly and should instead go through the existing `CacheUser.register_user`. That one goes through the whole process with a few more checks and verifications e.g. checking if name is available and clean, email isn't already registered, password is valid....

jauggy:
If you want this feature purely for testing then you could check for a config variable. For example, there are certain endpoints that only are enabled by this flag:
```
  enable_hailstorm: true,
```
that is set inside `dev.exs`
So you could also set up some similar flag to tell if we are in a test/dev environment where adding/deleting users is safe.

geekingfrog:
> In my opinion, this shouldn't be creating users directly and should instead go through the existing `CacheUser.register_user`. That one goes through the whole process with a few more checks and verifications e.g. checking if name is available and clean, email isn't already registered, password is valid....

oh you're right, it's much better there.
There is some logic around sending email as well, that is bypassed for some special domains (hailstorm related). So I'd suggest you default the emails of these new user to use the `@agents` domain, so you get a problem with emails.
One side effect of that is that you don't get the `Verified` role, but that may not be a problem, and that can be manually edited from the UI with an admin account, so no big deal.

jere0500:
Thanks for all the feedback!
I will look into fixing the parts. 

I tried it with Teiserver.CacheUser.register_user in a [previous version](https://github.com/beyond-all-reason/teiserver/commit/967b8f066a60e9ce2506125095c24f14fe548f41), 
but had some problems with signing in to the created user with chobby, because I could not get the verification code. 
Giving the role manually could have probably done the trick.
I will look into using it again! 

I also like the Idea of locking add/ delete user behind a flag for now, will check that one out.

lib/teiserver/account/tasks/delete_user_task.ex
jere0500: I did recreate my setup with the guide once again to make sure I did not miss anything.
However, the some columns still do not exist.
I observed that all ```telemetry_\*_event_types``` only have ```["id", "name"]```.
That makes sense, considering they only define the event types, not when events happen to users.
The ```telemetry_\*_events``` contain the ```user_id```, therefore I think this was originally meant. 

I also changed some queries, because the columns are named differently.

I don't have any problem using ```query!``` now

<details>
<summary>To look at the columns I used something like this:</summary>
<br>

```
[
"SELECT * FROM moderation_actions"
]
|> Enum.each(fn query ->
Ecto.Adapters.SQL.query!(Repo, query) |> IO.inspect()
end)
```
</details>

jere0500:
I now use ```Teiserver.CacheUser.register_user```.
When no Mailer is defined, no mail will be sent and the verification code will be set to empty. This allows signing with created users through chobby, bypassing the verification.
I also moved from ```manually_delete_user``` to ```Teiserver.Account.delete_user```.

Next, I will look at using ```delete``` and ```post``` to improve routing.

jere0500:
I now added the guards, changed to using ```delete``` and ```post``` for correct rooting.

The formatting (done with ```mix format```) should now be correct too.

### Idea
I also added a commit, which locks delete and add behind ```:test_mode``` which can be set in ```config/config.exs```. If this should be done differently or generally be available without setting ```:test_mode``` I will change that, or add some documentation to the local testing guide.

lib/teiserver_web/controllers/admin/user_controller.ex
jere0500: should be done now

lib/teiserver_web/router.ex
jere0500: should be correct now

lib/teiserver_web/controllers/admin/user_controller.ex
jere0500: This idea did not compile. I used the other approach mentioned further down.

lib/teiserver_web/controllers/admin/user_controller.ex
jere0500: ```manually_delete_user``` uses ```Teiserver.Admin.DeleteUserTask.delete_users``` internally, which I now also use. 
I could also use ```Teiserver.Account.delete_user``` if that is better.

lib/teiserver_web/templates/admin/user/index.html.heex
jere0500: I don't know what I originally thought here. I switched to ```is_server``` to have it the same way as the other.

lib/teiserver_web/router.ex
```diff
@@ -617,6 +621,11 @@ defmodule TeiserverWeb.Router do
     get("/tools/falist", ToolController, :falist)
     get("/tools/test_page", ToolController, :test_page)
 
+    if Application.compile_env(:teiserver, Teiserver)[:test_mode] do
```
geekingfrog, line 624: `compile_env` gets the value at compile time, and it's generally a bad idea. Because you need to recompile the app to change it. Also, this complicate matters when compiling the app from a different machine.

If you want to conditionally have these endpoints + template, a better way is to put that into the config files under `config/config.exs` set to false, and inside `config/test.exs` `config/dev.exs` put that to true.

But, tbh, I'm not sure all of this is warranted, requiring admin/server privilege to be able to create users manually may just be enough.

geekingfrog (commented):
I think you can drop the whole configuration bits and put a disclaimer on the templates like "this is for testing/local dev only, proceed with care in production" for both create&update endpoints. That should simplify the code.

The delete endpoint could just be deleted in my opinion.

Also, I'm very sorry for the super late feedback. It slipped my mind and then I was away. Don't hesitate to ping me here or on discord (channel teiserver-spads, I'm blackmelon on discord) if you need further feedback on this one.

lib/teiserver_web/router.ex
```diff
@@ -561,6 +561,10 @@ defmodule TeiserverWeb.Router do
 
     # User stuff
     put("/users/gdpr_clean/:id", UserController, :gdpr_clean)
+
+    if Application.compile_env(:teiserver, Teiserver)[:test_mode] do
+      delete("/users/delete_user/:id", UserController, :delete_user)
+    end
```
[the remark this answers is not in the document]
jauggy: Yeah just leave the delete feature out. Even in a dev environment it's not useful.

jauggy:
I do agree that the delete feature is not needed. But to be honest even the add user feature is not super useful too since `mix teiserver.fakedata` adds a whole bunch of fake users. What was the original driving force behind this PR?

jere0500:
Hey there, 
the original idea behind the feature was to simplify adding and removing of test users for testing purposes.
I understand that there are other ways to achieve the same. It was more meant as a shortcut to do it more quickly, e.g. when you need a quick test user with a short name. 

I plan to include the proposed requested changes in the next weeks. (remove the delete user endpoint, ...)

However, I understand if there is no need for this feature, then the PR can probably get closed for now.

jauggy:
One use case I need for testing purposes is the ability to give a user different ratings for Large Team/ Small Team/ Duel/ FFA, Team FFA. Not sure if this is something possible. Can be via editing an existing user or via creating a new user.

lib/teiserver/data/cache_user.ex
```diff
@@ -289,15 +291,24 @@ defmodule Teiserver.CacheUser do
             :ok
 
           true ->
-            case EmailHelper.new_user(user) do
-              {:error, error} ->
-                Logger.error("Error sending new user email - #{user.email} - #{error}")
+            case Application.fetch_env(:teiserver, Teiserver.Mailer) do
```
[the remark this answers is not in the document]
jere0500: I noticed that `EmailHelper.new_user` already implements some logic to disable the verification in the webui. So I decided to remove my workaround all together.

lib/teiserver_web/controllers/admin/user_controller.ex
```diff
@@ -208,6 +210,59 @@ defmodule TeiserverWeb.Admin.UserController do
...
+    end
+  end
+
+  @spec create_post(Plug.Conn.t(), map) :: Plug.Conn.t()
+  def create_post(conn, params \\ %{}) do
+    if is_nil(params["name"]) or String.trim(params["name"]) == "" do
+      conn
+      |> put_flash(:danger, "Invalid user name")
+      |> redirect(to: ~p"/teiserver/admin/user")
+    end
```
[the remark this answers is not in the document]
jere0500: Thanks for the hint. I now implemented it that way.

lib/teiserver_web/controllers/admin/user_controller.ex
```diff
@@ -208,6 +210,59 @@ defmodule TeiserverWeb.Admin.UserController do
...
+
+    if allow?(conn, "Server") and Application.get_env(:teiserver, Teiserver)[:test_mode] do
+      password =
+        if is_nil(params["password"]) or String.trim(params["password"]) == "" do
+          "password"
+        else
+          params["password"]
+        end
```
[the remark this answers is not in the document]
jere0500: I have tested it and this works well. In earlier iterations I had problems detecting "" as empty to trigger the default value.

lib/teiserver_web/router.ex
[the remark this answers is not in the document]
jere0500: I now removed that delete endpoint all together (for now).

lib/teiserver_web/router.ex
jere0500: I have removed that for now. I currently think relying on the admin/ server privilege should be sufficient.

lib/teiserver_web/templates/admin/user/section_menu.html.heex
```diff
@@ -40,6 +40,17 @@
       ) %>
     <% end %>
 
+    <%= if allow?(@current_user, "admin") do %>
```
L-e-x-o-n, line 43: ... but "admin" here.  I recommend using "Server" in both.

lib/teiserver_web/controllers/admin/user_controller.ex
```diff
@@ -210,6 +210,57 @@ defmodule TeiserverWeb.Admin.UserController do
     |> render("new.html")
   end
 
+  @spec create_form(Plug.Conn.t(), map) :: Plug.Conn.t()
+  def create_form(conn, _) do
+    if allow?(conn, "Server") do
```
L-e-x-o-n, line 215: You check for "Server" here...

geekingfrog:
I actually had the need for this recently. I ended up using the internal function in a repl, but I think this would be a nice addition.
The deletion feature, I think it's a bit dangerous and not really required locally.
But otherwise I'm up to get that in. Putting it behind Server permission is probably good enough.

L-e-x-o-n:
I am thinking about putting it behind dev env check as well, thoughts?

geekingfrog:
no strong opinions, bit more complex, but not much.

PR #308, opened 2024-06-05
Make CI great again

geekingfrog:
Alternative title: declare test bankruptcy.

This PR adds a tag to ignore all currently failing tests. The output of `mix test --exclude needs_attention` is still drowned in error logs, but at least all tests are passing. This is making the CI step useful, if a known good test starts to fail, people will notice.

And then we can start having a look at slowly removing these tags and fixing the tests. There is really a mix bags:

* Some very simple tests were generated once and obviously never run ever again, can probably delete them
* Some tests are about the old tachyon protocol based on plain tls connnection. These can be removed alongside the old tachyon code
* Some tests haven't been updated when the code changed, typically, spring protocol change/extension, or html output
* Some tests are exercising code looking at dirty cache and failing because of that


geekingfrog:
Should probably wait for https://github.com/beyond-all-reason/teiserver/pull/304 to be merged since the different otp/elixir version seems to be causing some issues.

StanczakDominik:
I'm still getting 29 failures locally, 3 more than here in CI :|

geekingfrog:
> I'm still getting 29 failures locally, 3 more than here in CI, when calling `MIX_ENV=test mix do ecto.drop + test --exclude needs_attention` :|

Yeah, this needs more work, I'll get to it at some point this week

geekingfrog:
riiiight, so this passes locally, but not in CI, that one isn't going to be fun

StanczakDominik:
Maybe even have multiple separate tags for those? Probably overkill and silly, but something to consider.

geekingfrog:
Alright, I managed to reproduce the errors.
I "fixed" them by disabling even more tests. They were all related to (old) tachyon stuff so I don't feel too bad. These tests are likely going to get deleted soon, once we get something working with the ws based tachyon.

jauggy (approved):
Nice looks like CI Tests all pass. Even though some tests have been disabled it at least makes CI useful again.

PR #309, opened 2024-06-05
Redirect to previous page after logging in

geekingfrog:
How to test?

* Log out of your account.
* go to http://localhost:4000/teiserver/account/parties
* get redirected to login page
* login correctly
* you should land on the `/teiserver/account/parties` page instead of the default `/`

This is convenient for now, but it's going to be important when implementing OAuth2 (for tachyon).


StanczakDominik:
Awesome! Tested, works nicely :)

PR #313, opened 2024-06-07
Free all the icons

### LINKED ISSUE #242: Stop depending on non-free fontawesome
Teiserver depends on the non-free version of https://fontawesome.com/. This makes it harder to build and distribute. Everyone that wants to, has to acquire the license themselves. This is very problematic in the open source project like this one.

geekingfrog:
Fix #242 

This is mostly a mindless search&replace and checking that everything looks good with a root account locally.
No attempt to refactor the icon and styling system, so there's still a little bit of duplication.

Some icons had to be swapped for a free version.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -7,7 +7,7 @@ config :teiserver, Teiserver,
   site_title: "BAR",
   site_suffix: "",
   site_description: "",
-  site_icon: "fa-duotone fa-robot",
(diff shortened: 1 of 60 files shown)


StanczakDominik:
This is one of those cases where I'm just going to merge this and gather feedback on any missing or broken icons later, as it's not at all critical for operation :)

PR #316, opened 2024-06-09
WIP: Fix sandbox not clearing after each test

jauggy:
## Context
If I create a user in test 1, it should be cleared after the test ends and then when I try to create the same user in test 2, it should work. If the sandbox db is not cleared and you try and create a user with a name that already exists, instead it creates a user with random name.

## Test Steps
```
mix test path/to/clear_db_each_test.exs
```

All tests should pass

## TODO
Currently the tests fail and not sure how to fix

https://github.com/beyond-all-reason/teiserver/assets/1305569/4448191a-9b59-45c3-b4f8-0a22e8c806ee


geekingfrog:
I found the issue and added a fix in https://github.com/beyond-all-reason/teiserver/pull/316
If you want to verify that, you can cherry pick the only commit of this PR (`git cherry-pick 8cc544e7eaf7e8e0627611e3d784dc88e08ab63c` on this branch) and run this test again.

The problem was that the DB was correctly reset between tests, but the cache wasn't. I don't think my fix is going to solve all cache issues but that's a start.


I don't want to have this PR merged though, since this is merely a test file to reproduce an issue.

jauggy:
@geekingfrog did you link the wrong PR?

geekingfrog:
Oops, yes, I did. I amended my comment and put the correct link in place.

Issue #321, opened 2024-06-10
Mix dialyzer errors on master


jauggy:
After checkout master and run `mix dialyzer` on local machine I get these errors:

```
lib/teiserver/libs/test_lib.ex:124:unknown_function
Function ExUnit.Callbacks.on_exit/1 does not exist.
________________________________________________________________________________
lib/teiserver_web/controllers/api/spads_controller.ex:221:unknown_type
Unknown type: T.lobby/0.
________________________________________________________________________________
lib/teiserver/account/libs/user_lib.ex:361:no_return
Function can_login/1 has no local return.
________________________________________________________________________________
lib/teiserver_web/controllers/account/session_controller.ex:102:pattern_match
The pattern can never match the type.

Pattern:
{:ok, _user}, _conn

Type:
{:error, <<_::152>>}, _

________________________________________________________________________________
lib/teiserver_web/controllers/api/spads_controller.ex:219:pattern_match
The pattern can never match the type.

Pattern:
nil

Type:
[any()] | integer()

________________________________________________________________________________
lib/teiserver_web/live/account/profile/overview.ex:294:pattern_match
The pattern can never match the type.

Pattern:
:ok

Type:
{:error, binary()}

________________________________________________________________________________
lib/teiserver_web/live/account/profile/overview.ex:315:pattern_match
The pattern can never match the type.

Pattern:
true

Type:
{:failure, binary()}

________________________________________________________________________________
lib/teiserver_web/live/account/profile/overview.ex:320:unused_fun
Function join_lobby/2 will never be called.
________________________________________________________________________________
done (warnings were emitted)
```

PR #322, opened 2024-06-10
Fix spads_controller type errors

jauggy:
Run this
```
mix dialyzer
```

And all the errors related to spads_controller will be gone. Full list of errors are in #321 (I didn't fix the other errors)

### DIFF
--- a/lib/teiserver_web/controllers/api/spads_controller.ex
+++ b/lib/teiserver_web/controllers/api/spads_controller.ex
@@ -3,6 +3,7 @@ defmodule TeiserverWeb.API.SpadsController do
   alias Teiserver.Config
   alias Teiserver.{Account, Coordinator, Battle}
   alias Teiserver.Battle.{BalanceLib, MatchLib}
+  alias Teiserver.Data.Types, as: T
   import Teiserver.Helper.NumberHelper, only: [int_parse: 1]
   require Logger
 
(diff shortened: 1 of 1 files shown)


lib/teiserver_web/controllers/api/spads_controller.ex
```diff
@@ -216,8 +217,6 @@ defmodule TeiserverWeb.API.SpadsController do
     end
   end
 
-  defp get_member_lobby(nil), do: nil
-
   @spec get_member_lobby(non_neg_integer()) :: T.lobby() | nil
```
StanczakDominik, line 220: Wouldn't a saner fix be something like...
```suggestion
  @spec get_member_lobby(non_neg_integer() | nil) :: T.lobby() | nil
```

(This might then need moving to the main definiton, removed by this PR?)

jauggy: That won't work. I tried it. I also asked on Elixir forums:
https://elixirforum.com/t/pattern-can-never-match-the-type/64094/6

Basically dialyzer reckons that all the calls it found to this function do not use nil. If you search there's only one line that calls it
```
    target_id = int_parse(target_id_str)
    lobby = get_member_lobby(target_id)
```
and dialyzer reckons that will never be nil. Just to confirm, I asked on Elixir forums if deleting is the appropriate answer. We'll see.

StanczakDominik: Oooh, okay, gotcha. Let's wait what the wise wizards reply :)

PR #328, opened 2024-06-12
Update split_one_chevs balancemode v2

jauggy:
## Context
I played a game with split_one_chevs on and noticed a few issues. The 1Chevs were not split. This was because Macwhite was recognised by Teiserver as a 2Chev whereas Chobby showed them as a 1Chev. This is likely because Chobby gets a player's rank on login and then never updates it so they get out of sync with TeiServer.

## How to resolve?
To resolve this my algorithm will now group both 1 and 2Chevs into the "Noob" bucket. There's also going to be slight change on how we draft players who are in the "Noob" bucket detailed below.

### Other Findings
When putting the replay
```
https://www.beyondallreason.info/replays?gameId=39096966518c40a66b2130b057864aa5
```
into https://openskill-test.web.app/ I noticed that the library expects Team 1 to win, despite that most humans would bet on Team 2. 

Since the library expects Team 1 to win, if my team were to win I would gain a lot of OS i.e. +1. If my team were to lose, I only lose -0.37.

From this we can conclude that if you were to choose from the three "Noobs": _CindersFire, Macwhite, Victorious_Dead_ you probably want to avoid the overrated players, which are likely the ones with highest uncertainty. If an overrated player is on the other team, you stand to win more and lose less, and since they're overrated, you're also more likely to win. Therefore, for those in the "Noobs" category, we probably want to pick those with low uncertainty as they are less likely to be overrated.

## split_one_chevs Algorithm v2
Based on these findings, the algorithm will now draft players based on these criteria:
1. Always pick experienced (3Chev+) players over noobs (1-2Chevs).
2. Prefer higher OS for experienced players.
3. Prefer lower uncertainty for noobs.

This draft mimics how a human might draft players with the given visible information in a lobby. It's not super mathematical. Players generally look at chevron level to determine how overrated someone might be. Someone did complain in chat about the lobby balance in the game I played mentioned above. They were obviously eyeballing the chevron levels and assuming those two players were overrated.

## Further enchancements

1. Previously, if the balancer was called it would check the result and if the deviation in ratings between teams was too high, it would rerun the balancer but split all parties into solo players. Now the balance result will have a new field: `has_parties?`. If this is false we do not need to rerun the balancer again.
2. Permissions have been reduced for fake users that are not admin. The purpose of this is if you want to test that the Balance tab doesn't appear to normal users, you need to reduce the permissions of fake users.
3. Anyone with Staff role e.g. Tester, Contributor, etc. can now see the balance tab. The balance tab now has a dropdown allowing you to switch to difference balance modes.
4. fuzz_multiplier (randomness added to match rating) is now only enabled for Teifion's algos. This makes it easier to debug issues.
5. If there are no noobs, an alternate balancer (`loser_picks`) will be called. This is the default algo and it supports parties.

## Known Bugs
Teiserver doesn't know the rank shown to the user in Chobby. Chobby gets the rank on login and then never updates. Teiserver, therefore, may classify a user as 2Chev but they might be shown as 1Chev in Chobby.

## Unit Tests
Run this to run multiple unit tests that relate to balance
```
  mix test --only balance_test
```

## Local Dev Tests
See comment [here](https://github.com/beyond-all-reason/teiserver/pull/328#issuecomment-2197302191) for test steps.

## Theoretical Testing on past replays
Go here: https://balance-algo-web.web.app/
And enter a past replay. Change algorithm to Split One Chevs v2


lib/teiserver_web/templates/admin/match/index.html.heex
jauggy, line 1: I worked on this PR before admin/match was removed in a PR about 10 hours ago. However I need admin/match page as it has a balance tab that is useful for balance testing. A lot of this code is not adding new stuff but restoring removed stuff.

jauggy:
Have updated and now you can use the url to see various balances
```
http://localhost:4000/battle/4/balance/split_one_chevs
```

jauggy:
Have updated and now you can see dropdown to change the balancer. Also removed the admin pages (now redirects) to be consistent with the previous PR made by Perfi to remove them.

jauggy:
Sample video of testing balancer tab in Integration server:
https://www.youtube.com/watch?v=KUqzvU6GBug

jauggy:
## Local Dev Tests
You must rerun the fake data task. This is because if you ran it previously, the fake users will have too high permissions. I modified the fakedata task so fake users will have normal permissions.

mix teiserver.fakedata
The create fake data task made by Teifion doesn't give fake users any playtime stats. So we need to add them:

mix teiserver.fake_playtime
Launch the website

mix phx.server
Now go to Admin > Matches > Select a Match > Balance Tab.
You will see the logs for the loser_picks algo.

There is a dropdown with the label "Balancer" near the top. Change this to split_one_chevs
You will now see the logs for split_one_chevs.

#### Testing permissions of normal users
Login as root admin and find a user. Copy their email which should be a guid. Relogin as that user. Check that they cannot see the balance tab.
Relogin as admin and then give that user a contributor/tester role. Relogin as the user and now they should have access.

lib/teiserver/battle/balance/split_one_chevs.ex
```diff
@@ -1,53 +1,99 @@
...
+      new_logs =
+        ["Not enough noobs; calling another balancer.", @splitter, result.logs]
+        |> List.flatten()
+
+      Map.put(result, :logs, new_logs)
+    end
+  end
+
+  @spec has_enough_noobs?([BT.expanded_group()]) :: bool()
+  def has_enough_noobs?(expanded_group) do
```
L-e-x-o-n, line 48: `has_enough_noobs` only checks if there are noobs, not how many?
Is this intended or not?

L-e-x-o-n (commented):
> Teiserver doesn't know the rank shown to the user in Chobby. Chobby gets the rank on login and then never updates. Teiserver, therefore, may classify a user as 2Chev but they might be shown as 1Chev in Chobby.

I thought rank was only calculated on login, when/where else does Teiserver calculate rank?

lib/teiserver/mix_tasks/fake_data.ex
```diff
@@ -96,7 +96,7 @@ defmodule Mix.Tasks.Teiserver.Fakedata do
             name: generate_throwaway_name() |> String.replace(" ", ""),
             email: UUID.uuid1(),
             password: root_user.password,
-            permissions: ["admin.dev.developer"],
```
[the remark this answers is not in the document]
jauggy: So in order to test that a user cannot see the balance tab, they must have zero permissions. That admin.dev.developer permission basically gives them admin role.

lib/teiserver/battle/balance/split_one_chevs.ex
jauggy: I had some ideas like enough noobs be something like 
- At least a single 1chev
- At least two, 2chevs
But decided to keep it simple for now.

lib/teiserver_web/live/battles/match/show.ex
```diff
@@ -115,24 +122,24 @@ defmodule TeiserverWeb.Battle.MatchLive.Show do
...
             |> Map.new(fn userid ->
-              {userid, rating_logs[userid].value["rating_value"]}
+              {userid, rating_logs[userid].value}
             end)
         end)
         |> List.flatten()
 
       past_balance =
-        BalanceLib.create_balance(groups, match.team_count, mode: :loser_picks)
+        BalanceLib.create_balance(groups, match.team_count, algorithm: balancer)
```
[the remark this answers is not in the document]
jauggy: Yeah I can update my word usage to be more consistent. Will look into it.

lib/teiserver/mix_tasks/fake_playtime.ex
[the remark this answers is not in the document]
jauggy: I'll investigate to see if I can do it. This is much more about my Elixir skills not being strong enough to safely modify this code.

jauggy:
> > Teiserver doesn't know the rank shown to the user in Chobby. Chobby gets the rank on login and then never updates. Teiserver, therefore, may classify a user as 2Chev but they might be shown as 1Chev in Chobby.
> 
> I thought rank was only calculated on login, when/where else does Teiserver calculate rank?

To be honest this bug really baffled me. I still have no idea why they are out of sync.

L-e-x-o-n:
> > > Teiserver doesn't know the rank shown to the user in Chobby. Chobby gets the rank on login and then never updates. Teiserver, therefore, may classify a user as 2Chev but they might be shown as 1Chev in Chobby.
> > 
> > 
> > I thought rank was only calculated on login, when/where else does Teiserver calculate rank?
> 
> To be honest this bug really baffled me. I still have no idea why they are out of sync. From my searching it only gets calculated on login.

That was my understanding as well. I checked again, not sure what we are missing...

jauggy:
Currently at least there is an issue for it: https://github.com/beyond-all-reason/teiserver/issues/332
So that the bug is at least recorded.

jauggy:
@L-e-x-o-n I have updated the PR now with the following changes:
- fakedata task now also calls task to add playtime stats
- balancer renamed to be more consistent with other usage

lib/teiserver/battle/libs/balance_lib.ex
```diff
@@ -586,26 +614,39 @@ defmodule Teiserver.Battle.BalanceLib do
...
     # This call will go to db or cache
-    # The cache for ratings is :teiserver_user_stat_cache
+    # The cache for ratings is :teiserver_user_ratings
     # which has an expiry of 60s
     # See application.ex for cache settings
     rating_type_id = MatchRatingLib.rating_type_name_lookup()[rating_type]
-    rating = get_user_balance_rating_value(userid, rating_type_id)
+    {skill, uncertainty} = get_user_rating_value_uncertainty_pair(userid, rating_type_id)
+    rating = calculate_rating_value(skill, uncertainty)
     rating = fuzz_rating(rating, fuzz_multiplier)
```
jauggy, line 623: This is a mistake

PR #339, opened 2024-06-19
Reserve Host[.* when renaming accounts

Beherith:


lib/teiserver/chat/libs/word_lib.ex
```diff
@@ -82,6 +82,7 @@ defmodule Teiserver.Chat.WordLib do
       String.contains?(name, "[lm]") -> true
       String.contains?(name, "[ts]") -> true
       String.contains?(name, "[tei]") -> true
+      String.contains?(name, "Host[") -> true
```
geekingfrog, line 85: Somewhat of a nitpick, I think [starts_with?/2](https://hexdocs.pm/elixir/String.html#starts_with?/2) is more fitted for this. The other checks could also be changed to the same starts_with.

geekingfrog (approved):

PR #342, opened 2024-06-26
Clear cache users between tests

geekingfrog:
This must be done for data cases as well as server cases. Because most of the querying logic rely under the hood on the caches, when someone creates a user, it won't be persisted to the DB after the test, but the cache will not be rolled back.

### DIFF
--- a/lib/teiserver/libs/test_lib.ex
+++ b/lib/teiserver/libs/test_lib.ex
@@ -660,7 +660,12 @@ defmodule Teiserver.TeiserverTestLib do
       :telemetry_simple_client_event_types_cache,
       :telemetry_simple_lobby_event_types_cache,
       :telemetry_simple_match_event_types_cache,
-      :telemetry_simple_server_event_types_cache
+      :telemetry_simple_server_event_types_cache,
+      :users,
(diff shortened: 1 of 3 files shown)


jauggy:
I tested and it worked with this test:
https://github.com/beyond-all-reason/teiserver/blob/f6421ac748e1a215d61959cc9fafea6144717288/test/teiserver/common/clear_db_each_test.exs

Would be better if some sort of test was included in this PR.

geekingfrog:
I've added a test in f543bccb7bd2e738029595369d421cf452d988dd. Not sure of its value but it's something at least.

jauggy:
Yeah I was thinking of the add user test that caused me to find the bug in the first place.

geekingfrog:
Alright, I amended the commit and ended up using your test file. Not the biggest fan, but it's a start and definitely better than nothing.

jauggy (approved):

geekingfrog:
I'm going to hold off merging that, there's something fishy going on with some tests :(
async + global caches => weird shit happening

geekingfrog:
unlikely I'll spend too much time on that since most of it is going to get replaced by tachyon.

Issue #334, opened 2024-06-16
Error on exporting Match Ratings (dev environment)


jauggy:
# Test Steps
Login as root@localhost
Go to Reports > Open Skill > Data Exports > Match Ratings > Generate Report Button

![1](https://github.com/beyond-all-reason/teiserver/assets/1305569/33a8e618-0cdc-46b4-bc06-73f1146903bf)

![2](https://github.com/beyond-all-reason/teiserver/assets/1305569/2797d96c-1b4b-4e3d-b604-6fdb8468968e)

@StanczakDominik is this just a dev environment bug or is this also a bug in integration/prod server?

PR #344, opened 2024-06-30
Fix data export

jauggy:
## Work Done

Fixes https://github.com/beyond-all-reason/teiserver/issues/334
Also allows filtering by Small Team and Large Team when exporting data

## Test Steps

1. Login website as root@localhost (or whatever is the admin)
2. Go to Reports > Data Exports > Match Ratings
3. Press Generate Export. There should be a download with no errors
4. Change Game type to "Small Team" and generate export. There should be a download with no errors
5. Change Game type to "Large Team" and generate export. There should be a download with no errors
6. Repeat with Player Ratings, Ratings Logs, Match Datatable tabs

### DIFF
--- a/lib/teiserver/game/exports/match_ratings_export.ex
+++ b/lib/teiserver/game/exports/match_ratings_export.ex
@@ -91,21 +91,8 @@ defmodule Teiserver.Game.MatchRatingsExport do
 
   # Gets the list of ids but chunks them so we don't try to do too much at once
   defp calculate_match_pages(start_date, end_date, rating_type_id) do
-    query = """
-      SELECT COUNT(id)
-      FROM teiserver_battle_matches
-      WHERE
-        started >= $1
-        AND finished < $2
-        AND rating_type_id = $3
-        AND processed = true
-        AND winning_team IS NOT NULL
-        AND finished IS NOT NULL
-        AND started IS NOT NULL
-    """
-
     match_count =
-      case Ecto.Adapters.SQL.query(Repo, query, [start_date, end_date, rating_type_id]) do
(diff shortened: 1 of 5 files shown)


geekingfrog:
Looks alright to me.
I couldn't test that locally (yet) because there's another error if there is no match at all in the DB.

jauggy:
So before this line:
```
AND rating_type_id = $3
```
would fetch zero rows when the $3 is nil. And we would get the error as shown in #334

Now, it should not error if you actually have matches. If you don't have any matches at all it would error but that would only occur if you didn't run the `mix teiserver.fakedata` locally.

jauggy:
This no longer passes tests so setting to draft. Also it might not be needed because there may other ways to access the data if we set up data dumps.

lib/teiserver/game/exports/match_ratings_export.ex
```diff
@@ -123,6 +110,28 @@ defmodule Teiserver.Game.MatchRatingsExport do
...
+
+    if(rating_type_id == nil) do
+      Ecto.Adapters.SQL.query(Repo, query, [start_date, end_date])
+    else
+      query = query <> " AND rating_type_id = $3"
+
+      Ecto.Adapters.SQL.query(Repo, query, [start_date, end_date, rating_type_id])
+    end
+  end
+
```
L-e-x-o-n, line 134: Only this change is needed now. The rest has been fixed with team split fixes.

L-e-x-o-n (changes requested):
Only query changes are needed, the rest has been fixed with team split fixes.

PR #333, opened 2024-06-15
Minor improvements to whoami and whois commands

jauggy:
## Improvements
1. Add chevron level to `$whois X` command. The purpose of this is to help diagnose issues where teiserver and chobby have a different idea of someone's chev level

2. There's a joke that BAR is a lobby simulator. So let's add lobby hours to `$whoami` and find out for sure if this is the case

## Test Steps
Inside a lobby run 
`$whoami`
`$whois <username>`
Probably best testing this on integration server with real users to see what is saved for lobby hours.

PR #351, opened 2024-07-05
Merge queue attempt 1

geekingfrog:
This is an attempt to create a merge queue to reduce the amount of work required to deploy and test various branches on the integration server.
The assumption is that everything will be fine and we can merge all of that at once, which hopefully reduce the burden on the people having to deploy&test.

Here's a summary of what's in this bundle and a checklist of stuff to check:

# #330 better restriction titles

```
!boss me
$rename 8v8 Glitters Noobs
$setratinglevels 1 20
```
- [ ] The lobby should show "[…] | Rating: 1-20"

# #331 some more icon fixes
Shouldn't need to test anything, worst case, an icon doesn't show

# #333 Improvement to `whoami` and `whois` commands

- [ ] connect and send command `$whoami` and `$whois`, should display chevron level

# #339 Prevent renaming accounts when they start with "[Host"

- [ ] attempt to rename an account as "[Host]hello" and make sure it doesn't work

# #308 make CI great again
Ignore many tests and add a few more dialyzer ignores
- [ ] the CI should pass

# #291 move cache initialisation under supervision trees
There should be no functional change.

- [ ] the server starts alright
- [ ] can login, join a lobby, play a match and exit and nothing explodes server side (this is more a sanity check than anything else)


geekingfrog:
Closing it as all but one of the mentioned PR have been merged.

PR #354, opened 2024-07-10
More lobby tips

jauggy:
## Purpose
Add more lobby tips that will trigger on certain lobby titles. In the future these tips can be removed if we can incorporate it as part of Chobby UI.

Screenshot 2024-07-10 180216

## Test Steps
In a new lobby boss yourself then rename the lobby to contain noob or rotato (or both) e.g.
```
$rename Noob Rotato
```
Or 
```
$rename Noob
$rename Rotato
```

### DIFF
--- a/lib/teiserver/lobby/libs/lobby_restrictions.ex
+++ b/lib/teiserver/lobby/libs/lobby_restrictions.ex
@@ -199,17 +199,7 @@ defmodule Teiserver.Lobby.LobbyRestrictions do
   end
 
   defp get_tips(name) do
-    tips =
-      case is_noob_title?(name) do
-        true -> get_noob_looby_tips()
-        false -> []
-      end
-
-    tips =
-      case is_rotato_title?(name) do
-        true -> tips ++ get_rotato_tips()
-        false -> tips
-      end
+    tips = [] ++ get_noob_looby_tips(name) ++ get_rotato_tips(name)
 
     case length(tips) do
       0 -> nil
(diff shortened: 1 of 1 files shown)


lib/teiserver/lobby/libs/lobby_restrictions.ex
```diff
@@ -278,4 +309,19 @@ defmodule Teiserver.Lobby.LobbyRestrictions do
...
+  def is_rotato_title?(title) do
+    title =
+      title
+      |> String.downcase()
+
+    regex = ~r/\b(rotat)/
+
+    matches =
+      Regex.scan(regex, title)
+      |> Enum.count()
```
geekingfrog, line 323: You can avoid both the downcase and the count with the following:
`Regex.match?(~r/rotat/i, title)`

* The [`i` modifier](https://hexdocs.pm/elixir/Regex.html#module-modifiers) makes the regex case insensitive.
* [Regex.match/2](https://hexdocs.pm/elixir/Regex.html#match?/2) returns what you want, without need for capture group or counting anything.

lib/teiserver/lobby/libs/lobby_restrictions.ex
```diff
@@ -193,23 +193,54 @@ defmodule Teiserver.Lobby.LobbyRestrictions do
...
+      case is_noob_title?(name) do
+        true -> get_noob_looby_tips()
+        false -> []
+      end
+
+    tips =
+      case is_rotato_title?(name) do
+        true -> tips ++ get_rotato_tips()
+        false -> tips
+      end
```
geekingfrog, line 212: I find these repeated shadowing and case a bit awkward and easy to mess up.
I think a nicer approach would be something like:

` tips = [] ++ noob_tips(name) ++ rotato_tips(name)` with these functions doing the check themselves and returning an empty list of that doesn't match.

```elixir
def rotato_tips(name) do
  if Regex.match?(~r/rotat/i, name) do
    # ....
  else
    []
  end
end
```

jauggy:
Nice ideas. I'll come back to this later and incorporate them.

jauggy:
Updated based on @geekingfrog feedback

lib/teiserver/lobby/libs/lobby_restrictions.ex
```diff
@@ -291,37 +293,16 @@ defmodule Teiserver.Lobby.LobbyRestrictions do
   """
   @spec is_noob_title?(String.t()) :: boolean()
   def is_noob_title?(title) do
-    title =
-      title
-      |> String.downcase()
+    anti_noob_regex = ~r/no (noob|newb|nub)/i
+    noob_regex = ~r/\b(noob|newb|nub(s|\b))/i
```
geekingfrog, line 297: what does `\b` do?

geekingfrog (approved):

lib/teiserver/lobby/libs/lobby_restrictions.ex
jauggy: It's a word boundary. So nub or nubs would be a match. But a player named "Anubis" who has their name in the room title would not be a match.

L-e-x-o-n:
Not sure how relevant this PR still is considering the addition of votable Teiserver commands and pending changes to boss mode?

jauggy:
I guess it could be adjusted to use ! instead of $. Setting to draft

jauggy:
Have updated to use ! commands instead of $
```
          "Useful commands for Noob lobbies",
          @splitter,
          "To restrict this lobby to players who are new, use either:",
          "!maxchevlevel <chevlevel>",
          "!maxratinglevel <rating>",
          "",
          "To ensure new players are distributed evenly across teams:",
          "!balancealgorithm split_one_chevs",
          ""
        ]
  ```

PR #357, opened 2024-07-10
Allow Tester role to use $balancemode in unbossed lobby

jauggy:
## Context
In discussions with Sun_Tzu, they mentioned it was difficult to test a new balance algorithm as it requires boss and getting people to vote you boss is nearly impossible. https://discord.com/channels/549281623154229250/1245516077547389009/1260223439067811980

As a short term solution, and to enable us to test balance algorithm this PR will enable someone with Tester role to use `$balancemode` without being boss. However, they can only do this in a non-bossed lobby. It will automatically succeed and not trigger a vote

## Test Steps
1. Launch Lobby and run `$balancemode split_one_chevs`. It should fail
2. Change this user to have Tester role. Login to Chobby again and retry the command. It should succeed.
3. Add another user to the lobby and make that user boss. Try the command again and it should fail.

### DIFF
--- a/lib/teiserver/coordinator/consul_server.ex
+++ b/lib/teiserver/coordinator/consul_server.ex
@@ -27,6 +27,8 @@ defmodule Teiserver.Coordinator.ConsulServer do
 
   @always_allow ~w(status s y n follow joinq leaveq splitlobby afks roll players password? newlobby jazlobby tournament)
   @boss_commands ~w(balancemode gatekeeper welcome-message meme reset-approval rename minchevlevel maxchevlevel resetchevlevels resetratinglevels minratinglevel maxratinglevel setratinglevels)
+  # A tester can use these commands when there is no boss
+  @tester_commands ~w(balancemode)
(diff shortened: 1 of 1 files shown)


L-e-x-o-n:
Code looks good, but I am not sure about the feature itself.
Right now only mods can bypass command restrictions. Lobby members should have an opinion on balance mode changes because it can affect things like parties. They might not want random testers changing it and breaking up parties for example. I think a better solution would be adding `balancemode` as a votable command.

jauggy:
Fair enough. Hopefully there is progress to make commands voteable:
https://ptb.discord.com/channels/549281623154229250/1258404939164680242

L-e-x-o-n:
> Fair enough. Hopefully there is progress to make commands voteable: https://discord.com/channels/549281623154229250/1258404939164680242/1258404942087979139

Progress has been made, `!balancealgorithm <algorithm` has been added.
Closing this PR for now.

PR #361, opened 2024-07-13
Blog Helper + Permission improvements + Toggle contributor icon

jauggy:
## Improvements
1. Contributors can now toggle whether or not they want to show their contributor icon in game. The checkbox is accessible on Contributor page and saves immediately on change.
2. Contributor will now inherit permissions from Tester and Tester will no longer inherit permission from Contributor. Previously Tester would inherit permissions from Contributor, making the Tester potentially more powerful which doesn't make sense. 
3. Added a Blog helper role to coincide with this discord post: https://discord.com/channels/549281623154229250/1243869733988864042/1243869733988864042
The name could be changed to something else but it was called "Blog helper" in the discord post.
4. Posting to blog now requires "Blog helper" permission. Contributors will inherit this permission. Since Tester no longer inherits from Contributor, a Tester can no longer post to blog.

## Other Thoughts
The UI for toggling rank icon is very bare bones (there's no submit button - you just have to toggle it). I am not wanting to do more in this ticket but could be improved in the future.

### Testing contributor rank toggle
Login as a user with Contributor role. Go to my account > Contributor. Tick the checkbox to disable showing rank icon. It will auto save on toggle. Now login to Chobby. Your rank icon will be like a regular user. 

### Testing Blog helper role
Give someone Blog helper role (remove Contributor role if needed). They should be able to create a blog post. Access via Blog > Blog Admin

### Testing Tester role reduced permissions
Remove Blog helper role and give that user Tester role. Try relogin and on the blog page they shouldn't have permission to create a blog post. They still might see a preferences button but this button was already exposed to regular users.

### DIFF
--- a/lib/teiserver_web/live/battles/match/show.ex
+++ b/lib/teiserver_web/live/battles/match/show.ex
@@ -57,7 +57,7 @@ defmodule TeiserverWeb.Battle.MatchLive.Show do
     # Restrict the balance tab to certain roles.
     # Note that Staff roles like "Tester" also contain Contributor role.
     socket
-    |> mount_require_any(["Reviewer", "Contributor"])
+    |> mount_require_any(["Reviewer", "Tester"])
     |> assign(:page_title, "#{match_name} - Balance")
   end
 
--- a/lib/teiserver_web/live/battles/match/show.ex
+++ b/lib/teiserver_web/live/battles/match/show.ex
@@ -55,7 +55,7 @@ defmodule TeiserverWeb.Battle.MatchLive.Show do
 
   defp apply_action(%{assigns: %{match_name: match_name}} = socket, :balance, _params) do
     # Restrict the balance tab to certain roles.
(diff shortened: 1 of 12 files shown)


jauggy:
Now tested with Tester role I cannot see the Blog Admin button. I can see the preferences button but this was already exposed to regular users as evidenced here https://ptb.discord.com/channels/549281623154229250/549281623577722899/1261663425273139291

lib/teiserver_web/live/battles/match/show.ex
```diff
@@ -57,7 +57,7 @@ defmodule TeiserverWeb.Battle.MatchLive.Show do
     # Restrict the balance tab to certain roles.
     # Note that Staff roles like "Tester" also contain Contributor role.
```
L-e-x-o-n, line 58: The comment should be changed now that Contributor contains Tester and not the other way.

jauggy: Have updated in last commit

PR #362, opened 2024-07-14
Add brute force balance algo

jauggy:
## Context
Many people want a balance algorithm that keeps parties together more often. This algorithm goes through every possible combination and picks the best combination. The original idea was from Suuwassea here: https://discord.com/channels/549281623154229250/855772061095559179/1235344149713260584

## Conditions the algorithm will run
1. Team count must equal 2
2. Player count must be <= 16
3. Even number of players

If any of these conditions aren't met, the default balancer will be called instead

## Scoring each combination
Each combination will be given a score as follows:
```
    score = rating_diff_penalty + broken_party_penalty
```
rating_diff_penalty  is just the difference in team ratings
broken_party_penalty is the number of broken parties * broken_party_multiplier
broken_party_multiplier = 3

Pick the combination that scores the lowest.

## Test Steps
Go to to any match on integration server. Preferably find one where parties got split e.g.
https://server5.beyondallreason.info/battle/2092529/players
You can see the parties as they are represented by same coloured dice in the player list. In this case kyutoryu and fbots1998 got split.

Go to balance tab and check brute_force algo. You should see broken party penalty of 0 meaning parties weren't split.

In lobby if you do
```
$balancemode x
```
It will give you a warning and tell you the allowed balance algo. brute_force will not be listed as I have restricted it to mod only.

### DIFF
--- a/lib/teiserver/battle/balance/brute_force.ex
+++ b/lib/teiserver/battle/balance/brute_force.ex
@@ -1,0 +1,251 @@
+defmodule Teiserver.Battle.Balance.BruteForce do
+  @moduledoc """
+  Algorithm idea by Suuwassea
+  Adjusted by jauggy
+
+  Overview:
+  Go through every possible combination and pick the best one. Ideal for keeping parties together.
+  The best combination will have the lowest score.
+  Score = difference in team rating + broken party penalty
+  broken party penalty = num broken parties * broken party multiplier
+
+  Only use for games with two teams and <=16 players
+  """
+  alias Teiserver.Battle.Balance.BalanceTypes, as: BT
+  alias Teiserver.Battle.Balance.BruteForceTypes, as: BF
+  import Teiserver.Helper.NumberHelper, only: [format: 1]
+  require Integer
+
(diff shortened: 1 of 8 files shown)


jauggy:
From my testing on integration server, the time taken for this algo on a 16 player game takes between 20-40ms. Seems longer when there is a large party.

jauggy:
7v7 takes 5 ms
Screenshot 2024-07-18 at 4 50 20 PM

6v6 takes 3 ms
Screenshot 2024-07-18 at 4 51 38 PM

5v5 takes 0.5 
Screenshot 2024-07-18 at 4 53 08 PM
ms

jauggy:
Summary of how long balancer takes to complete
8v8: 20-40ms
7v7: 5ms
6v6: 3ms
5v5: 0.5ms

jauggy:
I'm going to update this so it is restricted to moderator or tester role due to it being slow with 16 players.

jauggy:
Updated so it's no longer selectable by regular players. It exists so that it can be called by `split_noobs`.

PR #363, opened 2024-07-16
Add restriction to number of symbols allowed in a username.

DeviousNull:
Adds an additional check at registration/rename time, which considers how many times each non-alphanumeric character appears in a requested username.

If any individual symbol appears three or more times, then the new username is rejected (and registration / renaming fails).

This patch is intended to restrict unpronounceable "meme names", such as:
- `___`
- `____`
- `_____`
- `[]]][[[]`
... while still allowing names such as:
- `[[CLAN]]First_Middle_Last`

These kinds of names have exceptionally poor recognizability, and if more than one player in a lobby uses these names then it becomes more difficult to 1) discuss and coordinate with them, or 2) identify misbehaving players for reporting/kickbanning.

Notably, all-symbol names with few repeats are still allowed, such as:
- `__`
- `[[]]`
... however, those names are short enough that I suspect identifying them will be less of an issue. If needed, a future patch could also add a requirement to use at least one alphanumeric character in usernames -- but I believe this change alone is sufficient to prevent confusion.

### DIFF
--- a/lib/teiserver/data/cache_user.ex
+++ b/lib/teiserver/data/cache_user.ex
@@ -96,7 +96,13 @@ defmodule Teiserver.CacheUser do
 
   @spec check_symbol_limit(String.t()) :: Boolean.t()
   def check_symbol_limit(name) do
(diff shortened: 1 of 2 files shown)


StanczakDominik (approved):
LGTM, conditional on checks passing (and once we merge this it will no longer require manual approvals for your PRs) I'll happily merge this in the morning - might not be able to deploy tomorrow though. Thanks!

PR #366, opened 2024-07-20
Fix inconsistent "mix test" results based on order of execution

DeviousNull:
Currently, several tests will pass/fail depending on whether they run before or after other tests. This happens because some tests create lobbies but do not remove them, while other tests assume that there are no lobbies when they begin.

For all tests which currently create lobbies, this PR:
- Sets a test-specific lobby name, for easier identification of the source for any "orphaned" lobbies.
- Adds code to each test, which removes any lobbies that the test created.
- Adds an extra `assert` to each test, to ensure that the lobby was removed.

I have used the following command to test this PR:
- mix test --exclude needs_attention --trace --seed 468657

Using the current master branch without these patches, this command results in:
```
Finished in 21.0 seconds (4.9s async, 16.0s sync)
441 tests, 19 failures, 162 excluded

Randomized with seed 468657
```

With these patches, the above command results in:
```
Finished in 70.0 seconds (5.0s async, 65.0s sync)
441 tests, 0 failures, 162 excluded

Randomized with seed 468657
```


test/teiserver_web/live/account/profile/overview_test.exs
```diff
@@ -75,6 +84,10 @@ defmodule TeiserverWeb.Live.Account.Profile.OverviewTest do
       |> render_click()
 
       assert render(view) =~ "Client is not connected"
+
```
geekingfrog, line 87: very small nitpick: I'd put the assert `lobby_id != nil` right after the call to `make_lobby`.

geekingfrog (approved):
Hum, interestingly, running the test command on master doesn't return any error for me.
However I think these changes are good nevertheless, so I'd say ship it.

jauggy:
I get 1 error on master and on this PR.

PR #374, opened 2024-07-23
Add "$set-config-teaser" Host-command. (Prerequisite to fixing #371)

### LINKED ISSUE #371: [Bug]: !rename not working after the first
### Describe the Bug
Recently there has been an updated to SPADS plugins allowing non bossed people to rename a lobby. Try these commands when alone in lobby.
```
!rename test
!rename test2
```

Second rename will fail

### Reproduce the bug

Ensure you have the latest files from: https://github.com/beyond-all-reason/spads_config_bar/pull/134/files
in your local copy of SPADS.

Launch Chobby
```
!rename test
!rename test2
```

Second one will fail. Try bossing yourself
```
!boss jauggy
$rename test
$rename test2
```
I believe it will now pass

### Screenshots

_No response_

### Additional context

I believe that there is a check inside consul_commands.ex
When you use `$rename` as boss it will enter here
```
      senderid != lobby.founder_id ->
        # We have to do this so we don't block the get_state call from the LobbyServer
        # when it tries to query the rating values
        spawn(fn ->
          :timer.sleep(500)
          Battle.rename_lobby(state.lobby_id, new_name, senderid)
        end)
 ```
 
 However, when SPADS tells teiserver to rename that part is not true and instead it goes here:
```
      lobby.player_rename ->
        state

      true ->
        Battle.rename_lobby(state.lobby_id, new_name, nil)

        state
```
The first time it will go to the "true" part and work. The next time it will go to the lobby.player_rename part and do nothing.

Need to understand why Teifion put a timer sleep when renaming the lobby when you do it via `$rename`

### PR DESCRIPTION
**Pull request by DeviousNull**
This command allows Hosts to specify a string that can be used to quickly describe a lobby's currently-configured settings.

The teaser is currently used in the title of a lobby, similar to the existing behavior but without using the "$rename" command. This gives the benefit of not potentially overriding a custom player-set lobby name.

____

This PR is a complement to https://github.com/beyond-all-reason/spads_config_bar/pull/136 , and implements the Teiserver-side command used in that PR. Since BarManager's rename functionality is currently broken, it does not particularly matter which of these two PRs is merged first (no harm will be done if BarManager is using a non-existent command).

This PR specifically **does not** update the `$rename` command, as that would impact player-set lobby names for any SPADS instances that still use the old BarManager code. Once the new plugin code is running on all servers, I can submit a very short PR to update `$rename` and fix #371.

### DIFF
--- a/lib/teiserver/coordinator/consul_commands.ex
+++ b/lib/teiserver/coordinator/consul_commands.ex
@@ -1510,6 +1510,21 @@ defmodule Teiserver.Coordinator.ConsulCommands do
     ConsulServer.say_command(cmd, state)
   end
 
+  def handle_command(%{command: "set-config-teaser", remaining: new_teaser} = cmd, state) do
+    ConsulServer.say_command(cmd, state)
+    new_teaser = String.trim(new_teaser)
+
+    new_teaser =
+      case Regex.run(~r/^[a-zA-Z0-9_\-\[\] \<\>\+\|:]+$/, new_teaser) do
+        [s] -> s
+        _ -> ""
+      end
+
+    Battle.update_lobby_values(state.lobby_id, %{teaser: new_teaser})
+    LobbyLib.cast_lobby(state.lobby_id, :refresh_name)
+    state
+  end
+
   def handle_command(%{command: "rename", remaining: new_name, senderid: senderid} = cmd, state) do
     new_name = String.trim(new_name)
 
@@ -1530,7 +1545,7 @@ defmodule Teiserver.Coordinator.ConsulCommands do
 
     cond do
       new_name == "" ->
-        Battle.rename_lobby(state.lobby_id, lobby.name, nil)
+        Battle.rename_lobby(state.lobby_id, lobby.base_name, nil)
         state
 
       WordLib.flagged_words(new_name) > 0 ->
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Review by jauggy — APPROVED** [review]
Tested locally with latest SPADS plugins (master). No issues other than !rename which you've stated will be fixed later.

File: lib/teiserver/coordinator/consul_commands.ex [hunk 1]
```diff
@@ -1510,6 +1510,21 @@ defmodule Teiserver.Coordinator.ConsulCommands do
     ConsulServer.say_command(cmd, state)
   end
 
+  def handle_command(%{command: "set-config-teaser", remaining: new_teaser} = cmd, state) do
+    ConsulServer.say_command(cmd, state)
+    new_teaser = String.trim(new_teaser)
+
+    new_teaser =
+      case Regex.run(~r/^[a-zA-Z0-9_\-\[\] \<\>\+\|:]+$/, new_teaser) do
```
geekingfrog, line 1518: hum, what?
What is this thing supposed to do? At the very least a comment explaining that would be welcome.
I see that you copied that from the `rename` command right? Perhaps putting that in a module-private function would help clarifying the intent behind this regex.

PR #384, opened 2024-07-30
$playerlimit as Admin only command

L-e-x-o-n:
`$playerlimit <number>` - sets maximum player count limit for current host only, used to change the default 16 player limit.

This PR makes `$playerlimit` command Admin only, first of its type.

### DIFF
--- a/lib/teiserver/coordinator/coordinator_lib.ex
+++ b/lib/teiserver/coordinator/coordinator_lib.ex
@@ -251,7 +251,7 @@ Multiple locks can be engaged at the same time
       :everybody -> true
       :host -> CacheUser.is_moderator?(user) or host
       :moderator -> CacheUser.is_moderator?(user)
-      :admin -> CacheUser.has_any_role?(user, "Admin")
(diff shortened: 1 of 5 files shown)


lib/teiserver/coordinator/coordinator_lib.ex
```diff
@@ -249,6 +251,7 @@ Multiple locks can be engaged at the same time
       :everybody -> true
       :host -> CacheUser.is_moderator?(user) or host
       :moderator -> CacheUser.is_moderator?(user)
+      :admin -> CacheUser.has_any_role?(user, "Admin")
```
geekingfrog, line 254: nitpick: for consistency perhaps have a `is_admin?` function and use that.

geekingfrog (approved):

lib/teiserver/coordinator/coordinator_lib.ex
L-e-x-o-n: Done

PR #386, opened 2024-07-31
Add design doc for tachyon basic systems

geekingfrog:
Better to see the [rendered version](https://github.com/geekingfrog/teiserver/blob/tachyon-doc/documents/dev_designs/tachyon_system.md) on the fork's branch.

This is a high level overview of how I see the lobby/matchmaking/player/autohost interactions in tachyon.

### DIFF
--- a/documents/dev_designs/tachyon_system.md
+++ b/documents/dev_designs/tachyon_system.md
@@ -177,4 +177,44 @@ If a player fail to answer a ready request within a timeout, that player is
 evicted from the queue, and all other players are put back where they were.
 
 
+## Listings
+
+We need a way to list online players + some info, as well as the lobbies. This
+should also support some kind of filtering.
+
+For example:
+* show me all public lobbies from 4v4 and up, with no OS restriction and 1 min chevron.
+* show me all my friends and their status (playing in game, spectating, chilling, in lobby)
+
+This part is the most speculative as of yet. There are basically 3 options:
+
+1) store this data in the DB and query it through SQL.
+  * pros: queries are simple and well understood.
+  * cons: put a limit on scaling based on write capacity of the DB. Need to deal with stale data.
+
+2) store the data in globally unique processes (ETS backed)
+  * pros: simple to implement. No stale data problem (can monitor when entities go away)
+  * cons: potentially requires a lot of memory. Querying may be slow. Single bottleneck.
+
+3) distribute the data across all nodes in their own ETS tables.
+  * pros: scale well. Fairly easy to deal with stale data
+  * cons: querying is a lot more complex, it's more akin to map/reduce.
+
+
+For option 3:
+
+* One local process per node receives updates of local process to update the ETS table of the node.
+  The writer process may be partitionned for throughput.
+
+* Players broadcast relevant updates to the local node's writer(s).
+* The writer creates/updates local ETS store with the data.
+* The writer also monitor the player's session to clear up the ETS store when it goes away.
+
+
+* Querying process call a dedicated query process (can potentially use [poolboy](https://github.com/devinus/poolboy)).
+* Dedicated query process uses [Genserver.multi_call/4](https://hexdocs.pm/elixir/GenServer.html#multi_call/4)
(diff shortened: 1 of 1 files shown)


documents/dev_designs/tachyon_system.md
```diff
@@ -0,0 +1,220 @@
...
+## Listings
+
+We need a way to list online players + some info, as well as the lobbies. This
+should also support some kind of filtering.
+
+For example:
+* show me all public lobbies from 4v4 and up, with no OS restriction and 1 min chevron.
+* show me all my friends and their status (playing in game, spectating, chilling, in lobby)
+
+This part is the most speculative as of yet. There are basically 3 options:
```
p2004a, line 189: I agree it's speculative, we could probably start with option 2 and see to how many messages etc does this part need to scale. For option 1, it depends on database backend e.g. I'm not sure it will do better then option 2 that is purely in memory.

documents/dev_designs/tachyon_system.md
```diff
@@ -0,0 +1,220 @@
...
+
+Autohosts connects to the server and are ready to be summoned to host a game by a lobby.
+Maybe expand tachyon so that they can provide some info about themselves, like region, engine version?
+
+Later, we can expand the autohost system to have some form of capacity scheduling.
+
+## Chatrooms
+Separate processes dedicated to relay messages across a list of recipients (player or autohost).
+Each process has a ring buffer of recent messages to allow replaying messages in case a player
+reconnects after crash.
```
p2004a, line 76: I assume chatrooms are a very generic thing that can be associated with lobby but maybe also for other purposes. I think the interesting part in that is if we ever wanted to support anything like chatroom history, e.g. modeling moderaion messages as DMs via special channel? Also, do people see room history from before they joined?

I'm not sure about any of that.

p2004a (approved):
Nice, thank you!

documents/dev_designs/tachyon_system.md
```diff
@@ -0,0 +1,220 @@
...
+
+    Connection2 ->> Connection2: disconnect
+    Connection2 ->> Session: {:DOWN, _, :process, _, _}
+    Session ->> Session: wait until timeout
+```
+
+## Autohosts
+
+Autohosts connects to the server and are ready to be summoned to host a game by a lobby.
+Maybe expand tachyon so that they can provide some info about themselves, like region, engine version?
```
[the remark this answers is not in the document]
geekingfrog: oh that works too. I thought an autohost would be restricted to one/some versions and the server would need to pick a correct autohost. But if that's not an issue, then we're all good. Just need to send the required engine version.

documents/dev_designs/tachyon_system.md
geekingfrog: You're correct, the chatroom bit isn't attached to lobby. I was thinking of the ability to attach one to a matchmaking queue for example.
Yes, chatroom will have the responsibility to persist durably the messages (for moderation purposes), I'm going to use the same existing code for that.

As for the history, maybe. I was thinking of having only the recent history in memory that can be retrieved, to support a player crashing and reconnecting so they can catch up. Not sure if a new player should be able to see chat history prior to joining, I would err on the side of "no" for that one, but not sure.

documents/dev_designs/tachyon_system.md
```diff
@@ -0,0 +1,220 @@
...
+changes to existing components so we might as well aim for all of them.
+
+## Player and connection
+
+A player connects to teiserver following the [tachyon doc](https://github.com/beyond-all-reason/tachyon/blob/master/docs/connection.md).
+
+Entities:
+
+* `PlayerConnection`: this is the process handling the http connection then upgraded
+  to websocket. It holds the transient state of a player like lobby membership.
```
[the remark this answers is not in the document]
geekingfrog: Yes, that's the intention. I'm going to add a few bits there to clarify how that's supposed to work. I'm not 100% sure myself yet.
But yes, the end goal is that a player that crashed can seamlessly rejoin a lobby.

PR #389, opened 2024-08-01
Get broken party stats per balance algorithm

jauggy:
## Purpose
This PR adds a mix task that gets stats about broken parties. It will get the latest 100 matches for Large Team that have at least one party. Then it will run these balance algos
```
["loser_picks", "cheeky_switcher_smart", "split_noobs", "brute_force"]
```
for each match. Then it will check the number of broken parties and output the results into a text file `results.txt` in the same folder as teiserver root.

It will then repeat for Small Team.

## Other info
Since this is testing `split_noobs`, the code from that [unmerged PR](https://github.com/beyond-all-reason/teiserver/pull/369) is also on this branch.

## Test Steps
On integration server run
```
mix teiserver.party_balance_stats
```
Or 
```
mix teiserver.party_balance_stats /var/log/teiserver/results.txt
```

Then get the `results.txt` file for results


jauggy:
Reminder to me to add check for errors when writing to file

jauggy:
I'm currently working on adjusting this so that it uses 5v5 instead of all variations of Small team.

PR #392, opened 2024-08-01
Remove teaser if lobby title too long

### LINKED ISSUE #391: [Feature]: Remove teaser if lobby title is too long
### Describe the Feature

Have teiserver detect that a lobby title is too long, and if so remove the teaser. The teaser is the bit of information that exists between the title and the rating restrictions. Teasers are given by SPADS.

### Expected Behavior

Check the length of the title with the teaser and rating restrictions, and if too long (not sure the char count but something like 50) then do not display the teaser. See the image attached. The `| Teams 8 vs 8 | ` should be removed if too long.

### Additional context

_No response_

jauggy:
Fix #391 

The teaser is the part of the title between the base name and rating restrictions. Remove if it too long.
Teasers are given by SPADS.

### DIFF
--- a/lib/teiserver/lobby/servers/lobby_server.ex
+++ b/lib/teiserver/lobby/servers/lobby_server.ex
@@ -545,17 +545,34 @@ defmodule Teiserver.Battle.LobbyServer do
           " " <> state.lobby.teaser
       end
 
+    generate_short_name(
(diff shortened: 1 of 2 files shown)


jauggy:
Work is being done in SPADS

Issue #341, opened 2024-06-24
Extend RING with initiator


FIr3baL:
About 11 years ago a nice wrapper for the spring-protocol **RING**-command was introduced to spads.
**!ring** features a nice way to control how often a user can be rung. And users can set their preferred minRingDelay.
It also enables context specific multi-user-rings, e.g "!ring" results in ringing all players, that are not ready or didn't vote yet.

Our ring command travels this way for "client to client"
(Similiar way for multi-rings by spads in other situations)

A user sends the spads command !ring by chat
```
[info] <-- UserInitiator: SAYBATTLE !ring UserToRing
[info] --> UserInitiator: SAIDBATTLE UserInitiator !ring UserToRing
[info] --> SpadsHostUser: SAIDBATTLE UserInitiator !ring UserToRing
[info] --> UserToRing: SAIDBATTLE UserInitiator !ring UserToRing
```
spads decides if the user will be rang and sends a public notification about the ongoing RING.
```
[info] <-- SpadsHostUser: SAYBATTLEEX * Ringing UserToRing (by UserInitiator)
[info] --> SpadsHostUser: SAIDBATTLEEX SpadsHostUser * Ringing UserToRing (by UserInitiator)
[info] --> UserInitiator: SAIDBATTLEEX SpadsHostUser * Ringing UserToRing (by UserInitiator)
[info] --> UserToRing: SAIDBATTLEEX SpadsHostUser * Ringing UserToRing (by UserInitiator)
```
spads sends RING to the rung user
```
[info] <-- SpadsHostUser: RING UserToRing
[info] --> UserToRing: RING SpadsHostUser
```

With the current implementation i see 2 downsides (at least for bar infrastructure):
1. The rung user does not get to know who rang him initialy. RING is always sent from SpadsHostUser and does not include the initiator anymore.
Of course the rung user can read the public message, but this info is not part of the RING command itself. (Because RING is limited by protocol specification)
2. MinRingDelay and other autohost specific ring settings must be set for each spads host individually
(this could be discussed in another issue: e.g add those settings to teiserver's global user prefs)

---
Suggestion for 1.:
Replace teiserver's RING by an extended version, that can only be used by autohosts and allows a second argument:
**The initiator of the current ring.**
This way clients could show, who actually rang them.
(or even decide to show notifications from ignored players or not - this could also be part of 2.)

---
In short:
```
RING <username>
-> RING <username> <initiator>
```
---
[...]

PR #397, opened 2024-08-02
extends RING with a second argument for originator

FIr3baL:
see: https://github.com/beyond-all-reason/teiserver/issues/341

With this PR spads can be extended to send the originator of !ring as a second argument to RING
The second format is only allowed for bot users, so that other clients can not impersonate others.

conventional RING:
```
<-- SpadsHostUser: RING UserToRing
--> UserToRing: RING SpadsHostUser
```

2nd format with this PR:
```
<-- SpadsHostUser: RING UserToRing Originator
--> UserToRing: RING Originator
```

### DIFF
--- a/lib/teiserver/protocols/spring/spring_in.ex
+++ b/lib/teiserver/protocols/spring/spring_in.ex
@@ -1363,29 +1363,22 @@ defmodule Teiserver.Protocols.SpringIn do
   # extended to accept originator as a second argument (only allowed when sent by bots)
   # this allows spads to inform clients who originally rang them with spads-command !ring
   defp do_handle("RING", data, _msg_id, state) do
-    case Regex.run(~r/(\S+) (\S+)/, data) do
-      [_, _, originator] ->
+    userid = CacheUser.get_userid(data)
+
+    case String.split(data) do
+      [_, originator] ->
         client = Client.get_client_by_id(state.userid)
 
-        cond do
-          client == nil ->
-            {:failure, "No client"}
-
-          not CacheUser.is_bot?(state.userid) ->
-            {:failure, "Not a bot"}
-
-          true ->
-            userid = CacheUser.get_userid(data)
-            originatorid = CacheUser.get_userid(originator)
-            CacheUser.ring(userid, originatorid)
-            state
+        if client != nil and not CacheUser.is_bot?(state.userid) do
+          originator_id = CacheUser.get_userid(originator)
+          CacheUser.ring(userid, originator_id)
         end
 
       _ ->
-        userid = CacheUser.get_userid(data)
         CacheUser.ring(userid, state.userid)
-        state
     end
+
+    state
   end
 
   # Not handled catcher
--- a/lib/teiserver/protocols/spring/spring_in.ex
+++ b/lib/teiserver/protocols/spring/spring_in.ex
@@ -1359,10 +1359,33 @@ defmodule Teiserver.Protocols.SpringIn do
     state
   end
 
-  defp do_handle("RING", username, _msg_id, state) do
-    userid = CacheUser.get_userid(username)
-    CacheUser.ring(userid, state.userid)
-    state
+  # https://springrts.com/dl/LobbyProtocol/ProtocolDescription.html#RING:client
+  # extended to accept originator as a second argument (only allowed when sent by bots)
+  # this allows spads to inform clients who originally rang them with spads-command !ring
+  defp do_handle("RING", data, _msg_id, state) do
+    case Regex.run(~r/(\S+) (\S+)/, data) do
+      [_, _, originator] ->
+        client = Client.get_client_by_id(state.userid)
+
+        cond do
+          client == nil ->
+            {:failure, "No client"}
+
+          not CacheUser.is_bot?(state.userid) ->
+            {:failure, "Not a bot"}
+
+          true ->
+            userid = CacheUser.get_userid(data)
+            originatorid = CacheUser.get_userid(originator)
+            CacheUser.ring(userid, originatorid)
+            state
+        end
+
+      _ ->
+        userid = CacheUser.get_userid(data)
+        CacheUser.ring(userid, state.userid)
+        state
+    end
   end
 
   # Not handled catcher
--- a/lib/teiserver/protocols/spring/spring_out.ex
+++ b/lib/teiserver/protocols/spring/spring_out.ex
@@ -72,6 +72,10 @@ defmodule Teiserver.Protocols.SpringOut do
     "DENIED #{reason}\n"
   end
 
+  defp do_reply(:protocol_extensions, nil) do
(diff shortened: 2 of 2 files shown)


jauggy:
How does the extension work? Does it send two commands?

lib/teiserver/protocols/spring/spring_in.ex
```diff
@@ -1359,10 +1359,33 @@ defmodule Teiserver.Protocols.SpringIn do
...
+    case Regex.run(~r/(\S+) (\S+)/, data) do
+      [_, _, originator] ->
+        client = Client.get_client_by_id(state.userid)
+
+        cond do
+          client == nil ->
+            {:failure, "No client"}
+
+          not CacheUser.is_bot?(state.userid) ->
+            {:failure, "Not a bot"}
```
geekingfrog, line 1375: `do_handle` must return the new state. If you return this tuple the caller's state is basically toast.
Since there's no simple mechanism to propagate this error I suggest you just ignore it.
So effectively, you can transform the `cond` into an `if`:

```elixir
if client != nil && not CacheUser.is_bot?(state.userid) do
...
end
```

lib/teiserver/protocols/spring/spring_in.ex
```diff
@@ -1359,10 +1359,33 @@ defmodule Teiserver.Protocols.SpringIn do
...
 
-  defp do_handle("RING", username, _msg_id, state) do
-    userid = CacheUser.get_userid(username)
-    CacheUser.ring(userid, state.userid)
-    state
+  # https://springrts.com/dl/LobbyProtocol/ProtocolDescription.html#RING:client
+  # extended to accept originator as a second argument (only allowed when sent by bots)
+  # this allows spads to inform clients who originally rang them with spads-command !ring
+  defp do_handle("RING", data, _msg_id, state) do
+    case Regex.run(~r/(\S+) (\S+)/, data) do
```
geekingfrog, line 1366: somewhat nitpick, for such a simple regex you could also pattern match on a split:
```elixir
case String.split(data) do
  [_, originator] -> 
    if ... do
      # new version
    end
  _ -> # fallback
end
```

I don't think it matters much since it's not in any hotpath, so the cost of compiling the regex can be ~ignored.

geekingfrog:
> How does the extension work? Does it send two commands? How does it not break SPADS?
> 
> Are you making a PR for SPADS too?

I believe that's what the [code to send the extension](https://github.com/beyond-all-reason/teiserver/pull/397/files#diff-389f925584533e2b41fe917996b7599d3d61b1a3127ddc37a0d4bbb1ea200889R75-R78) is for. This way, spads knows the server support this new command.

Although I couldn't find anything related to this extension mechanism in the [spring lobby protocol spec](https://springrts.com/dl/LobbyProtocol/ProtocolDescription.html), it's probably one of BAR custom thing.

FIr3baL:
> How does the extension work? Does it send two commands? How does it not break SPADS?
> 
> Are you making a PR for SPADS too?

It's not breaking anything, cause it still parses the conventional RING command.
But i had a talk with Yaribz and he said, that it could be implemented into spads, when server sends this @PROTOCOL_EXTENSIONS@ command directly after ACCEPTED.
So, for now, it would change nothing, but it would just enable the 2nd format to be used by bots in future.

lib/teiserver/protocols/spring/spring_in.ex
FIr3baL: Thanks for the review.
I just copied this part from OPENBATTLE above:
https://github.com/beyond-all-reason/teiserver/blob/7cf54198dd2fa13f6ea415860c5870730972458e/lib/teiserver/protocols/spring/spring_in.ex#L810-L815
Does this need to be changed as well ?

FIr3baL:
> > How does the extension work? Does it send two commands? How does it not break SPADS?
> > Are you making a PR for SPADS too?
> 
> I believe that's what the [code to send the extension](https://github.com/beyond-all-reason/teiserver/pull/397/files#diff-389f925584533e2b41fe917996b7599d3d61b1a3127ddc37a0d4bbb1ea200889R75-R78) is for. This way, spads knows the server support this new command.
> 
> Although I couldn't find anything related to this extension mechanism in the [spring lobby protocol spec](https://springrts.com/dl/LobbyProtocol/ProtocolDescription.html), it's probably one of BAR custom thing.

Correct, this new format is starting to exist with this PR, if it gets accepted.

jauggy:
Any documention on @PROTOCOL_EXTENSIONS@ ?

The reason I'm asking is that I want to extend command BATTLEOPENED to have a ranked/unranked flag so we can filter in chobby.

lib/teiserver/protocols/spring/spring_in.ex
geekingfrog: ooooh, I see.
That bit is fine because it's not returning this data directly. It is followed by [a check](https://github.com/beyond-all-reason/teiserver/blob/7cf54198dd2fa13f6ea415860c5870730972458e/lib/teiserver/protocols/spring/spring_in.ex#L851) and if there's a failure, it [replies with the correct error].

So if there's such a reply for the RING command, you should reply with that one.

Yaribz:
This @PROTOCOL_EXTENSIONS@ system is a way for lobby servers to declare they support optional protocol extensions, so that lobby clients know they can use them. It just consists in a standard `SERVERMSG` command sent by the lobby server upon successful login just after the `ACCEPTED` command. This server message must start with the `@PROTOCOL_EXTENSIONS@` prefix, followed by a space, then a JSON encoded dictionary / hash table whose fields are the extension names and the values are the version supported of these extensions. This is not part of the official Spring lobby protocol, as this is used to add optional extensions to this protocol.

This must not be confused with the compatibility flag system included in the official Spring lobby protocol, which is sort of the opposite: a way for lobby clients to declare that they support new commands/syntax, so that the lobby servers know that they can send these new commands/syntax to them (refer to the `compFlags` parameter of the [LOGIN](https://springrts.com/dl/LobbyProtocol/ProtocolDescription.html#LOGIN:client) command, and the [Compatibility Flags section](https://springrts.com/dl/LobbyProtocol/ProtocolDescription.html#Changelog) of the Spring lobby protocol description).

> The reason I'm asking is that I want to extend command BATTLEOPENED to have a ranked/unranked flag so we can filter in chobby.

This requires sending new commands/syntax to lobby clients which support them only, so this means you need to use the compatibility flag system for that, not the @PROTOCOL_EXTENSIONS@ system.

lib/teiserver/protocols/spring/spring_in.ex
L-e-x-o-n: Should be fixed

lib/teiserver/protocols/spring/spring_in.ex
L-e-x-o-n: Should be fixed.

geekingfrog (approved):

Issue #399, opened 2024-08-04
[Feature]: Make webchat read only for most people


jauggy:
### Describe the Feature

Make webchat read only for most people. Currently anyone can use webchat from website including spammers or people who have been banned from the lobby.

### Expected Behavior

Login to website. Go to lobbies. Select Chat. The Send button and text input should only be visible for mods/admins. It should not be visible for academy managers and mentors. This will allow normal users to read chat but not send messages.

Allowing contributors access is optional. I don't really see a use case for them using it.

### Screenshots

![1](https://github.com/user-attachments/assets/b3e12293-72d8-4e36-8489-2541184b9df7)

### Additional context

Discord discussion: https://discord.com/channels/549281623154229250/1269606132024868986

L-e-x-o-n:
Resolved with #401

PR #401, opened 2024-08-04
Make web-based lobby chat read-only for non-Admin/Moderator users.

DeviousNull:
Currently, any non-Muted user is allowed to read and send lobby chat messages via the website.

However, the use-case for this feature is not very strong for the typical player, and we occasionally see it abused by players who have been kickbanned from the lobby (to flame anyone involved in the kickban).

Therefore, this patch restricts web-based chat such that only users with either the "Admin" or "Moderator" roles are able to send messages.
____

Implements #399.

Screenshots from local testing:

- Users with `Admin`/`Moderator` roles can chat:

- Users without those roles can only read chat:

To test:

1. Log in as an `Admin`/`Moderator` user
2. Browse to <TEISERVER_HOSTNAME>/battle/lobbies and click "CHAT" for any lobby
3. Observe that the chat bar and Send button are shown
4. Repeat steps 1-3 with a non-`Admin`/`Moderator` user, and observe that the chat bar and Send button are not shown

### DIFF
--- a/lib/teiserver_web/live/battles/lobbies/chat.ex
+++ b/lib/teiserver_web/live/battles/lobbies/chat.ex
@@ -61,10 +61,8 @@ defmodule TeiserverWeb.Battle.LobbyLive.Chat do
 
       true ->
         allowed_to_send =
-          CacheUser.has_any_role?(current_user.id, [
-            "Admin",
-            "Moderator"
-          ]) and not CacheUser.has_mute?(current_user.id)
+          CacheUser.allow?(current_user.id, "Moderator") and
+            not CacheUser.has_mute?(current_user.id)
 
         :timer.send_interval(10_000, :tick)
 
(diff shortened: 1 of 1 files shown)


lib/teiserver_web/live/battles/lobbies/chat.ex
```diff
@@ -60,7 +60,11 @@ defmodule TeiserverWeb.Battle.LobbyLive.Chat do
         index_redirect(socket)
 
       true ->
-        allowed_to_send = not CacheUser.has_mute?(current_user.id)
+        allowed_to_send =
+          CacheUser.has_any_role?(current_user.id, [
+            "Admin",
+            "Moderator"
+          ]) and not CacheUser.has_mute?(current_user.id)
```
L-e-x-o-n, line 67: Admin contains moderator. In theory only `CacheUser.is_moderator?` should be enough

L-e-x-o-n (approved):

PR #406, opened 2024-08-12
Log email errors

L-e-x-o-n:
Logs verification and password reset E-mails sent as complex server events. Both successful and failed events are logged including error in case of failure.


geekingfrog (commented):
Code looks fine, minus the compiler warnings.
But I'm a little bit cautious about logging the user's email address.
Because that's one thing we would need to purge in case they request a gdpr "erasure", and I'm not sure how that would work if it's in our logs.

What's the drive behind this change? Would the user id be enough?

L-e-x-o-n:
> Code looks fine, minus the compiler warnings. But I'm a little bit cautious about logging the user's email address. Because that's one thing we would need to purge in case they request a gdpr "erasure", and I'm not sure how that would work if it's in our logs.
> 
> What's the drive behind this change? Would the user id be enough?

That's a good point about GDPR:
The reason for logging these are all the user reports of not receiving verification codes.
Email is not strictly necessary, but it might be useful as a clue to why sending email failed. Email can be changed so just userid might not be enough.

L-e-x-o-n:
Removed email from events due to GDPR concerns. It's a simpler solution than cleaning up all events on GDPR clean. If it turns out we really need email in the future, it can be added back and GDPR clean updated.

PR #408, opened 2024-08-13
Fix and improve relationships admin view

L-e-x-o-n:
I noticed user admin relationships page wasn't working.
This fixes it and expands it with a list of users the selected user is following, blocking, avoiding, avoided by, blocked by.

### DIFF
--- a/lib/teiserver_web/controllers/admin/user_controller.ex
+++ b/lib/teiserver_web/controllers/admin/user_controller.ex
@@ -952,9 +952,23 @@ defmodule TeiserverWeb.Admin.UserController do
   @spec relationships(Plug.Conn.t(), map) :: Plug.Conn.t()
   def relationships(conn, %{"id" => id}) do
     user = Account.get_user!(id)
+    id = user.id
+
+    relationships = %{
+      friends: Account.list_friend_ids_of_user(id),
+      friend_requests: Account.list_incoming_friend_requests_of_userid(id),
+      followed: Account.list_userids_followed_by_userid(id),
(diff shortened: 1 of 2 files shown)


lib/teiserver_web/controllers/admin/user_controller.ex
```diff
@@ -953,17 +953,37 @@ defmodule TeiserverWeb.Admin.UserController do
...
+
     user_ids =
-      (user.data["friends"] ++ user.data["friend_requests"] ++ user.data["ignored"])
+      relationships
+      |> Map.values()
+      |> List.flatten()
       |> Enum.uniq()
 
+    user_ids
+    |> IO.inspect(limit: :infinity)
```
Teifion, line 974: I suspect you meant to remove this, there's one a few lines later on 981 too.

L-e-x-o-n: Yea this is still draft, work in progress.

lib/teiserver_web/controllers/admin/user_controller.ex
```diff
@@ -952,9 +952,23 @@ defmodule TeiserverWeb.Admin.UserController do
   @spec relationships(Plug.Conn.t(), map) :: Plug.Conn.t()
   def relationships(conn, %{"id" => id}) do
     user = Account.get_user!(id)
+    id = user.id
```
L-e-x-o-n, line 955: Special thanks to @geekingfrog for fixing this bug.
Id was a string, not an integer.

PR #412, opened 2024-08-15
Display *all* the friends (and avoid and other)

geekingfrog:
something I missed for https://github.com/beyond-all-reason/teiserver/pull/408 is that the [default user query](https://github.com/geekingfrog/teiserver/blob/348972752fe01e876adefa723c16fbff29054c49/lib/teiserver/account/queries/user_queries.ex#L18) has a limit of 50. So for people like Lexon with a ton of friends, that page would only show up to 50 users total.


L-e-x-o-n:
Great, thanks

PR #414, opened 2024-08-16
Add balancealgorithm command

jauggy:
## Context
Currently SPADS has a command balancealgorithm. It is calling teiserver command balancemode. For consistency we should add a teiserver command called `balancealgorithm`. This PR adds balancealgorithm command that does exactly the same as balancemode but it can be called by SPADS one to one.

## Other enhancements
When you give an invalid command parameter via `!balancealgorithm blah`, the error isn't displayed in lobby. So Lobby.sayprivateex has changed to ChatLib.say

For example currently in production if I use `!balancealgorithm split_one_chevs` no error is visible (but it should show that it is not a valid balance algo).

## Test Steps
Set yourself as boss. Then run these commands
```
$balancealgorithm split_noobs
$balancemode loser_picks
$balancealgorithm x
```

### DIFF
--- a/lib/teiserver/coordinator/consul_commands.ex
+++ b/lib/teiserver/coordinator/consul_commands.ex
@@ -629,7 +629,18 @@ defmodule Teiserver.Coordinator.ConsulCommands do
   end
 
   #################### Boss
+  # balancemode is deprecated. Use balancealgorithm instead, which is also a SPADS command
   def handle_command(%{command: "balancemode", remaining: remaining, senderid: senderid}, state) do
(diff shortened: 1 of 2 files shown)


L-e-x-o-n:
balancemode should be removed completely imo.

L-e-x-o-n:
> It's used by SPADS at the moment. I can remove balancemode in a seperate PR once SPADS is using balancealgorithm.

Ah, I see. Please let me know once SPADS is using `balancealgorithm`. I would like to clean up unnecessary code where possible.

PR #415, opened 2024-08-16
Tachyon dev conveniences

geekingfrog:
Some utilities to tie up some lose ends with oauth stuff and make developing tachyon easier.
* a mix task to generate an oauth token for a given user
* extend fakedata to seed a basic oauth application
* a cleanup task to delete from DB the expired codes and tokens

### DIFF
--- a/lib/teiserver/o_auth/tasks/gen_token.ex
+++ b/lib/teiserver/o_auth/tasks/gen_token.ex
@@ -1,0 +1,63 @@
+defmodule Teiserver.OAuth.Task.GenToken do
+  @moduledoc """
+  Generates a token for the given user. This is for a convenience task to ease
+  the development of anything requiring OAuth tokens like tachyon protocol.
+  """
+
+  alias Teiserver.OAuth.{ApplicationQueries, Token}
+
+  @spec create_token(String.t(), String.t() | nil) :: {:ok, Token.t()} | {:error, term()}
+  def create_token(username_or_email, app_uid \\ nil) do
+    with {:ok, user} <- get_user(username_or_email),
+         {:ok, app} <- get_app(app_uid) do
+      token_attr = %{
+        value: Base.hex_encode32(:crypto.strong_rand_bytes(32), padding: false),
+        owner_id: user.id,
+        application_id: app.id,
+        scopes: app.scopes,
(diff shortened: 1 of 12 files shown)


lib/teiserver/o_auth/tasks/gen_token.ex
```diff
@@ -0,0 +1,65 @@
...
+      |> Token.changeset(token_attr)
+      |> Teiserver.Repo.insert()
+    else
+      {:error, msg} -> {:error, msg}
+    end
+  end
+
+  defp get_user(username_or_email) do
+    user =
+      if String.contains?(username_or_email, "@") do
```
L-e-x-o-n, line 33: Why not use `UserCacheLib.get_user_by_name` and `UserCacheLib.get_user_by_email`?

test/support/fixtures/tachyon.ex
```diff
@@ -0,0 +1,2 @@
+defmodule Teiserver.TachyonFixtures do
```
L-e-x-o-n, line 1: This is for future use?

test/teiserver/o_auth/token_test.exs
```diff
@@ -76,4 +77,30 @@ defmodule Teiserver.OAuth.TokenTest do
     assert {:ok, new_token_fresh} = OAuth.get_valid_token(new_token.value)
     assert new_token_fresh.id == new_token.id
   end
+
+  test "can delete expired token", %{user: user, app: app} do
+    assert {:ok, token, attr} =
```
L-e-x-o-n, line 82: Not a big deal, `attr` unused

test/support/fixtures/tachyon.ex
geekingfrog: ooops, that's something I have in the pipes and should not have been commited, removed.

test/teiserver/o_auth/token_test.exs
geekingfrog: amended.

lib/teiserver/o_auth/tasks/gen_token.ex
geekingfrog: Originally I had issues because the caches were not setup. But it seems that it's fine now, not sure how though.
Anyway, I replaced the queries with the higher level ones using cache.

PR #416, opened 2024-08-17
Turn off boss blocks after playercount + queue size threshold reached

jauggy:
## Context
Some people want boss blocks off. Some people want them on. This is a compromise. It will guarantee your boss blocks work for your very first match, but not guaranteed afterwards if you lobby gets popular.

## Feature
Once a certain player count + queue size is reached (set to 16) then boss blocks are turned off. If you start a lobby from scratch, you might have to wait a long time for it to fill, so it's good to guarantee you won't have to play with your avoid list for your first match. However, after the first match, it will no longer be guaranteed since your lobby might hit these thresholds.

The reason we consider queue size is so that if you set it to 7v7 you can still hit the threshold if there's two people on your queue.

## Other enhancements
There was a check after a match to recheck blocks/avoids. This has been removed. 
```
      send(self(), :recheck_membership)
```
So once a user makes it as a player, even if the lobby falls below the threshold, they are safe.

## Test Steps
Requires 2 players.
Have player 1 join a lobby and boss themselves.  Have player 2 try and join the lobby. Have player 1 block player 2 then have player 2 try and leave and join the lobby; they will be blocked.

Now go to admin site config > Lobbies . Change the setting below to 0

Now have player 2 try and join again. They may still get blocked due to hitting the player count trigger block; but they will not receive the boss block message.

### DIFF
--- a/lib/teiserver/coordinator/consul_server.ex
+++ b/lib/teiserver/coordinator/consul_server.ex
@@ -172,8 +172,6 @@ defmodule Teiserver.Coordinator.ConsulServer do
         name = Account.get_username_by_id(userid)
         Coordinator.send_to_host(state.coordinator_id, state.lobby_id, "!mute #{name}")
       end
-
-      send(self(), :recheck_membership)
     end)
 
     {:noreply, %{state | timeouts: %{}}}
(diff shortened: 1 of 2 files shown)


geekingfrog (approved):
I would amend the commit (`git commit --amend`) and add most of your PR description as commit comment.

After the first commit line, you skip a line, and then you can add more context.
For single commit like that, achieving a very specific result, it's a great way to document the intent of a change. It is easier to find commit messages when looking at the code than tying a specific change to a potential pull request.

The section "context" and "feature" can be copy/pasted perhaps with a slight reword as commit message.

jauggy:
Closing PR as it doesn't have enough support. I think they are in favour of just turning off boss blocks due to drama.

PR #420, opened 2024-08-20
Add welcome text when using custom balance algo

jauggy:
## Context
There's currently no visual indicator when using a custom balance algorithm. Not everyone knows about `$s`. Ideally it should show in a dropdown in Chobby, but given Chobby is end of life it might not come. As a workaround show it in welcome message.

## Improvement
Screenshot 2024-08-20 125613

## Other enhancements
Minor change to remove splitter from when you don't meet rating restrictions because this should appear directly below welcome message resulting in two splitters back to back which looks weird.
Screenshot 2024-08-20 100853

Also standardise splitter for explain command

## Testing
Login as a user and boss yourself. Then change balance algo
```
$balancemode split_noobs
```

Join as another user and you will see a message about the custom balance algorithim.

### DIFF
--- a/lib/teiserver/coordinator/consul_server.ex
+++ b/lib/teiserver/coordinator/consul_server.ex
@@ -23,7 +23,14 @@ defmodule Teiserver.Coordinator.ConsulServer do
   alias Phoenix.PubSub
   alias Teiserver.Battle.BalanceLib
   alias Teiserver.Data.Types, as: T
-  alias Teiserver.Coordinator.{ConsulCommands, CoordinatorLib, SpadsParser, CoordinatorCommands}
+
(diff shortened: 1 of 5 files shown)


L-e-x-o-n:
I think knowing the balance algorithm is not important enough for the players to be mentioned in welcome message.  same for the $explain command. Those who know what balance algorithms exist and what they do will know where to check for them and where to find the explain logs. Telling everyone who joins that current algorithm might not be very informational when they don't know how it works.

jauggy:
I do agree. I think I will put balance algo information here instead:
https://server4.beyondallreason.info/battle/lobbies/show/8919
I think most normal users will not use that page often but it's useful for me to debug stuff and examine how algos are used in the wild.

PR #421, opened 2024-08-22
Add respect_avoids balancer

jauggy:
## Context
Would people use a balancer that respect avoids if it ignores parties? I have evidence to suggest yes:
https://discord.com/channels/549281623154229250/549281623577722899/1274512773061017704

This balancer acts like an anti-party balance algorithm

## How to run?
```
$balancemode respect_avoids
```
To get it voteable we will need to make some changes to spads_config_bar

## Technical Overview
This takes split_noobs and copies it but changes the scoring function.

### brute_force_avoids.ex
This file is a copy of brute_force only it is more like a lib (it cannot be called by players) and changes the scoring function. It goes through all combinations and finds the combination with the lowest score.

Score = rating diff penalty + broken avoid penalty
Rating diff penalty = difference in team ratings
Broken avoid penalty = avoid importance * number of broken avoids

Avoids are a list of player ids like [[1,2][3,4]]. The avoid is broken if the two players are on the same team. It assumes there are exactly two teams.

### respect_avoids.ex
This file is a copy of split_noobs but it calls brute_force_avoids instead. Just like split_noobs, we limit the number of players to 14 that will be fed into the brute force algo. For 16 player lobbies, two players will be drafted and won't be part of the brute force.

Just like split_noobs, noobs will be spread evenly across teams. To guarantee that, you cannot avoid noobs.

If there are no relevant avoids in the lobby, split_noobs or loser_picks will be called instead (depending on if there are newish players or not). That means we do support parties if there are no relevant avoids.

### relationship_lib.ex
There's a couple of new functions that allows you to get a list of avoids/blocks given a list of player ids. One of the functions allows getting avoids that are at least x hours old. 

## Ranked versus unranked
For ranked games, avoids will only be respected if they are at least 24 hours old. This is to prevent lobby manipulation in real time. However for unranked games and debugging via the website, avoids will work immediately.

## Changes to boss avoids
Boss avoids used to block you from playing. However, we expect avoids to rise due this algo. Therefore I've changed it so that boss avoids do nothing now. Boss blocks will still work.

I have also removed rechecking of avoids/blocks after match ends. If you make it as a player you are always safe.

I have also removed checking of behaviour_score when assessing avoids/blocks. It's a little risky if it's left since we expect avoids to rise.

## Cleanup
cheeky_switcher_smart and split_one_chevs code has been removed

## Testing
You can test via the website using the balance tab. Try viewing a past match and checking players in your team, then avoid one of them and see how the balance changes.


test/teiserver_web/controllers/api/spads_controller_test.exs
```diff
@@ -1,30 +1,13 @@
 defmodule TeiserverWeb.API.SpadsControllerTest do
   use TeiserverWeb.ConnCase, async: false
```
jauggy, line 2: These are removal of unused imports/functions due to dialyzer complaining

lib/teiserver/account/libs/relationship_lib.ex
```diff
@@ -298,23 +299,12 @@ defmodule Teiserver.Account.RelationshipLib do
...
-
-        user.behaviour_score <= 8000 ->
-          {3, 30}
-
-        true ->
-          {
-            Config.get_site_config_cache("lobby.Block count to prevent join"),
-            Config.get_site_config_cache("lobby.Block percentage to prevent join")
-          }
-      end
```
L-e-x-o-n, line 317: This doesn't seem relevant to for the avoid balancer. Should be done as a different PR, but IMO it's already fine as is.

jauggy:
Closing for now. I will separate out the cleanup.

Buttons wants to try writing an API that I can call that will try and respect avoids and parties. https://discord.com/channels/549281623154229250/1276469277297217536/1276495379436797992

I will investigate the feasibility of limiting people's avoids. Maybe as a first step I will write a script to see what is the average amount of avoids/blocks people have out of people who have them and that could be our target.

PR #429, opened 2024-08-26
default balance algo determine whether to pick loser_picks or split_noobs depending on circumstances

jauggy:
## Feature
This PR creates a new balance algo called "default" which becomes the default. It will call 

If team size != 2 then call loser_picks
Otherwise if at least one newish player call split_noobs
Otherwise call loser_picks

Finally complaints over years where people say that new players are overrated can be put to rest. I have been observing multiple noob lobby games with split_noobs and am happy with the results (other than the chev icon being wrong for an unknown reason sometimes).

In SPADS we would later need to make a change to make loser_picks not voteable, but instead make default voteable instead.

## Testing
You can go to website and pick default from the balance dropdown when inspecting a match

### DIFF
--- a/lib/teiserver/battle/balance/default_balance.ex
+++ b/lib/teiserver/battle/balance/default_balance.ex
@@ -1,0 +1,81 @@
+defmodule Teiserver.Battle.Balance.DefaultBalance do
+  @moduledoc """
+  This will call other balancers depending on circumstances
+  """
+  alias Teiserver.Battle.Balance.SplitNoobs
+  alias Teiserver.Battle.Balance.LoserPicks
+  alias Teiserver.Battle.Balance.SplitNoobsTypes, as: SN
+  alias Teiserver.Battle.Balance.BalanceTypes, as: BT
+
+  @doc """
+  Main entry point used by balance_lib
+  """
+  @spec perform([BT.expanded_group()], non_neg_integer(), list()) :: any()
+  def perform(expanded_group, team_count, opts \\ []) do
+    case should_use_algo(expanded_group, team_count) do
+      "split_noobs" ->
+        SplitNoobs.perform(expanded_group, team_count, opts)
+
+      _ ->
+        LoserPicks.perform(expanded_group, team_count, opts)
+    end
+  end
+
+  @spec should_use_algo([BT.expanded_group()], integer()) ::
+          String.t()
+  def should_use_algo(expanded_group, team_count) do
+    cond do
+      team_count != 2 ->
+        "loser_picks"
+
+      true ->
+        players = flatten_members(expanded_group)
+        has_noobs? = has_noobs?(players)
+
+        cond do
+          has_noobs? -> "split_noobs"
+          true -> "loser_picks"
+        end
+    end
+  end
+
+  @doc """
+  Converts the input to a simple list of players
+  """
+  @spec flatten_members([BT.expanded_group()]) :: any()
+  def flatten_members(expanded_group) do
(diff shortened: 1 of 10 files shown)


lib/teiserver/battle/balance/default_balance.ex
```diff
@@ -0,0 +1,79 @@
...
+          rating: rating,
+          name: name,
+          id: id,
+          uncertainty: uncertainty,
+          rank: rank,
+          in_party?:
+            cond do
+              count <= 1 -> false
+              true -> true
+            end
```
geekingfrog, line 69: you can directly do `in_party?: count > 1`

lib/teiserver/battle/balance/default_balance.ex
```diff
@@ -0,0 +1,79 @@
...
+          has_noobs? -> "split_noobs"
+          true -> "loser_picks"
+        end
+    end
+  end
+
+  @doc """
+  Converts the input to a simple list of players
+  """
+  @spec flatten_members([BT.expanded_group()]) :: any()
```
geekingfrog, line 45: It would be nice to type the return as well

lib/teiserver/battle/balance/default_balance.ex
```diff
@@ -0,0 +1,79 @@
...
+          ratings: ratings,
+          ranks: ranks,
+          names: names,
+          uncertainties: uncertainties,
+          count: count
+        } <- expanded_group,
+        # Zipping will create binary tuples from 2 lists
+        {id, rating, rank, name, uncertainty} <-
+          Enum.zip([members, ratings, ranks, names, uncertainties]),
+        # Create result value
```
geekingfrog, line 58: Useless comment

lib/teiserver/battle/balance/default_balance.ex
```diff
@@ -0,0 +1,79 @@
...
+        SplitNoobs.perform(expanded_group, team_count, opts)
+
+      _ ->
+        LoserPicks.perform(expanded_group, team_count, opts)
+    end
+  end
+
+  @spec should_use_algo([BT.expanded_group()], integer()) ::
+          String.t()
+  def should_use_algo(expanded_group, team_count) do
```
geekingfrog, line 26: I would refactor this function into something like `balance_module`, and instead of returning a string, you can directly return the module. Then the caller can do something like
```elixir
balance_module(expanded_group, team_count, opts).perform(expanded_group, team_count, opts)
```

This is a nice approach to define an interface. If you feel like it you can also define a [behaviour](https://hexdocs.pm/elixir/typespecs.html#behaviours) for the balance modules.

lib/teiserver/battle/libs/balance_lib.ex
```diff
@@ -27,7 +27,7 @@ defmodule Teiserver.Battle.BalanceLib do
   # which one will get to pick first
   @shuffle_first_pick true
 
-  @default_balance_algorithm "loser_picks"
+  @default_balance_algorithm "default"
```
geekingfrog, line 30: this doesn't make sense anymore to keep this as a hardcoded value. Especially since there is already `get_default_algorithm/1` a bit below. You can modify this function to take some inputs and return the correct module.
This way you keep a clean separation between "choosing the algorithm" and "running it". Right now it's a bit muddy.

jauggy:
Updated based on feedback.

@L-e-x-o-n as requested the default balance algorithm is now in admin config. Have set to "loser_picks" but can change to "default".

PR #435, opened 2024-08-30
Respect avoids balancer with party supports

jauggy:
## Summary
This balancer respects parties and avoids. It does this by setting importance of different things:

  @max_team_diff max(10, 5% of team rating) (This means we break parties/avoids when the diff gets larger than this).
  @max_team_diff_importance 10000
  @party_importance 1000
  @avoid_importance 10

Go through each combination and give it a score

>   Score = team rating diff penalty + broken avoid penalty + broken party penalty + broken max team rating diff penalty
>   team rating diff penalty = difference in team ratings
>   broken avoid penalty = broken avoid count * @avoid_importance
>   broken party penalty = broken party count * @party_importance
>   broken max team rating diff penalty = @max_team_diff_importance (if the team rating diff > @max_team_diff)

We choose the combination with the lowest score.

## Performance
For 8v8 games we only brute force 14 players max and two remaining will be drafted and don't affect score in any way. This is for performance reasons.

## Per lobby avoid limits
To reduce computation of too many avoids, we limit the number of avoids that will be pulled from the database depending on player count

**7v7 or more** -> Max 7 avoids per lobby. However, for every two players in party, the number of avoids will be reduced by one (to a minimum of one). 
**6v6 or less** -> No limit. But there is a per player avoid limit.

## Per player avoid limits
In addition to a lobby avoid limit, there is a per player avoid limit (for all team sizes). We only pull the two oldest relevant avoids from the database per player. So if you are avoiding a,b,c,d and b,c,d is in lobby, then we pull only b,c from the database.

## Testing
On integration server just view a match and then the balance tab. Select respect_avoids. Add some avoids and see the teams change.


jauggy:
Need to test on integration server to get the actual performance.

lib/teiserver/battle/balance/brute_force_avoid.ex
```diff
@@ -0,0 +1,161 @@
...
+
+  @spec score_combo([BF.player()], [BF.player()], [[number()]], [[number()]]) :: any()
+  def score_combo(first_team, all_players, avoids, parties) do
+    first_team_rating = get_team_rating(first_team)
+    both_team_rating = get_team_rating(all_players)
+    rating_diff_penalty = abs(both_team_rating - first_team_rating * 2)
+
+    max_team_diff_penalty =
+      cond do
+        rating_diff_penalty > @max_team_diff -> @max_team_diff_importance
```
p2004a, line 72: Why use some hard coded value instead of open skill win prediction? Seems to me that hardcoded value is very binary and having a single value for all team sizes also sounds off.

lib/teiserver/battle/balance/brute_force_avoid.ex
```diff
@@ -0,0 +1,161 @@
...
+        true -> 0
+      end
+
+    # If max_team_diff_penalty is non zero don't even bother calculating avoid and party penalty
+    # since it's likely we'll discard this combo
+    {broken_avoid_penalty, broken_party_penalty} =
+      case max_team_diff_penalty do
+        0 ->
+          {count_broken_avoids(first_team, avoids) * @avoid_importance,
+           count_broken_parties(first_team, parties) * @party_importance}
```
p2004a, line 82: The more parties, the higher importance of this constraint. Did you consider dividing this by number of parties to get float 0-1 and only multiply that with importance factor? This way overall concept of parties has max bounded importance, same for avoids

lib/teiserver/battle/balance/brute_force_avoid.ex
jauggy: We don't have access to win prediction in Elixir lib. Hence why I asked on discord for this:
https://ptb.discord.com/channels/549281623154229250/855772061095559179/1276661091346153533

lib/teiserver/battle/balance/brute_force_avoid.ex
jauggy: Hmm let me rethink this one.

lib/teiserver/battle/balance/brute_force_avoid.ex
p2004a: Calling API to get win prediction as part of balancing when you are checking _many_ combinations and need score for each: very bad idea.

https://github.com/vivekjoshy/openskill.py/blob/main/openskill/models/weng_lin/plackett_luce.py#L846-L858 looks very easy to port.

I'm not asking about simply dividing the sum. But also what you do at the moment, it's arbitrary cutoff: < 17: all good max score, > 17: 0 score. For example with this as far as I understand the core, balancing [[24,24], [15,18]] will be just fine and get max score, that sounds very wrong to me.

jauggy: I'll take a look at if we can port it.

jauggy:
Here are some local tests. Here is the match with several parties


Output with 1 avoid


Now I'm going to avoid three more people. I am logged in as root and will now avoid
EnergisedBeigeFiend SpryLavenderAugur SpringyRoseMarauder
Afterwards

lib/teiserver/battle/balance/brute_force_avoid.ex
jauggy: So there is a max bound due to these parameters
```
 @max_team_diff 14
  @max_team_diff_importance 10000
  ```
  If the rating diff is greater than @max_team_diff then parties and avoids will not be respected as those have loser importance

**Comment by jauggy** [comment]
Updated with the following changes based on feedback:
- People avoiding you will no longer prevent you from playing
- Ranked/unranked/lobby type do not have effect on balancer

File: lib/teiserver/battle/balance/brute_force_avoid.ex (hunk 1 above)
Comment (jauggy) [reply lib/teiserver/battle/balance/brute_force_avoid.ex:72]: The addition of win probability to Elixir library will be done in a seperate PR. This PR will just calculate team ratings as before.

File: lib/teiserver/account.ex [hunk 3]
```diff
@@ -1570,9 +1570,6 @@ defmodule Teiserver.Account do
   @spec check_block_status(T.userid(), [T.userid()]) :: :ok | :blocking | :blocked
   defdelegate check_block_status(userid, userid_list), to: RelationshipLib
 
-  @spec check_avoid_status(T.userid(), [T.userid()]) :: :ok | :avoiding | :avoided
```
jauggy, line 1573: Function removed because not used anywhere

lib/teiserver/account.ex
```diff
@@ -1570,9 +1570,6 @@ defmodule Teiserver.Account do
   @spec check_block_status(T.userid(), [T.userid()]) :: :ok | :blocking | :blocked
   defdelegate check_block_status(userid, userid_list), to: RelationshipLib
```
jauggy, line 1572: Function removed because no longer used

jauggy:
@L-e-x-o-n FYI here is Teifion's old code
```
  @spec list_userids_blocked_by_userid(T.userid()) :: [T.userid()]
  def list_userids_blocked_by_userid(userid) do
    Teiserver.cache_get_or_store(:account_block_cache, userid, fn ->
      Account.list_relationships(
        where: [
          from_user_id: userid,
          state_in: ["avoid", "block"]
        ],
        select: [:to_user_id]
      )
      |> Enum.map(fn r ->
        r.to_user_id
      end)
    end)
  end
  ```
  This means when counting how many are blocking player A, it would include people avoiding player A. This is wrong and is back to front. I have fixed this in my PR.

File: lib/teiserver/battle/balance/respect_avoids.ex [hunk 5]
```diff
@@ -0,0 +1,529 @@
...
+  # Noobs have high uncertainty and chev 1,2,3
+  @spec get_noobs([RA.player()]) :: any()
+  def get_noobs(players) do
+    Enum.filter(players, fn player ->
+      is_newish_player?(player.rank, player.uncertainty)
+    end)
+  end
+
+  def is_newish_player?(rank, uncertainty) do
+    uncertainty >= @high_uncertainty && rank <= 2
```
L-e-x-o-n, line 527: I think it's better to avoid using rank where possible, rank can be calculated by other methods than playtime in which case this stops working as intended.

lib/teiserver/battle/balance/respect_avoids.ex
```diff
@@ -0,0 +1,529 @@
...
+      ) * rating
+    else
+      rating
+    end
+  end
+
+  @spec get_avoids([any()], number(), boolean()) :: [String.t()]
+  def get_avoids(player_ids, lobby_max_avoids, debug_mode? \\ false) do
+    # The max number of avoids per player to pull from db
+    player_limit = 2
```
L-e-x-o-n, line 455: It might be nice to have as a constant on top.

lib/teiserver/mix_tasks/party_balance_stats.ex
```diff
@@ -0,0 +1,299 @@
+defmodule Mix.Tasks.Teiserver.PartyBalanceStats do
```
L-e-x-o-n, line 1: We already have the results, will we need this again in future?

L-e-x-o-n (commented):
Please also change avoid explanation in lib/teiserver_web/live/account/relationship/index.html.heex to match the new use case.

lib/teiserver/battle/balance/brute_force_avoid.ex
```diff
@@ -0,0 +1,161 @@
...
+    %{
+      score: score,
+      rating_diff_penalty: rating_diff_penalty,
+      broken_avoid_penalty: broken_avoid_penalty,
+      broken_party_penalty: broken_party_penalty
+    }
+  end
+
+  def get_players_from_indexes(player_indexes, players_with_index) do
+    Enum.filter(players_with_index, fn {_player, index} ->
```
L-e-x-o-n, line 100: Something like
```ex
players_with_index
  |> Enum.filter(fn {_player, index} -> index in player_indexes end)
  |> Enum.map(fn {player, _index} -> player end)
  ```
might be a bit cleaner.

File: lib/teiserver/battle/balance/brute_force_avoid.ex [hunk 9]
```diff
@@ -0,0 +1,161 @@
...
+      Enum.map(combos, fn x ->
+        get_players_from_indexes(x, players_with_index)
+      end)
+      |> Enum.map(fn team ->
+        result = score_combo(team, players, avoids, parties)
+        Map.put(result, :first_team, team)
+      end)
+      |> Enum.min_by(fn z ->
+        z.score
+      end)
```
L-e-x-o-n, line 44: You can remove the extra `Enum.map`

```ex
  combos
  |> Enum.map(fn combo ->
    team = get_players_from_indexes(combo, players_with_index)
    result = score_combo(team, players, avoids, parties)
    Map.put(result, :first_team, team)
  end)
  |> Enum.min_by(fn z -> z.score end)
  ```

File: lib/teiserver/mix_tasks/party_balance_stats.ex (hunk 7 above)
Comment (jauggy) [reply lib/teiserver/mix_tasks/party_balance_stats.ex:1]: Since I've made changes to the value where parties get broken, ideally we put this PR on integration again and rerun:
```
mix teiserver.party_balance_stats /var/log/teiserver/results.txt
```

Previously I hardcoded parties to be broken when rating diff > 14
Now I change to ignore parties when rating diff > max(10, 5% of team rating)

lib/teiserver/battle/balance/respect_avoids.ex
jauggy: I won't change it for this PR as there are players who have high uncertainty that are 4 chevs who do not play like new players. New players do dumb stuff like make multiple labs and go t2 without resources. 4 chevs don't do that. 4 chevs can have high uncertainty from playing PvE, unranked, or different game modes like 1v1.

However, in the future I can look to improving this by giving playtime as a parameter to the balancer and could use that instead of rank.

PR #442, opened 2024-09-03
Tachyon matchmaking: join and leave queues

geekingfrog:
This is the first step to implement matchmaking, in a very mvp state.
With this, one can join and leave queues, although there is exactly one queue (1v1). There is nothing happening after this point, the actual grouping of players and messaging should be the next pull request.
Also a note, party will be supported since a queue member can be made of multiple players.

The queues are heavily inspired from the [the existing queues made by teifion](https://github.com/beyond-all-reason/teiserver/blob/e584d6bac272c5ef7bd7973b8676f499dcf9cad8/lib/teiserver/data/matchmaking.ex#L1-L35) but ultimately I chose a slightly different approach:
* existing queues have attributes that tie them to a UI like `colour` or `icon`. I think it's a distraction for the mvp and would rather focus on the core logic
* existing queues are quite tied to a db storage. I don't think they will change much so hardcoding them is a lot simpler.
* The concept of bucket is likely going away. I may end up using it after some load testing, but the recurring issue of matchmaking in games is long wait time, meaning not enough players in queue. So optimizing for large queue size seems premature

### DIFF
--- a/lib/teiserver/application.ex
+++ b/lib/teiserver/application.ex
@@ -137,6 +137,7 @@ defmodule Teiserver.Application do
         # Matchmaking
         {DynamicSupervisor, strategy: :one_for_one, name: Teiserver.Game.QueueSupervisor},
         Teiserver.Data.MatchmakingCache,
+        Teiserver.Matchmaking.System,
 
         # Coordinator mode
         {DynamicSupervisor,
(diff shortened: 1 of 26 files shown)


L-e-x-o-n:
Really great work!

PR #444, opened 2024-09-06
Naming improvements for balance algos

jauggy:
The following algos are now callable via boss
```
$balancealgorithm default
$balancealgorithm auto
$balancealgorithm split_noobs
```

default will check the config for what it should pick (loser_picks probably). 

In the future maybe could rename DefaultBalance to be AutoBalance but I wanted to keep this PR short.

For SPADS I need to adjust so that auto is an option.

auto will pick between split_noobs or loser_picks depending on circumstances

## Testing
You can simply call the algos listed above.
Screenshot 2024-09-06 224747

When you call "default" the text output will say loser_picks


jauggy:
SPADS changes here: https://github.com/beyond-all-reason/spads_config_bar/pull/162/files

jauggy:
@L-e-x-o-n the SPADS changes have made it through to integration

So if you can test on integration via
```
!balancealgorithm auto
!balancealgorithm default
!balancealgorithm split_noobs
```

PR #446, opened 2024-09-08
Allow chat-restricted users to run whitelisted commands.

DeviousNull:
This patch allows users with an active chat restriction to use several common/necessary commands. Limitations are placed on both the commands that are allowed, as well as on the parameters (if any) allowed with each command. The following are allowed:
- `!vote`, and the various yes/no/blank parameters for it
- `!endvote` and `!ev`
- `!help`
- `!balance` (or a vote for it)
- `!start` (or a vote for it)
- `!forcestart` (or a vote for it)
- `!stop` (or a vote for it)
- `!resign` (or a vote for it)

Other commands, including commands to adjust various game settings, are still restricted. This is because most other commands would require almost-arbitrary parameters to be useful, and so they would provide an avenue to bypass a chat restriction.

A few commands were not whitelisted because they are more rarely used; such commands could be whitelisted in the future if there is a need.

### DIFF
--- a/lib/teiserver/lobby/libs/chat_lib.ex
+++ b/lib/teiserver/lobby/libs/chat_lib.ex
@@ -302,7 +302,7 @@ defmodule Teiserver.Lobby.ChatLib do
 
   @valid_mute_chat_regex [
     ~r/^!vote( (y|yes|n|no|b|blank))?$/i,
-    ~r/^!(ev|endvote|help)$/i,
+    ~r/^!(y|n|b|ev|endvote|help)$/i,
     ~r/^!(cv |callvote )?balance$/i,
     ~r/^!(cv |callvote )?start$/i,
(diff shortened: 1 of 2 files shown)


lib/teiserver/lobby/libs/chat_lib.ex
```diff
@@ -299,4 +299,27 @@ defmodule Teiserver.Lobby.ChatLib do
...
+    cond do
+      Enum.any?(@valid_mute_chat_regex, fn regex ->
+        String.match?(msg, regex)
+      end) ->
+        true
+
+      true ->
+        false
+    end
+  end
```
L-e-x-o-n, line 324: Why not just return
```ex
Enum.any?(@valid_mute_chat_regex, fn regex ->
        String.match?(msg, regex)
end)
```
without `cond`?

PR #448, opened 2024-09-08
Add party to fake data task

jauggy:
This PR adds two things:
1. Reduces the amount of 1-3 chevs (which are counted as noobs for split_noobs). There will now be only 1-2 noobs in a 16 player game.
2. Adds parties to fake data task

# Testing
Run
```
mix teiserver.fakedata
```
Go to a match on website and you will see parties represented as dice

### DIFF
--- a/lib/teiserver/mix_tasks/fake_data.ex
+++ b/lib/teiserver/mix_tasks/fake_data.ex
@@ -338,8 +338,9 @@ defmodule Mix.Tasks.Teiserver.Fakedata do
         shuffled_users = Enum.shuffle(users)
         team1 = shuffled_users |> Enum.take(team_size)
         team2 = shuffled_users |> Enum.reverse() |> Enum.take(team_size)
-
-        game_type = MatchLib.game_type(team_size, 2)
+        team_count = 2
+        num_players = team_size * team_count
+        game_type = MatchLib.game_type(team_size, team_count)
 
         team1_score =
           team1
@@ -362,7 +363,7 @@ defmodule Mix.Tasks.Teiserver.Fakedata do
             data: %{},
             tags: %{},
             winning_team: if(team1_score > team2_score, do: 0, else: 1),
-            team_count: 2,
+            team_count: team_count,
             team_size: team_size,
             passworded: false,
             processed: true,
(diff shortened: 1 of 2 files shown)


lib/teiserver/mix_tasks/fake_data.ex
```diff
@@ -458,4 +459,24 @@ defmodule Mix.Tasks.Teiserver.Fakedata do
...
+  defp get_party_id(num_players) do
+    case is_in_party?() do
+      true ->
+        num_parties = trunc(num_players / 4)
+        # party id is a string
+        "#{Enum.random(1..num_parties)}"
+
+      false ->
+        nil
+    end
```
StanczakDominik, line 473: So... an if clause? 😁 Ah, but you want to output that string or nil... I guess, yeah, it works :)

lib/teiserver/mix_tasks/fake_data.ex
```diff
@@ -458,4 +459,24 @@ defmodule Mix.Tasks.Teiserver.Fakedata do
...
+  end
+
+  defp is_in_party? do
+    chance_in_party = 40
+    # Number from 0 to 100
+    random = Enum.random(0..100)
+    random <= chance_in_party
```
StanczakDominik, line 480: Just for my own curiosity, is there a quicker way to do this? How's elixir on random numbers?

StanczakDominik (commented):
Looks okay, got two places where I wonder whether cleaner ways to do these might not exist but can't check today - tomorrow, sure :) if I find them I'll put them in as suggestions.

lib/teiserver/mix_tasks/fake_data.ex
jauggy: https://stackoverflow.com/a/42295898
I just copied this

StanczakDominik (approved):

lib/teiserver/mix_tasks/fake_data.ex
geekingfrog: For this usecase you can also use `:rand.uniform/1` from [erlang stdlib](https://www.erlang.org/doc/apps/stdlib/rand.html#uniform/1) like for the entire function
```
:rand.uniform(100) <= 40
```

PR #471, opened 2024-09-15
Ignore avoids/blocks once player

jauggy:
## Current implementation
Avoids are checked:
1. When going from spec to player
2. When pressing the ready button
3. After a match ends

Blocks are checked
1. When joining a lobby
2. After a match ends

## New implementation
This PR removes the avoid/block checks when the person is already a player.

Avoids are checked:
1. When going from spec to player

Blocks are checked
1. When joining a lobby

## Other fixes
Avoids now no longer count as blocks
Blocks count as avoids. 
They seem to have been inverted

## Testing
Testing this PR is very difficult but I did manage to test locally using two players.

### DIFF
--- a/lib/teiserver/account/libs/relationship_lib.ex
+++ b/lib/teiserver/account/libs/relationship_lib.ex
@@ -175,7 +175,7 @@ defmodule Teiserver.Account.RelationshipLib do
       Account.list_relationships(
         where: [
           to_user_id: userid,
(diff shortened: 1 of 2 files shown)


jauggy:
Given more awareness about this topic, this is not really an issue. It's an issue for some players due to behaviour score and behaviour score is planned to be removed.

PR #472, opened 2024-09-16
Reveal balance algorithm to regular users on website

jauggy:
Currently if we go here: https://server4.beyondallreason.info/battle/lobbies/show/9517
only testers can see the balance algorithm

However, it is useful for normal users to know such as if they want to play in party certain algos are better for them. It is not very intrusive as most people won't even use the website for joining games so it would only be used by advanced users.

An example is MVPete https://ptb.discord.com/channels/549281623154229250/1284942772482347159/1285110298306674729
who dodged games to get parties to work. It might be more convenient to use the website to find games where the balance algo is favourable to parties.

### DIFF
--- a/lib/teiserver_web/live/battles/lobbies/show.html.heex
+++ b/lib/teiserver_web/live/battles/lobbies/show.html.heex
@@ -108,8 +108,8 @@ spectators =
               <li>
                 Queue size: <%= Enum.count(@consul.join_queue ++ @consul.low_priority_join_queue) %>
               </li>
+              <li>Balance algorithm: <%= @consul.balance_algorithm %></li>
               <%= if @tester_perms do %>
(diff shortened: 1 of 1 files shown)


geekingfrog:
Side note, I tend to prefer more linear histories. So instead of merging a branch into yours, use `git rebase <other branch>`, this way this PR would still only have one commit and tracking changes is easier.

jauggy:
> Side note, I tend to prefer more linear histories. So instead of merging a branch into yours, use `git rebase <other branch>`, this way this PR would still only have one commit and tracking changes is easier.

That merge was me using amend commit to try and get the tests to pass

geekingfrog:
But for this PR, you can still rebase, it'll put the commit on top of the <target branch>, so you get the changes for the test, and a clean linear history

PR #474, opened 2024-09-18
 Recheck lobby restrictions after certain commands are used.

DeviousNull:
Currently, changing rating/chevron restrictions for a lobby does not affect existing lobby members until the end of the next match.

This patch changes that behavior, so that the new rating/chevron restrictions are enforced shortly after any of the following commands run:
- $minchevlevel
- $maxchevlevel
- $minratinglevel
- $maxratinglevel
- $setratinglevels

This is done by scheduling a `:recheck_membership` message in the handlers for each of the above commands, making use of the pre-existing `handle_info` in `consul_server.ex`.

____

As discussed in https://discord.com/channels/549281623154229250/1279328508811808828/1280640548830707713 (before the thread _once again_ went off topic and was locked).


L-e-x-o-n:
> Currently, changing rating/chevron restrictions for a lobby does not affect existing lobby members until the end of the next match.

This is a feature, not a bug. It allows players who waited in the queue for a while to play one game before being removed.

lib/teiserver/lobby.ex
```diff
@@ -766,6 +766,10 @@ defmodule Teiserver.Lobby do
       battle.founder_id == changer.userid ->
         true
 
+      # Coordinator can do anything
+      Coordinator.get_coordinator_userid() == changer.userid ->
+        true
+
```
L-e-x-o-n, line 772: Why is this necessary?

jauggy:
One thing that is not obvious is that when you press the ready button, there is actually already a check for rating restrictions. This gets called when you press the ready button. 
```
  defp request_user_change_status(new_client, %{userid: userid} = existing, state) do
    user = Account.get_user_by_id(userid)
    list_status = get_list_status(userid, state)

    new_client =
      if new_client.player == true and user_allowed_to_play?(user, existing, state) do
      ```
      
   Basically when ready is pressed we will have existing.ready == false and new_client.ready == true

**Comment by L-e-x-o-n** [comment]
> One thing that is not obvious is that when you press the ready button, there is actually already a check for rating restrictions. This gets called when you press the ready button.
> 
> ```
>   defp request_user_change_status(new_client, %{userid: userid} = existing, state) do
>     user = Account.get_user_by_id(userid)
>     list_status = get_list_status(userid, state)
> 
>     new_client =
>       if new_client.player == true and user_allowed_to_play?(user, existing, state) do
>       
> ```
> 
> Basically when ready is pressed we will have existing.ready == false and new_client.ready == true
> 
> So if someone is unready then the rating thresholds are changed, then they press ready, they may not be able to play. I'm just pointing it out as a thing I learned that is not obvious.

Oh nice find, I missed this. Doesn't this also take effect when they ready up even if ratings changed during queue time?

**Comment by jauggy** [comment]
Correct. If you're in queue then become player later you have to press the ready button and therefore will just be removed as player. Confirmed by Laudy: https://discord.com/channels/549281623154229250/1277718157074759681/1278101155246899253

**Comment by jauggy** [comment]
If you want someone to bypass check when they press ready button then you need to add something like this inside `request_user_change_status`
```
ignore_allow_play_check? = new_client.ready == true && existing.ready == false
```

They would still be removed from player list after match end

L-e-x-o-n:
> If you want someone to bypass check when they press ready button then you need to add something like this inside `request_user_change_status`
> 
> ```
> ignore_allow_play_check? = new_client.ready == true && existing.ready == false
> ```
> 
> They would still be removed from player list after match end

Thats makes the chack a bit useless, no? If there is an exception it should only be for the rating limits not the other checks. It would now trigger on unready which is a weird time to check if someone is allowed to play IMO.

However, seeing this is not a feature as I originally thought, I don't think there are issues with leaving it as is and also checking memberships immediately as this PR does.

jauggy:
That line of code I wrote means you check rating when going from spec to player, but not when going from unready to ready. But yes, you'd have to make it only check for rating filters and not bans.

At the moment though, if someone readies up before the rating change then they'd be safe.

jauggy:
> I found other situations where the current code would force-spec players who don't meet the rating requirements, including adding a new bot to the lobby (which does not require a vote, currently). For typical players, it can seem pretty random what triggers/does-not-trigger enforcement of the lobby restrictions.
> 
> 
> 
> If nothing else, this PR makes things more consistent: restrictions are applied immediately and are always in effect.

I guess if adding a bot forces a check then it does make sense to just make it consistent.

PR #475, opened 2024-09-19
Add task to calculate balance stats

jauggy:
Goes through 500 games of 8v8 and 500 games 5v5 to calculate some stats using loser_picks and auto balance

It will return team rating diff and also the adjusted rating team diff. Adjusted team rating diff is where was adjust the rating of newish players. split_noobs assumes new players are the worst in game (i.e. 0 rating) and we trust their rating once they hit 6.65 uncertainty. We can use a formula to interpolate their adjusted rating:

```
    starting_uncertainty = 25 / 3
    uncertainty_cutoff = 6.65

    # When uncertainty is less than 6.65 this will return just rating
    # When uncertainty is default this will return 0
    min(
      1,   (starting_uncertainty - uncertainty) /   (starting_uncertainty -    uncertainty_cutoff)
      ) * rating
```

## Testing
  On integration server it is recommended you output to a specific path as follows:
```
  mix teiserver.balance_stats /var/log/teiserver/results.txt
```

### DIFF
--- a/lib/teiserver/mix_tasks/balance_stats.ex
+++ b/lib/teiserver/mix_tasks/balance_stats.ex
@@ -1,0 +1,306 @@
+defmodule Mix.Tasks.Teiserver.BalanceStats do
+  @moduledoc """
+  Get stats on different balance algorithms
+  mix teiserver.balance_stats
+  On integration server it is recommended you output to a specific path as follows:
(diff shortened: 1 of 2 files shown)


L-e-x-o-n:
We don't need this anymore, right?

PR #478, opened 2024-09-20
Get custom deviation result for split_noobs

jauggy:
The deviation stat when calculating balance is displayed in chobby whenever you call !balance
If the result is very high, players can get upset. With split_noobs, we do not trust the rating (17) of new players, which can lead to high deviations. This PR aims to adjust new player's ratings in the balancer so that the deviation result is smaller.

Note that the deviation result in teiserver is already a custom deviation method that is not the same as SPADS.

Also note that teiserver adjusts people's ratings already because it applies randomness from -0.5 to 0.5. So in $explain someone could have negative rating. But in chobby nobody is displayed as negative.

split_noobs doesn't apply randomness, but does adjust ratings with this PR.

## Adjusted rating formula
```
     min(
        1,
        (starting_uncertainty - uncertainty) /
          (starting_uncertainty - uncertainty_cutoff)
      ) * rating
```
When you are at starting uncertainty, this should return 0.
When you are at the uncertainty cutoff, this should return 1 * rating.
When you are in between, it should return x*rating (where x is a number between 0 and 1)

## Testing
You can select an old match in teiserver and compare the team sums of loser_picks and split_noobs. The team sum of split_noobs will be smaller when newish players are involved since their adjusted rating ranges from 0 to their true rating.

### DIFF
--- a/lib/teiserver/battle/balance/split_noobs.ex
+++ b/lib/teiserver/battle/balance/split_noobs.ex
@@ -344,13 +344,35 @@ defmodule Teiserver.Battle.Balance.SplitNoobs do
           Enum.zip([members, ratings, ranks, names, uncertainties]),
         # Create result value
         do: %{
-          rating: rating,
+          rating: adjusted_rating(rating, uncertainty, rank),
           name: name,
           id: id,
           uncertainty: uncertainty,
           rank: rank,
           in_party?: count > 1
         }
+  end
+
+  # This balance algorithm will use an adjusted rating for newish players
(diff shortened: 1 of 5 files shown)


jauggy:
Here is a sample match with high deviation
image
image

The team ratings are

team 1: 64, 
Team 2: 97

After using adjusted ratings the team ratings are:
```
    assert ratings == %{1 => 31.950594059405947, 2 => 27.78324752475248}

    deviation = BalanceLib.get_deviation(ratings)
    assert deviation == 13
```

This result is in a test in this PR

Issue #430, opened 2024-08-26
[Bug]: (Enum.EmptyError) empty error during balance_battle request


L-e-x-o-n:
### Describe the Bug

Logs:
```
Request: POST /teiserver/api/spads/balance_battle
** (exit) an exception was raised:
    ** (Enum.EmptyError) empty error
        (elixir 1.14.5) lib/enum.ex:1855: anonymous fn/0 in Enum.max/1
        (teiserver 0.1.0) lib/teiserver_web/controllers/api/spads_controller.ex:164: TeiserverWeb.API.SpadsController.balance_battle/2
        (teiserver 0.1.0) lib/teiserver_web/controllers/api/spads_controller.ex:1: TeiserverWeb.API.SpadsController.action/2
        (teiserver 0.1.0) lib/teiserver_web/controllers/api/spads_controller.ex:1: TeiserverWeb.API.SpadsController.phoenix_controller_pipeline/2
        (phoenix 1.7.6) lib/phoenix/router.ex:430: Phoenix.Router.__call__/5
        (teiserver 0.1.0) lib/teiserver_web/endpoint.ex:1: TeiserverWeb.Endpoint.plug_builder_call/2
        (teiserver 0.1.0) lib/teiserver_web/endpoint.ex:1: TeiserverWeb.Endpoint.call/2
        (plug_cowboy 2.6.1) lib/plug/cowboy/handler.ex:11: Plug.Cowboy.Handler.init/2
```

```ex
          team_size =
            balance_result.team_sizes
            |> Map.values()
            |> Enum.max()
            ```
It seems `team_sizes` might be empty.

### Reproduce the bug

Unfortunately the logs don't contain any other details. Don't know the team count, team size or balance algorithm used. There are only 2 such errors recently in logs so I think it's probably some edge case.

jauggy:
@L-e-x-o-n you can assign to me.

jauggy:
So I had an idea to press balance as spec when there are no players - but it won't even let you do that.

jauggy:
@L-e-x-o-n where are the logs stored? Want to try deliberately raising errors on my local environment.

L-e-x-o-n:
When running it locally over WSL? Not sure if they are saved, but wouldn't you be able to see them when starting the server with `mix phx.server`?

jauggy:
[...]

PR #480, opened 2024-09-21
Add logs for errors inside balance result

jauggy:
Add logs for: https://github.com/beyond-all-reason/teiserver/issues/430
This doesn't fix the bug. But will have more logs to look at.

Does a check to see if team_sizes is empty


L-e-x-o-n:
Thanks

PR #481, opened 2024-09-22
Auto balance call split noobs when at least 3 parties

jauggy:
- Makes auto balance call split noobs when there are at least 3 parties.  
- Rename default balance to auto balance in code.

loser_picks is stricter and tends to break up parties more often than split_noobs. this was deliberate design so we don't have one stacked team versus all solos. However, preserving parties when there are at least 3 parties should be less of an issue because this means that there's a high chance both teams will have parties.

Here's examples of supporting many parties and with both teams having parties:
image

image

### Other minor improvements
When logging parties change grouping from [ to (. This makes it easier to read since many use clan tags that use [].

### DIFF
--- a/test/teiserver/protocols/spring/spring_auth_async_test.exs
+++ b/test/teiserver/protocols/spring/spring_auth_async_test.exs
@@ -1,5 +1,5 @@
 defmodule Teiserver.SpringAuthAsyncTest do
-  use Teiserver.ServerCase, async: true
+  use Teiserver.ServerCase, async: false
   alias Teiserver.Client
   alias Teiserver.Protocols.Spring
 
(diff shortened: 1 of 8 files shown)


test/teiserver/protocols/spring/spring_auth_async_test.exs
```diff
@@ -1,5 +1,5 @@
 defmodule Teiserver.SpringAuthAsyncTest do
-  use Teiserver.ServerCase, async: true
+  use Teiserver.ServerCase, async: false
   alias Teiserver.Client
   alias Teiserver.Protocols.Spring
```
jauggy, line 4: This was causing error in automated tests...

Issue #368, opened 2024-07-21
[Bug]: Errors on mix test related to spring auth


jauggy:
### Describe the Bug

Red errors when I run mix test for spring_auth_test.exs

### Reproduce the bug

```
mix test pathto/spring/spring_auth_test.exs
```

### Additional context

```
...2024-07-21 22:53:49.096 [error] Postgrex.Protocol (#PID<0.691.0>) disconnected: ** (DBConnection.ConnectionError) owner #PID<0.1598.0> exited

Client #PID<0.1602.0> is still using a connection from owner at location:

    (erts 13.1.5) :prim_inet.recv0/3
    (postgrex 0.17.1) lib/postgrex/protocol.ex:3188: Postgrex.Protocol.msg_recv/4
    (postgrex 0.17.1) lib/postgrex/protocol.ex:2213: Postgrex.Protocol.recv_bind/3
    (postgrex 0.17.1) lib/postgrex/protocol.ex:2193: Postgrex.Protocol.rebind_execute/4
    (ecto_sql 3.10.1) lib/ecto/adapters/sql/sandbox.ex:375: Ecto.Adapters.SQL.Sandbox.Connection.proxy/3
    (db_connection 2.5.0) lib/db_connection/holder.ex:354: DBConnection.Holder.holder_apply/4
    (db_connection 2.5.0) lib/db_connection.ex:1432: DBConnection.run_execute/5
    (db_connection 2.5.0) lib/db_connection.ex:1527: DBConnection.run/6
    (db_connection 2.5.0) lib/db_connection.ex:656: DBConnection.parsed_prepare_execute/5
    (db_connection 2.5.0) lib/db_connection.ex:648: DBConnection.prepare_execute/4
    (postgrex 0.17.1) lib/postgrex.ex:340: Postgrex.query/4
    (ecto_sql 3.10.1) lib/ecto/adapters/sql.ex:955: Ecto.Adapters.SQL.struct/10
    (ecto 3.10.3) lib/ecto/repo/schema.ex:764: Ecto.Repo.Schema.apply/4
    (ecto 3.10.3) lib/ecto/repo/schema.ex:377: anonymous fn/15 in Ecto.Repo.Schema.do_insert/4
    (teiserver 0.1.0) lib/teiserver/telemetry/libs/simple_server_event_lib.ex:22: Teiserver.Telemetry.SimpleServerEventLib.log_simple_server_event/2
[...]

PR #493, opened 2024-09-28
fix for flaky renameuser test using polling-until-expected-result

NortySpock:
### Changes
- fixes flaky test
- re-enables that test
- Probably has no impact on the issues reported on https://github.com/beyond-all-reason/teiserver/issues/368 , because those error reports about red text appear to be due to the database connection throwing errors, not failing tests.

### Potential risks
... I'm actually not sure how long it polls or what the timeout is -- if it never times out the poll would run forever, and thus the test would never finish.

_I'm open to critiques or improvements on this PR. I am open to troubleshooting and trying to fix other tests after this if this PR is accepted._

### DIFF
--- a/test/teiserver/protocols/spring/spring_auth_test.exs
+++ b/test/teiserver/protocols/spring/spring_auth_test.exs
@@ -479,7 +479,8 @@ CLIENTS test_room #{user.name}\n"
     # Rename with existng name
     _send_raw(socket, "RENAMEACCOUNT #{watcher_user.name}\n")
     reply = _recv_raw(socket)
-    assert reply == "SERVERMSG Username already taken\n"
+    assert String.starts_with?(reply, "SERVERMSG")
+    assert String.contains?(reply, "Username already taken")
 
     # Perform rename
     _send_raw(socket, "RENAMEACCOUNT test_user_rename\n")
--- a/test/teiserver/protocols/spring/spring_auth_test.exs
+++ b/test/teiserver/protocols/spring/spring_auth_test.exs
@@ -18,9 +18,23 @@ defmodule Teiserver.SpringAuthTest do
       new_user: 2
     ]
 
+  import Teiserver.Support.Tachyon, only: [poll_until: 2]
+
   setup do
     %{socket: socket, user: user} = auth_setup()
     {:ok, socket: socket, user: user}
+  end
+
+  # wrapped login attempt in a function to make polling for expected output easier
+  def _send_login_attempt_and_parse_reply_msg_from_server_socket(socket, login_name) do
+    _send_raw(
+      socket,
+      "LOGIN #{login_name} X03MO1qnZdYdgyfeuILPmQ== 0 * LuaLobby Chobby\t1993717506 0d04a635e200f308\tb sp\n"
+    )
+
+    reply = _recv_until(socket)
+    [server_response | _remainder] = String.split(reply, "\n")
+    server_response
   end
 
   test "Test bad match", %{socket: socket} do
@@ -442,9 +456,6 @@ CLIENTS test_room #{user.name}\n"
     assert reply =~ "s.battles.id_list "
   end
 
-  # this is a flakey test, but only when other tests are running around.
-  # you may be able to reproduce with the seed 846266
-  @tag :needs_attention
   test "RENAMEACCOUNT", %{socket: socket, user: user} do
     old_name = user.name
     new_name = "test_user_rename"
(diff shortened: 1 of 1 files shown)


NortySpock:
Looks like the error message from the test is 

```
Error:      test/teiserver/protocols/spring/spring_auth_test.exs:459
     Assertion with == failed
     code:  assert reply == "SERVERMSG Invalid characters in name (only a-z, A-Z, 0-9, [, ] allowed)\n"
     left:  "SERVERMSG Invalid characters in name (only a-z, A-Z, 0-9, [, ] allowed)\nSAIDPRIVATE Coordinator Invalid characters in name (only a-z, A-Z, 0-9, [, ] allowed)\n"
     right: "SERVERMSG Invalid characters in name (only a-z, A-Z, 0-9, [, ] allowed)\n"
     stacktrace:
       test/teiserver/protocols/spring/spring_auth_test.exs:475: (test)

```

This test failure is not occurring when I run locally, but on the other hand I am getting other errors locally that indicate some sort of admin permissions problem, so possibly my local setup is incomplete.

```
 69) test update user renders errors when data is invalid (TeiserverWeb.Admin.UserControllerTest)
     test/teiserver_web/controllers/admin/user_controller_test.exs:94
     ** (KeyError) key :management_roles not found in: %{
       user: %Teiserver.Account.User{
         __meta__: #Ecto.Schema.Metadata<:loaded, "account_users">,
         id: 1810,
         name: "Current user",
```

NortySpock:
Somehow managing to get (in my local build) in the PR branch than in master, even when I pin the seed and exclude ignored tests...

```
from master:
mix test --exclude needs_attention --seed 0
...
Finished in 95.6 seconds (14.4s async, 81.2s sync)
517 tests, 3 failures, 85 excluded


from fix-flaky-renamuser-test-with-polling:

mix test --exclude needs_attention --seed 0
...

Finished in 102.3 seconds (12.9s async, 89.3s sync)
518 tests, 4 failures, 84 excluded

```

Since it does not seem I'm making progress on this front, I will likely close this PR and attempt to work on a different issue.

test/teiserver/protocols/spring/spring_auth_test.exs
```diff
@@ -461,12 +472,15 @@ CLIENTS test_room #{user.name}\n"
...
+    assert String.starts_with?(reply, "SERVERMSG")
+    assert String.contains?(reply, "Invalid characters in name")
+    assert String.contains?(reply, "(only a-z, A-Z, 0-9, [, ] allowed)")
 
     # Rename with existng name
     _send_raw(socket, "RENAMEACCOUNT #{watcher_user.name}\n")
     reply = _recv_raw(socket)
-    assert reply == "SERVERMSG Username already taken\n"
+    assert String.starts_with?(reply, "SERVERMSG")
+    assert String.contains?(reply, "Username already taken")
```
geekingfrog, line 483: This is fine as it is. But you can leverage Elixir's pattern matching for something a little bit more direct:

```elixir
assert "SERVERMSG" <> rest = _recv_raw(socket)
assert String.contains?(rest, …)
```

test/teiserver/protocols/spring/spring_auth_test.exs
```diff
@@ -478,9 +492,6 @@ CLIENTS test_room #{user.name}\n"
     wreply = _recv_raw(watcher)
     assert wreply == "REMOVEUSER #{user.name}\n"
 
-    # Give it a chance to perform some of the actions
-    :timer.sleep(250)
```
geekingfrog, line 482: Isn't that somewhat still time dependent? I would rather use a polling function to assert `Client.get_client_by_id` returns `nil`

test/teiserver/protocols/spring/spring_auth_test.exs
```diff
@@ -497,14 +508,15 @@ CLIENTS test_room #{user.name}\n"
...
+    expected_server_response = "DENIED Flood protection - Please wait 20 seconds and try again"
 
-    reply = _recv_until(socket)
-    [accepted | _remainder] = String.split(reply, "\n")
-    assert accepted == "DENIED Flood protection - Please wait 20 seconds and try again"
+    server_response =
+      Teiserver.Support.Tachyon.poll_until(
+        fn -> _send_login_attempt_and_parse_reply_msg_from_server_socket(socket, new_name) end,
+        &(&1 == expected_server_response)
+      )
```
geekingfrog, line 517: I am not the biggest fan of the style here. The helper function is a mouthful, it's used only once and it's declared very far from this call site.
Since you're already using anonymous function anyway, I think I'd rather inline everything:

```elixir
    Teiserver.Support.Tachyon.poll_until(
      fn ->
        _send_raw(
          socket,
          "LOGIN #{new_name} X03MO1qnZdYdgyfeuILPmQ== 0 * LuaLobby Chobby\t1993717506 0d04a635e200f308\tb sp\n"
        )

        reply = _recv_until(socket)
        [accepted | _remainder] = String.split(reply, "\n")
        accepted
      end,
      &(&1 == expected_server_response)
    )
```

test/teiserver/protocols/spring/spring_auth_test.exs
```diff
@@ -497,14 +508,15 @@ CLIENTS test_room #{user.name}\n"
...
+
+    assert server_response == expected_server_response
```
geekingfrog, line 519: This line is superfluous because it's already checked in the polling helper.

PR #495, opened 2024-10-01
Fix logs for loser picks

jauggy:
Currently logs for loser_picks show user ids when there are parties

This PR shows names of players instead of user ids for logs

## Testing
Simply go to a match on website and look at loser_picks balance logs


L-e-x-o-n:
There are some interesting players in those parties

PR #507, opened 2024-10-17
Better balance display

jauggy:
This PR adds standard deviation to the balance tab as well as shows a table of theoretical players based on the balance result

## Testing
To test this, simply go to a match and select the balance tab and you will see the ui enhancements.


lib/teiserver_web/live/battles/match/show.html.heex
```diff
@@ -303,27 +303,80 @@
...
+                <%= if party_colour do %>
+                  <td style={"background-color: #{rgba_css(party_colour, 0.3)};"} width="50">
+                    <Fontawesome.icon icon={party_idx} style="solid" size="lg" />
+                  </td>
+                <% else %>
+                  <td width="50">
+                    &nbsp;
+                  </td>
+                <% end %>
+                <td><%= m.team_id + 1 %></td>
```
jauggy, line 363: `team_id + 1` seems weird but that's what Teifion did already in line 263 in this file

PR #509, opened 2024-10-18
Add captain diff penalty to respect avoids

jauggy:
In the image below the two best players DmE and Jeff are on the same team. If we add a captain rating diff penalty then they will be split
image

Here is the new result:
```
               "Team 1: Gmans, AbyssWatcher, AcowAdonis, Jarial, [DmE], Spaceh, fraqzilla, Pantsu_Ripper",
               "Team 2: Faeton, Kaa, Zippo9, Threekey, mighty, MeowCat, Jeff, Saffir"
```

Note that this is the exact same result if we use a stdev penalty with importance of 2:
```
     "Team 1: Gmans, AbyssWatcher, AcowAdonis, Jarial, [DmE], Spaceh, fraqzilla, Pantsu_Ripper",
     "Team 2: Faeton, Kaa, Zippo9, Threekey, mighty, MeowCat, Jeff, Saffir"
```
See this PR for using stdev diff penalty instead:
https://github.com/beyond-all-reason/teiserver/pull/508

We can still use stdev penalty in brute_force.ex (called by split_noobs) if we want.

## More details
The score will have a new component captain diff penalty which is simply the rating difference between both teams' captains. If both captains have similar rating, this score will be lower which is preferred.

## Fix for new players in parties
In addition, while doing this PR I noticed a bug in respect_avoids
Normally newish players are not part of the brute force since we want to draft and spread them evenly across teams after the brute force. However, if a newish player is in a party, we do want them to be part of the brute force calculation to try and get them together with their party.

### DIFF
--- a/lib/teiserver/battle/balance/brute_force.ex
+++ b/lib/teiserver/battle/balance/brute_force.ex
@@ -16,6 +16,7 @@ defmodule Teiserver.Battle.Balance.BruteForce do
   import Teiserver.Helper.NumberHelper, only: [format: 1]
   require Integer
 
+  @stdev_diff_importance 4
   @party_importance 7
   @splitter "------------------------------------------------------"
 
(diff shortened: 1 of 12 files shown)


L-e-x-o-n:
Seeing that the result is the same, what is the reason for using captain penalty over st. dev. penalty?

Continuing the discussion from Discord, I think st. dev. penalty should have the added benefit of reducing the difference in ratings between players in the same team. This approach would lead to better quality team games.

Currently all top players on a team can stack on one side of the map resulting in the other side of the map having all the weakest players in that team. 
Depending on enemy teams positioning, you can have top players fighting top players or a race which team's top players beat the weakest one faster. I think there is a lot of uncertainty involved here because you can't predict enemy start positions.
In theory if everyone in a team is closer in ratings the weak side wouldn't be as weak nor the strong side as strong... each could hold their own better against the enemy team with less chance of entire sides of map collapsing. I think this would lead to more enjoyable games for everyone involved.

jauggy:
My main concern is that stdev can still put the top two players on the same team (as it did when using stdev importance of 1). But what if I add to this PR stdev diff for brute_force.ex ? Then we can test both at the same time and then choose which has the least complaints.

So basically we have `brute_force.ex` used by `split_noobs` using a stdev diff penalty (importance of 2).
and `brute_force_avoids` used by `respect_avoids` using a captain diff penalty.

L-e-x-o-n:
> My main concern is that stdev can still put the top two players on the same team (as it did when using stdev importance of 1). But what if I add to this PR stdev diff for brute_force.ex ?

That is solved by using importance of 2 or maybe going even higher.

> Then we can test both at the same time and then choose which has the least complaints.
> So basically we have `brute_force.ex` used by `split_noobs` using a stdev diff penalty (importance of 2). and `brute_force_avoids` used by `respect_avoids` using a captain diff penalty.

I don't think number of complaints is a reliable metric. With `auto` balance being default,` split_noobs` will be used significantly more than `respect_avoids`, then there is additional logic for avoids that make it harder for players to visually determine if there are issues. With that said, I think we can try it out anyway, but I don't think the final decision should be based on the number of complaints.

jauggy:
ok I agree with that. I'll add stdev diff penalty in this PR so we can check both.

Issue #149, opened 2023-07-29
Add config option to hide ratings in games


Teifion:
Currently ratings are sent to games via the spads controller. When this option is enabled the controller should instead send back a rating based on their position in an order of ratings (lowest rated player has 1 rating, next has 2 etc etc).

jauggy:
Here are some people's opinion's about this ticket discussed in discord main: https://discord.com/channels/549281623154229250/549281623577722899/1297501458354143355

It's probably worth it's own thread which I might do later.

L-e-x-o-n:
Not planned for the Spring protocol.

PR #512, opened 2024-10-19
Add config option for sending different rating values to SPADS

bigbluejay9:
Addresses #149

### DIFF
--- a/lib/teiserver_web/controllers/api/spads_controller.ex
+++ b/lib/teiserver_web/controllers/api/spads_controller.ex
@@ -250,25 +250,25 @@ defmodule TeiserverWeb.API.SpadsController do
         BalanceLib.get_user_rating_value_uncertainty_pair(userid, rating_type_id)
 
       "Position by in game rating (1 is lowest)" ->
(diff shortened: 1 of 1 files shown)


lib/teiserver/account.ex
```diff
@@ -1030,6 +1030,61 @@ defmodule Teiserver.Account do
     Repo.delete(rating)
   end
 
+  @doc """
+  Returns the player's rank when compared to other players by rating.
+
```
jauggy, line 1035: In teiserver "rank" means their chevron level indexed at 0. So we have to be careful here since you are using the term "rank" in a different way. Perhaps add a comment that this is ranking relative to other players completely unrelated to chevron levels. In teiserver db, the rank of a user relates to chev levels.

jauggy:
I think this would be improved if it is up to the boss of the lobby to anonymise the ratings or not. Currently with this PR, it is up to the Admin to change the config and this would affect all lobbies. I have a feeling that you'd get pushback from some people to hide the rating in all lobbies. 

In the future you could make an improvement to allow some sort of command for the boss to run like
```
$showratings off
```
Or something similar and this would only anonymise for this particular lobby.

PR #517, opened 2024-10-24
Fix error when unprocessed match

jauggy:
Fixes https://discord.com/channels/549281623154229250/1273218599653343262/1298564953107857419

### DIFF
--- a/lib/teiserver_web/live/battles/match/show.ex
+++ b/lib/teiserver_web/live/battles/match/show.ex
@@ -87,6 +87,7 @@ defmodule TeiserverWeb.Battle.MatchLive.Show do
         |> Enum.sort_by(fn m -> m.user.name end, &<=/2)
         |> Enum.sort_by(fn m -> m.team_id end, &<=/2)
 
+      # For unprocessed matches this will return %{}
       rating_logs =
         Game.list_rating_logs(
           search: [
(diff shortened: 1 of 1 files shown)


L-e-x-o-n:
While this should work fine, I think it might be better to use the `processed` boolean from the match instead. It would make it immediately clearer why it's used and it seems like good practice that could be applied to similar cases in the future where some match data might not yet be available.

jauggy:
Ok will do later in the week

L-e-x-o-n:
It seems the error happens for AI matches as well, do you know if are marked as processed?

jauggy:
Processed doesn't mean rated. Inside match_rating_lib, we only rate matches that are already processed. Meaning it's possible for a match to be processed = true, but not rated. 


```
  def rate_match(match, override) do
    logs = Game.list_rating_logs(search: [match_id: match.id], limit: 1, select: [:id])

    sizes =
      match.members
      |> Enum.group_by(fn m -> m.team_id end)
      |> Enum.map(fn {_, members} -> Enum.count(members) end)
      |> Enum.uniq()

    cheating =
      match.data
      |> Map.get("export_data", %{})
      |> Map.get("cheating", 0)

    cond do
      not Enum.member?(@rated_match_types, match.game_type) ->
        {:error, :invalid_game_type}

      match.processed == false ->
        {:error, :not_processed}
```

jauggy:
@L-e-x-o-n can you try deploy this on integration and check a bot match? I have one:
https://server5.beyondallreason.info/battle/3376408
Also any unrated match too.

L-e-x-o-n:
> @L-e-x-o-n can you try deploy this on integration and check a bot match? I have one: https://server5.beyondallreason.info/battle/3376408 Also any unrated match too.
> 
> We cannot use processed boolean because it's possible for match to be processed but not have any rating logs (which triggers the error). I've just added more comments to the code to make it more obvious.

Sorry for the delay, I deployed and tested, it's working!
Thanks

PR #523, opened 2024-11-12
Make OAuth discovery work locally

CollinsSpencer:
Resolves #499 

### Problems

1. If the environment variable `TEI_DOMAIN_NAME` isn't set in local dev, then the `domain_name` local in runtime.exs used the default of `"beyondallreason.info"`.
2. Even if `TEI_DOMAIN_NAME` was set, the manual string interpolation `"https://#{domain_name}"` didn't give the correct urls if TLS/SSL wasn't enabled.
3. This causes the http://localhost:4000/.well-known/oauth-authorization-server url to return incorrect endpoints.

### Solution

* Use `TeiserverWeb.Endpoint.static_url()` [static_url/0 docs](https://hexdocs.pm/phoenix/Phoenix.Endpoint.html#c:static_url/0) to get the base url.
* Since the `domain_name` is set on the endpoint module with `url: [host: domain_name]`, the domain should work the exact same as before in test/prod.
* The uri scheme should also pull in correctly based on the `http:` vs `https:` config used.
  * If just `http:` is used for local dev:
    * http://localhost:4000/.well-known/oauth-authorization-server --> `{ issuer: "http://localhost:4000/" ...`
  * If `http:` and `https:` are used, the https url will take precedence.
    * https://localhost:4040/.well-known/oauth-authorization-server --> `{ issuer: "https://localhost:4040/" ...`
    * http://localhost:4000/.well-known/oauth-authorization-server --> `{ issuer: "https://localhost:4040/" ...`

PR #528, opened 2024-11-19
Fix tests: local oauth discovery

CollinsSpencer:
Fixes the tests failing in the pipeline from https://github.com/beyond-all-reason/teiserver/pull/523


NortySpock:
@CollinsSpencer  Nice! Glad it was a quick fix!

PR #529, opened 2024-11-19
Matchmaking starts autohost

geekingfrog:
Integration with [recoil-host](https://github.com/beyond-all-reason/recoil-autohost)

https://github.com/user-attachments/assets/0514001d-6a94-4779-bff5-49ee6cdde8c6

I've added a few comments after this video was recorded.

### DIFF
--- a/lib/teiserver/autohost/tachyon_handler.ex
+++ b/lib/teiserver/autohost/tachyon_handler.ex
@@ -49,6 +49,19 @@ defmodule Teiserver.Autohost.TachyonHandler do
     {:push, {:text, [start_matchmaking_message |> Jason.encode!()]}, new_state}
   end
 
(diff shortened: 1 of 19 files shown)


lib/teiserver/autohost/tachyon_handler.ex
```diff
@@ -105,11 +141,76 @@ defmodule Teiserver.Autohost.TachyonHandler do
     {:stop, :normal, 1000, [{:text, resp}], state}
   end
 
+  # Generic handler for when a response arrives too late
```
geekingfrog, line 144: I am planning to refactor the transport bits, so that this logic about awaiting for the response and timeout management will be centralised and abstracted out of the Autohost and Player modules.

PR #545, opened 2024-12-21
Add num_matches column

jauggy:
# Feature
After running
```
mix ecto.migrate
```
This will add a new num_matches column to the database and also populate it with values. Whenever a match is rated, the num_matches numer will increase by one, with the exception that if it is a rating override, the num_matches will stay the same.

## Purpose
The num_matches column can be used in the **future** for many possible use cases:

### 1. We can change the rating to become 
```
min(1, num_matches/30) * skill
```
This idea has been approved by Marek: [Source](https://www.reddit.com/r/beyondallreason/comments/1h9fc0v/comment/m10eh2r/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button)

This means that new players start at 0 and converge towards their skill over time. We won't have tons of people stuck at 0 due to negative ratings since there's no more subtraction as with Teifion's formula of `skill-uncertainty`. Having new players start at 0 also means we can create noob lobbies that are max rating 10 instead of max 20. 

We can keep leaderboard rating the same.

### 2. The balancer can classify noobs using num_matches
Generally players prefer new players to be spread evenly across teams. The balancer does this by classifying noobs as someone who has uncertainty > 6.65 and chev level < 4. Using a num_matches to classify noobs is more intuitive and future proof.

Note these are future ideas and are not included in the PR.

### DIFF
--- a/lib/teiserver/account/schemas/rating.ex
+++ b/lib/teiserver/account/schemas/rating.ex
@@ -13,6 +13,7 @@ defmodule Teiserver.Account.Rating do
     field :leaderboard_rating, :float
 
     field :last_updated, :utc_datetime
+    field :num_matches, :integer
   end
 
   @doc false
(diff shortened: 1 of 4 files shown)


L-e-x-o-n:
It's a bit unclear what override does just from the name without following the usage, it might be better to rename it to something more descriptive. If t's just used when rerating matches maybe it should be called something like `update_match_counter,` `rerate` or similar.

L-e-x-o-n:
@geekingfrog I am interested in what you think about this, do you see any issues or possible improvements?

jauggy:
> It's a bit unclear what override does just from the name without following the usage, it might be better to rename it to something more descriptive. If t's just used when rerating matches maybe it should be called something like `update_match_counter,` `rerate` or similar.

OK I can update the name.

PR #551, opened 2024-12-31
WIP: Add new rating system setup task

jauggy:


lib/teiserver/battle/libs/balance_lib.ex
```diff
@@ -704,7 +705,27 @@ defmodule Teiserver.Battle.BalanceLib do
...
+  end
+
+  def new_players_start_at_zero?() do
+    config = Config.get_site_config_cache("hidden.Rating method")
+    config == "start at zero; converge to skill"
+  end
+
+  def get_num_matches_for_rating_to_equal_skill() do
+    Config.get_site_config_cache("profile.Num matches for rating to equal skill")
   end
```
L-e-x-o-n, line 729: Not sure about functions that just call another function, I guess it could make sense if there is some additional logic or some other considerations but I don't think it's true in this case. What do you think?

lib/teiserver/battle/libs/balance_lib.ex
```diff
@@ -704,7 +705,27 @@ defmodule Teiserver.Battle.BalanceLib do
...
+      min(1, num_matches / num_matches_target) * skill
+    else
+      max(skill - uncertainty, 0)
+    end
+  end
+
+  def new_players_start_at_zero?() do
+    config = Config.get_site_config_cache("hidden.Rating method")
+    config == "start at zero; converge to skill"
+  end
```
L-e-x-o-n, line 725: Could this not just be `Config.get_site_config_cache("hidden.Rating method") == "start at zero; converge to skill"`?

L-e-x-o-n:
Left some minor comments about the code, the feature itself is a much bigger change.
I am not sure about removal of uncertainty from calculation, it also makes things like season reset, uncertainty growth over inactivity and similar ideas more difficult.

jauggy:
Discussions still ongoing on discord, but here is some charts on rating distributions under a different system: https://ptb.discord.com/channels/549281623154229250/1336472718584840333/1337953956894212097

20 matches to lose provisional rating seems more sane than 30 because it seems so many people don't play that many matches. And this lines up with other games like counter strike (10 wins or 20 games with 50% win rate).

jauggy:
# Updates to code in PR
I've modified the rating calc to include both num wins and num matches. You lose your provisional rating if you get 10 wins or 20 matches, whichever comes first. The purpose is to make it so smurfs lose provisional rating faster, but non-smurfs experience progression for their first 20 matches no matter what.

# Updates on discussions
**Blodir** is in support of the new rating system and did something similar in FaF

> as reply to the first post, i think that's a good idea yes, iirc we did that in faf matchmaker. Just linearly interpolate mu based on game count or something like that

> not just int he autobalance either, but everywhere such that there's no inconsistencies
> it feels right for a new player to visibly start at zero, and assuming there's no smurfs each new account should be a new player so there's a pretty safe assumption that their skill will not be "average" 

https://ptb.discord.com/channels/549281623154229250/1341639473783636059/1343336677145448559

There was some discussion about trying to explain win predict and that it becomes 50% when team skills are equal.

>  Can just use the mu difference for autobalance because the denom is constant (within the same set of players)

https://ptb.discord.com/channels/549281623154229250/1341639473783636059/1343572438138621952

## Analysis of changing from skill minus uncertainty to skill
Openskill library uses `skill` not `skill minus uncertainty` when calculating win predict. **Vivek**, who is the author of the python version of openskill, did some analysis:
https://ptb.discord.com/channels/549281623154229250/1341639473783636059/1343590512153858078
From what I understand using `skill` is slightly more accurate than using `skill-uncertainty`.
However, I'm not quite sure how he did the analysis. How many matches did he test with? I have asked him to provide more details and will update with comments in the future.

PR #552, opened 2024-12-31
Add uncertainty to player page

jauggy:
## Feature
Removed units produced column (because useless)
Instead added uncertainty column. This is the prematch value.

Screenshot 2025-01-01 083354

### DIFF
--- a/lib/teiserver_web/live/battles/match/show.ex
+++ b/lib/teiserver_web/live/battles/match/show.ex
@@ -444,4 +444,12 @@ defmodule TeiserverWeb.Battle.MatchLive.Show do
      |> assign(:algorithm, value)
      |> get_match()}
   end
+
+  def format_num_matches(num_matches) do
+    cond do
+      num_matches == nil -> ""
+      num_matches <= 1000 -> num_matches
+      num_matches > 1000 -> ">1000"
+    end
+  end
 end
--- a/lib/teiserver_web/live/battles/match/show.ex
+++ b/lib/teiserver_web/live/battles/match/show.ex
@@ -78,16 +78,6 @@ defmodule TeiserverWeb.Battle.MatchLive.Show do
 
       match_name = MatchLib.make_match_name(match)
 
-      members =
-        match.members
-        |> Enum.map(fn member ->
-          Map.merge(member, %{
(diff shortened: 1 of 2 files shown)


L-e-x-o-n:
Can you put this on the left side, after the team column and if possible sort players by rating (after sorting by team)?

jauggy:
Yeah sure will do.

lib/teiserver_web/live/battles/match/show.ex
```diff
@@ -328,14 +328,24 @@ defmodule TeiserverWeb.Battle.MatchLive.Show do
...
     old_uncertainty = uncertainty - uncertainty_change
     old_skill = skill - skill_change
 
+    old_num_matches =
+      if(num_matches == nil) do
+        nil
+      else
+        num_matches - 1
+      end
+
```
L-e-x-o-n, line 343: Something like `if num_matches, do: num_matches - 1, else: nil` seems nicer.

L-e-x-o-n:
> Screenshot 2025-01-08 at 11 04 00 am Added also num_matches. This will show the number of matches you played prior to this match (to be consistent with rating and uncertainty which is prematch). This allows players to understand the balance more since new players will be treated special by the balancer. If you have more than 1000 matches it is obfuscated and just shows >1000.
> 
> Old matches won't have this value, but I tested by manually adjusting `teiserver_game_rating_logs` `value` column e.g.
> 
> ```
> {"skill": 21.948025791764863, "num_matches": 123, "uncertainty": 5.09997450626287, "rating_value": 16.84805128550199, "skill_change": -1.0064758196425458, "uncertainty_change": -0.06046185070328658, "rating_value_change": -0.9460139689392584}
> ```

What is the reasoning behind obfuscating the number if the user has more than 1000 matches, I think people would still want to know it and I can't think of any disadvantages?

jauggy:
I'm happy to show the number. Can reuse the normalise function Teifion made which will use "K" when it gets above 1000. He uses this function for the damage values. Will update on the next push.

jauggy:
Updated.

Screenshot 2025-01-08 at 12 46 31 pm
Using Teifion's normalize function it looks like this.

L-e-x-o-n:
Thanks

PR #563, opened 2025-01-29
Fix team size calculation for balancer server

jauggy:
The balancer will sometimes use large team ratings even if it is a small team game. This is a potential fix. In the old code, team size was calculated as the largest team size of all teams. So if you were in a lobby with 6 players on team 1, then 4 players on team 2, it would use large team ratings. However, this is wrong since after balancing, the teams would become 5v5 and should be small teams.

The calculation should simply divide players by team_count rounded up.

### DIFF
--- a/lib/teiserver/game/servers/balancer_server.ex
+++ b/lib/teiserver/game/servers/balancer_server.ex
@@ -164,14 +164,7 @@ defmodule Teiserver.Game.BalancerServer do
 
   @spec do_make_balance(non_neg_integer(), [T.client()], List.t()) :: map()
   defp do_make_balance(team_count, players, opts) do
-    teams =
-      players
-      |> Enum.group_by(fn c -> c.team_number end)
-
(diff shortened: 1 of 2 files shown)


lib/teiserver/game/servers/balancer_server.ex
```diff
@@ -202,6 +195,14 @@ defmodule Teiserver.Game.BalancerServer do
     end
   end
 
+  # This calculates the team size after balancing, which is important for determining whether a game is small or large team
+  # After balancing the team size will equal the number of players divided by team count rounded up.
+  # After balancing, the team size will be as even as possible.
```
L-e-x-o-n, line 200: After balancing? Doesn't this happen before balancing?
I think the function itself is fine, not sure about the comment.

jauggy: Added more comments. Hope it makes more sense.

Issue #557, opened 2025-01-20
[Feature]: store map data in teiserver


geekingfrog:
### Describe the Feature

For tachyon, with more responsibilities pushed to teiserver, it needs to know a few things about maps. I was thinking of exposing a http endpoint on the API and push some json blob on it.

The main goal is to be able to have valid and relevant maps used in matchmaking (and later, for lobbies).

Something like: `POST server4.beyondallreason.info/teiserver/api/map`
Auth would be using an oauth bearer token `Authorization: Bearer xxxx`

What should the payload look like?
```
{
  "maps": [
    "springName": "Talus_V2",
    "version": "1.3",  // (semver?) strings? floats? int?
    "displayName": "Talus",
    "authors": ["Beherith"],  // array? or just a string?
    "startboxes": [
      {"x0": 0, "y0": 100, "x1": 100, "y1": 200},
      {"x0": 100, "y0": 0, "x1": 200, "y1": 100}
    ],
    // to support map lists, like rotato/duel and so on
    "tags": [
      "1v1", "2v2", "ffa3", "ffa4"
    ]
  ]
}

```

That would be stored in the DB. Any subsequent POST would overwrite the data.

Questions:
* Not sure about the `tags` stuff for map lists, is that enough?
* do we want some kind of versioning on this data? I'd say no since it already exists in a git repo, but that may be useful.

geekingfrog:
From [the discord server](https://discord.com/channels/549281623154229250/564591092360675328/1331377670482563166), here's what the minimum data should look like:

[teiserver_maps.validated.json](https://github.com/user-attachments/files/18510181/teiserver_maps.validated.json)

geekingfrog:
The actual json is live at https://maps-metadata.beyondallreason.dev/latest/teiserver_maps.validated.json

PR #569, opened 2025-02-02
Map api endpoint

geekingfrog:
This introduces the api endpoint for https://github.com/beyond-all-reason/teiserver/issues/557 
This api is restricted using oauth bearer token, and the map endpoint requires `admin.map` scope.

### DIFF
--- a/test/support/fixtures/o_auth/o_auth_fixtures.ex
+++ b/test/support/fixtures/o_auth/o_auth_fixtures.ex
@@ -77,11 +77,16 @@ defmodule Teiserver.OAuthFixtures do
   given a user id, will create a OAuth application and a valid token for that
   user, returning both
   """
-  def setup_token(%{user: user}), do: setup_token(user.id)
-  def setup_token(%{id: id}), do: setup_token(id)
+  def setup_token(user_or_id, opts \\ [])
+  def setup_token(%{user: user}, opts), do: setup_token(user.id, opts)
+  def setup_token(%{id: id}, opts), do: setup_token(id, opts)
 
-  def setup_token(user_id) do
-    app = app_attrs(user_id) |> create_app()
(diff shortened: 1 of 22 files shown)


geekingfrog:
Hold off reviewing that, I have misgiving about adding some data through a migration, I think a better way is through a mix task + some docs.

geekingfrog:
Alright, I got rid of the commit that introduced the migration.
I’ll improve the setup with some mix task and doc in another PR, this one is big enough as it is.

test/support/fixtures/o_auth/o_auth_fixtures.ex
```diff
@@ -77,11 +77,16 @@ defmodule Teiserver.OAuthFixtures do
   given a user id, will create a OAuth application and a valid token for that
   user, returning both
   """
-  def setup_token(%{user: user}), do: setup_token(user.id)
-  def setup_token(%{id: id}), do: setup_token(id)
+  def setup_token(user_or_id, opts \\ [])
+  def setup_token(%{user: user}, opts), do: setup_token(user.id, opts)
+  def setup_token(%{id: id}, opts), do: setup_token(id, opts)
```
L-e-x-o-n, line 82: When would this `def setup_token(%{id: id}, opts), do: setup_token(id, opts)` get called? What map other than user would have the id key for this to match?

geekingfrog: iirc, this is to allow [some setup code](https://github.com/geekingfrog/teiserver/blob/fbf05891476945a759e84b41573037819a826ae4/test/teiserver_web/controllers/tachyon_controller_test.exs#L144) with an access to a user to directly call it. It's more for convenience, so you can pass a user (map with an id) or just the id.

This change adds the option to specify the scopes for the token.

PR #584, opened 2025-02-08
Fix #581

### LINKED ISSUE #581: [Bug]: Invalid handling of basic auth credentials
### Describe the Bug

From: https://datatracker.ietf.org/doc/html/rfc6749#section-2.3.1

> The client identifier is encoded using the "application/x-www-form-urlencoded" encoding algorithm per [Appendix B](https://datatracker.ietf.org/doc/html/rfc6749#appendix-B), and the encoded value is used as the username; the client password is encoded using the same algorithm and used as the password.

So, the `client_id` and `client_secret` are first url encoded, and only then put into basic auth. The header file is effectively:

`"Basic " + base64(urlencode(client_id) + ":" urlencode(client_secret))'`

Teiserver client secrets contain `=` in them, which is urlencoded to `%3D`, if not for that, I wouldn't even notice this.

I've missed it entirely myself and I have bad implementation in autohost that I will fix, that's why it was working right now.

### Reproduce the bug

Can't use the standard oauth2 compliant libraries like Rust's `oauth2` crate because they handle it correctly: https://github.com/ramosbugs/oauth2-rs/blob/f3424b4b2190c83c6d031fdc71eed2351d49e0df/src/endpoint.rs#L115-L127

### Screenshots

_No response_

### Additional context

_No response_

geekingfrog:
Because ofc f*inck oauth does basic auth in a special manner

### DIFF
--- a/lib/teiserver/o_auth.ex
+++ b/lib/teiserver/o_auth.ex
@@ -467,6 +467,37 @@ defmodule Teiserver.OAuth do
     count
   end
 
+  # Because OAuth does some special basic auth handling, see:
+  # https://datatracker.ietf.org/doc/html/rfc6749#section-2.3.1
+  # and especially the Appendix B:
+  # https://datatracker.ietf.org/doc/html/rfc6749#appendix-B
+  @doc """
+  Similar to Plug.BasicAuth.encode_basic_auth but compliant with OAuth special handling
+  Takes client id, client secret and returns the basic auth header
+  """
+  @spec encode_basic_auth(String.t(), String.t()) :: String.t()
+  def encode_basic_auth(client_id, client_secret) do
+    encoded =
+      Base.encode64("#{URI.encode_www_form(client_id)}:#{URI.encode_www_form(client_secret)}")
+
+    "Basic #{encoded}"
+  end
+
+  @doc """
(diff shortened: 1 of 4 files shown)


L-e-x-o-n:
What about Plug-BasicAuth.encode_basic_auth?

geekingfrog:
This is what I was using before, but as noted in the bug report, OAuth does basic auth differently by first www_form_urlencoding each component, and then applying base64. The normal basic auth mechanism doesn't have the urlencoding part.
My implementation is lifted from `Plug.BasicAuth`, with this additional step.

L-e-x-o-n:
> This is what I was using before, but as noted in the bug report, OAuth does basic auth differently by first www_form_urlencoding each component, and then applying base64. The normal basic auth mechanism doesn't have the urlencoding part. My implementation is lifted from `Plug.BasicAuth`, with this additional step.

I understand that, but is the old `Plug.BasicAuth.encode_basic_auth` still necessary or can it be deleted?

geekingfrog:
It is coming from the [plug module](https://hexdocs.pm/plug/Plug.BasicAuth.html), which is required for many other things, anything http connection related. So there’s nothing more to do.

Issue #561, opened 2025-01-26
[Feature]: implement tachyon `system/serverStats` event


geekingfrog:
### Describe the Feature

Add the code to handle the requests for `system/serverStats` that returns the player count.
This can be obtained with [Registry.count/1](https://hexdocs.pm/elixir/Registry.html#count/1) using the [session registry](https://github.com/geekingfrog/teiserver/blob/778450852dd63e087174a62853aae563b6723479/lib/teiserver/player/session_registry.ex)

PR #586, opened 2025-02-09
Tachyon system/serverStats

L-e-x-o-n:
Completes #561

### DIFF
--- a/lib/teiserver/player/session_registry.ex
+++ b/lib/teiserver/player/session_registry.ex
@@ -10,7 +10,7 @@ defmodule Teiserver.Player.SessionRegistry do
   end
 
   @doc """
-  how to reach a given session
+  How to reach a given session
   """
   @spec via_tuple(T.userid()) :: GenServer.name()
   def via_tuple(user_id) do
@@ -34,5 +34,9 @@ defmodule Teiserver.Player.SessionRegistry do
       id: __MODULE__,
       start: {__MODULE__, :start_link, []}
     )
+  end
+
+  def count() do
+    Horde.Registry.count(__MODULE__)
   end
 end
--- a/lib/teiserver/player/tachyon_handler.ex
+++ b/lib/teiserver/player/tachyon_handler.ex
@@ -140,6 +140,12 @@ defmodule Teiserver.Player.TachyonHandler do
     {:stop, :normal, state}
   end
 
+  def handle_command("system/serverStats" = cmd_id, "request", _message_id, _message, state) do
+    user_count = Teiserver.Player.SessionRegistry.count()
+
+    {:response, cmd_id, %{userCount: user_count}, state}
+  end
+
   def handle_command("matchmaking/list" = cmd_id, "request", message_id, _message, state) do
     queues =
       Matchmaking.list_queues()
--- a/test/support/tachyon.ex
+++ b/test/support/tachyon.ex
@@ -1,6 +1,7 @@
 defmodule Teiserver.Support.Tachyon do
   alias WebsocketSyncClient, as: WSC
   alias Teiserver.OAuthFixtures
+  alias Teiserver.Player
 
   def setup_client(_context), do: setup_client()
 
@@ -29,7 +30,20 @@ defmodule Teiserver.Support.Tachyon do
       {:ok, _user_updated} = recv_message(client)
     end
 
-    ExUnit.Callbacks.on_exit(fn -> WSC.disconnect(client) end)
+    ExUnit.Callbacks.on_exit(fn ->
+      WSC.disconnect(client)
+
+      case Player.lookup_session(token.owner_id) do
+        nil -> :ok
+        pid -> Process.exit(pid, :test_cleanup)
+      end
+
+      poll_until(
+        fn -> Player.lookup_session(token.owner_id) end,
+        fn x -> is_nil(x) end
+      )
+    end)
+
     client
   end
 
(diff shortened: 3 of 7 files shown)


lib/teiserver/player/session_registry.ex
```diff
@@ -35,4 +35,8 @@ defmodule Teiserver.Player.SessionRegistry do
       start: {__MODULE__, :start_link, []}
     )
   end
+
+  def get_count() do
```
geekingfrog, line 39: I’d go with `count` to mimic the `Registry` api.

lib/teiserver/player/tachyon_handler.ex
```diff
@@ -140,6 +140,14 @@ defmodule Teiserver.Player.TachyonHandler do
     {:stop, :normal, state}
   end
 
+  def handle_command("system/serverStats" = cmd_id, "request", message_id, _message, state) do
+    user_count = Teiserver.Player.SessionRegistry.get_count()
+
+    resp = Schema.response(cmd_id, message_id, %{userCount: user_count}) |> Jason.encode!()
+
+    {:reply, :ok, {:text, resp}, state}
```
geekingfrog, line 148: That works, but since https://github.com/beyond-all-reason/teiserver/pull/531 there's a shorter way to express that:
```elixir
{:response, cmd_id, %{userCount: user_count}, state}
```
so you don't need to deal with any json encoding by hand.

geekingfrog (commented):
Some tiny comments.
But you are missing a test for the feature.
You should add a function `server_stats` in [Support.Tachyon](https://github.com/geekingfrog/teiserver/blob/30771c45e97d45e24caf0d7575a177c44fdd2c2c/test/support/tachyon.ex) that sends the message and wait for the response. There are plenty of other function doing similar things, [list_queues!](https://github.com/geekingfrog/teiserver/blob/30771c45e97d45e24caf0d7575a177c44fdd2c2c/test/support/tachyon.ex#L184-L206) for example.

Then you should create a new file under `test/teiserver_web/tachyon/system_test.exs`, that `use TeiserverWeb.ConnCase`.

Then, you can follow the same pattern as the [pairing tests](https://github.com/geekingfrog/teiserver/blob/30771c45e97d45e24caf0d7575a177c44fdd2c2c/test/teiserver_web/tachyon/matchmaking_test.exs#L156-L176) to have several user connected.
You can then test that the request returns the correct response, then disconnect a client and try again to see the count updated.
Keep in mind that you’ll need the response schema in the repo as well since the end to end tests are validating the response against the expected schema.

lib/teiserver/player/session_registry.ex
L-e-x-o-n: Fixed

lib/teiserver/player/tachyon_handler.ex
L-e-x-o-n: Nice, thanks.

test/teiserver_web/tachyon/system_test.exs
```diff
@@ -0,0 +1,40 @@
...
+      user = Central.Helpers.GeneralTestLib.make_user(%{"data" => %{"roles" => ["Verified"]}})
+      token = OAuthFixtures.token_attrs(user.id, app) |> OAuthFixtures.create_token()
+      client = Tachyon.connect(token)
+      {:ok, %{user: user, token: token, client: client}}
+    end
+
+    setup [:setup_app]
+
+    test "works", %{app: app} do
+      {:ok, %{user: _user1, client: client1}} = setup_user(app)
```
geekingfrog, line 28: nitpick: you don't need to pattern match on `user` if you don't use it.

test/teiserver_web/tachyon/system_test.exs
```diff
@@ -0,0 +1,40 @@
...
+      {:ok, %{user: _user2, client: client2}} = setup_user(app)
+
+      assert %{"data" => %{"userCount" => 2}} = Tachyon.server_stats!(client1)
+      assert %{"data" => %{"userCount" => 2}} = Tachyon.server_stats!(client2)
+
+      Tachyon.disconnect!(client2)
+      :timer.sleep(1)
```
geekingfrog, line 35: This timer is brittle and will lead to flakyness.
I suggest instead to use `Tachyon.poll_until` to check continuously until the response to the second server stat becomes 1 (or times out).

test/teiserver_web/tachyon/system_test.exs
L-e-x-o-n: Ah, copy pasted this one from pairing test without checking... fixed.

test/teiserver_web/tachyon/system_test.exs
L-e-x-o-n: Done

geekingfrog (approved):

test/support/tachyon.ex
```diff
@@ -29,7 +30,20 @@ defmodule Teiserver.Support.Tachyon do
...
+    ExUnit.Callbacks.on_exit(fn ->
+      WSC.disconnect(client)
+
+      case Player.lookup_session(token.owner_id) do
+        nil -> :ok
+        pid -> Process.exit(pid, :test_cleanup)
+      end
+
+      poll_until(
+        fn -> Player.lookup_connection(token.owner_id) end,
```
geekingfrog, line 42: This should be `lookup_session`, to ensure the registry is updated.

geekingfrog (approved):

PR #588, opened 2025-02-10
Extend battle/start message with engine, game and map

### LINKED ISSUE #574: [Feature]: Extend `battle/start` message
### Describe the Feature

Now that [this PR](https://github.com/beyond-all-reason/tachyon/pull/44) is merged, the [battle/start](https://github.com/beyond-all-reason/tachyon/blob/master/docs/schema/battle.md#start) response should have more fields to specify the engine version, game version and map.

* The function [Teiserver.Player.battle_start](https://github.com/geekingfrog/teiserver/blob/b0a9d92e76bd2a2bf0dcb6e29acef6b6f7e7b72e/lib/teiserver/matchmaking/pairing_room.ex#L148-L150) should be modified so that it takes into account these fields when sending the message.
* Update the corresponding json schema

L-e-x-o-n:
Resolves #574

### DIFF
--- a/lib/teiserver/autohost/tachyon_handler.ex
+++ b/lib/teiserver/autohost/tachyon_handler.ex
@@ -20,7 +20,10 @@ defmodule Teiserver.Autohost.TachyonHandler do
 
   @type start_response :: %{
           ips: [String.t()],
-          port: integer()
+          port: integer(),
+          engine: %{version: String.t()},
+          game: %{springName: String.t()},
+          map: %{springName: String.t()}
         }
 
   @impl Handler
--- a/test/teiserver_web/tachyon/matchmaking_test.exs
+++ b/test/teiserver_web/tachyon/matchmaking_test.exs
@@ -436,7 +436,6 @@ defmodule Teiserver.Tachyon.MatchmakingTest do
         for ally_team <- data["allyTeams"],
             team <- ally_team["teams"],
             player <- team["players"] do
-          # TODO: actually run json schema validation on the payloadd instead of that
           assert Map.has_key?(player, "name")
           assert Map.has_key?(player, "password")
           player["userId"]
@@ -450,8 +449,16 @@ defmodule Teiserver.Tachyon.MatchmakingTest do
       end
 
       for client <- clients do
-        # TODO: same, json schema validation here
-        assert %{"commandId" => "battle/start", "data" => %{"ip" => _ip, "port" => _port}} =
+        assert %{
+                 "commandId" => "battle/start",
+                 "data" => %{
+                   "ip" => _ip,
+                   "port" => _port,
+                   "engine" => %{"version" => "105.1.1-2590-gb9462a0 bar"},
+                   "game" => %{"springName" => "Beyond All Reason test-26929-d709d32"},
+                   "map" => %{"springName" => "Red Comet Remake 1.8"}
+                 }
+               } =
                  Tachyon.recv_message!(client)
       end
     end
--- a/lib/teiserver/lobby/libs/lobby_lib.ex
+++ b/lib/teiserver/lobby/libs/lobby_lib.ex
@@ -135,7 +135,7 @@ defmodule Teiserver.Lobby.LobbyLib do
     cast_lobby(lobby_id, {:update_values, new_values})
   end
 
-  @spec rename_lobby(T.lobby_id(), String.t(), T.userid()) :: :ok | nil
+  @spec rename_lobby(T.lobby_id(), String.t(), T.userid() | nil) :: :ok | nil
   def rename_lobby(lobby_id, new_base_name, renamer_id) do
     cast_lobby(lobby_id, {:rename_lobby, new_base_name, renamer_id})
   end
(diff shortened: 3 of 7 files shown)


lib/teiserver/autohost/tachyon_handler.ex
```diff
@@ -20,7 +20,10 @@ defmodule Teiserver.Autohost.TachyonHandler do
 
   @type start_response :: %{
           ips: [String.t()],
-          port: integer()
+          port: integer(),
+          engine: %{String.t() => String.t()},
+          game: %{String.t() => String.t()},
+          map: %{String.t() => String.t()}
```
geekingfrog, line 26: Although these types are valid they are not great. You should prefer using known keys with atoms like so: `%{version: String.t()}`

lib/teiserver/autohost/tachyon_handler.ex
```diff
@@ -173,7 +176,10 @@ defmodule Teiserver.Autohost.TachyonHandler do
        {:ok,
         %{
           ips: data["ips"],
-          port: data["port"]
+          port: data["port"],
+          engine: %{"version" => data["engine"]},
+          game: %{"springName" => data["game"]},
+          map: %{"springName" => data["map"]}
```
geekingfrog, line 182: prefer using atoms for known keys, they are preallocated and nicer to use. When converted to json, they are automatically converted to string at that point.

test/teiserver_web/tachyon/matchmaking_test.exs
```diff
@@ -427,7 +427,10 @@ defmodule Teiserver.Tachyon.MatchmakingTest do
 
       host_data = %{
         ips: ["127.0.0.1"],
-        port: 48912
+        port: 48912,
+        engine: "recoil",
+        game: "BAR",
+        map: "Altored Divide"
```
geekingfrog, line 433: This is incorrect. The response from the autohost does NOT have these keys set, only the ips and port. See https://github.com/beyond-all-reason/tachyon/blob/master/docs/schema/autohost.md#response-8

What you need to check is on lines 454, where we call `recv_message!` for each client, that *these* responses have the correct data in them.
You can also remove the comments about running schema validation of these responses, this is now done

geekingfrog (changes requested):

lib/teiserver/autohost/tachyon_handler.ex
L-e-x-o-n: Done

lib/teiserver/autohost/tachyon_handler.ex
L-e-x-o-n: Done

test/teiserver_web/tachyon/matchmaking_test.exs
L-e-x-o-n: Done

lib/teiserver/autohost/tachyon_handler.ex
```diff
@@ -173,7 +176,11 @@ defmodule Teiserver.Autohost.TachyonHandler do
        {:ok,
         %{
           ips: data["ips"],
-          port: data["port"]
+          port: data["port"],
+          # TODO Geth these values from somewhere else
+          engine: %{version: "105.1.1-2590-gb9462a0 bar"},
+          game: %{springName: "Beyond All Reason test-26929-d709d32"},
+          map: %{springName: "Red Comet Remake 1.8"}
```
L-e-x-o-n, line 183: Is this fine for now? If not where else should I get these from?

geekingfrog: The flow should be the following:

* pairing room has got all the ready notification and create a start script from its data.
* it sends this start script to an autohost that starts a battle
* with the response of the autohost, it knows the ips and port
* the pairing room process can then send all the required data to the players, by merging
  * the ip and port from the autohost
  * the engine, map and game from the start script/internal parameters
This merging should be done around there: https://github.com/geekingfrog/teiserver/blob/master/lib/teiserver/matchmaking/pairing_room.ex#L143

For now this is how the information should flow.
I think in the future, we should refactor that a bit so that the battle process is responsible for sending this data to the player, if you feel like it, you can do that in another commit.

test/teiserver_web/tachyon/matchmaking_test.exs
```diff
@@ -450,8 +449,16 @@ defmodule Teiserver.Tachyon.MatchmakingTest do
...
-        # TODO: same, json schema validation here
-        assert %{"commandId" => "battle/start", "data" => %{"ip" => _ip, "port" => _port}} =
+        assert %{
+                 "commandId" => "battle/start",
+                 "data" => %{
+                   "ip" => _ip,
+                   "port" => _port,
+                   "engine" => %{"version" => "105.1.1-2590-gb9462a0 bar"},
+                   "game" => %{"springName" => "Beyond All Reason test-26929-d709d32"},
+                   "map" => %{"springName" => "Red Comet Remake 1.8"}
```
geekingfrog, line 459: :+1:

lib/teiserver/autohost/tachyon_handler.ex
L-e-x-o-n: Ah I see now. The latest changes should work then.

geekingfrog (approved):

PR #593, opened 2025-02-12
Messaging dm subscription

geekingfrog:
All the messaging stuff for DM. The scary +935 is a bit misleading, the addition of the json schema inflates the count quite a bit.
I believe there's going to be a special process for messaging later for lobbies and parties, but dms are simpler and can be handled this way, at least for now.

### DIFF
--- a/lib/teiserver/player/session.ex
+++ b/lib/teiserver/player/session.ex
@@ -15,6 +15,7 @@ defmodule Teiserver.Player.Session do
 
   alias Teiserver.Data.Types, as: T
   alias Teiserver.{Player, Matchmaking, Messaging}
+  alias Teiserver.Helpers.BoundedQueue, as: BQ
 
   @type conn_state :: :connected | :reconnecting | :disconnected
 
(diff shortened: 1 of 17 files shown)


lib/teiserver/player/session.ex
```diff
@@ -34,13 +35,56 @@ defmodule Teiserver.Player.Session do
...
 
+  @type messaging_state :: %{
+          store_messages?: boolean(),
+          subscribed?: boolean(),
+          # for simplicity, only hold one buffer for everything. This may lead to
+          # problems if a few sources are really noisy, they will force out
+          # the other messages. We can deal with that later with a smaller
+          # buffer per source, and the added complexity of having to limit
+          # that total size
+          buffer: BQ.t(Messaging.message())
```
bcdrme, line 46: I think it's fine honestly, especially if we implement some sort of rate limit on the websocket connection itself to prevent bad actors from spamming

L-e-x-o-n (approved):

PR #595, opened 2025-02-13
Remove deep caching hashmap in balancer server

### LINKED ISSUE #576: [Bug]: High memory usage of BalancerServer during large lagames
### Describe the Bug

# Introduction
The `Teiserver.Game.BalancerServer` is responsible for balancing games, one is started for each lobby.
Whenever it balances a lobby (by receiving the balance request from SPADS) it creates a hash consisting of team count, players, parties and current balance algorithm and keeps a map of past balance states and results accessible via that hash. This is useful because it allows quickly calculating the hash and returning existing balance results if the hash is present instead of calculating new balance results on each request (which can be expensive to calculate, players also like calling `!balance` often).

# The problem
Importantly the `hashes` map of states is only cleared after the match is rated (to force balance recalculation with new, updated rating).
This is not normally an issue because most lobbies are ranked and limited in number of players, however it becomes a serious issue in special event lobbies which can contain up to 160 players and are always unranked.
During these events the hashes state list only keeps growing and because the lobby is unranked it never gets cleared. With 160 players each balance result can get quite large, they quickly start adding up using 10s of GBs of memory until they consume all the available memory and the whole teiserver service gets killed 

# The solution
The solution should be quite simple, set a limit to how many hashes can be saved at any time and discard the older ones when the limit is reached and a new one is being added. Additionally it is probably worth considering adding a special condition that avoids saving any hashes if the player count is greater than 50/80/100 because at that scale the chances of exactly the same players being present to make use of the hash saved state are quite low anyway so might as well save some memory.

### Reproduce the bug

Reproduced multiple times during the last 40k event.

NortySpock:
Fixes the unbounded memory consumption behavior described by #576 ; adds tests for the changed function calls.

Now BalancerServer will only retain the most recent balance result, so long as it has not been cleared by another operation like `:reset_hashes` or `:set`.

Probably needs a spin on the testing server to verify we haven't broken a function call that was depending on `handle_call(:report_state)` returning the number of hashes, but other than that I think I've written all the unit-tests for all of the parts that changed.

### DIFF
--- a/lib/teiserver/game/servers/balancer_server.ex
+++ b/lib/teiserver/game/servers/balancer_server.ex
@@ -199,10 +199,9 @@ defmodule Teiserver.Game.BalancerServer do
     |> Base.encode64()
   end
 
-  # This function is public only for testing; it should not be called otherwise
-  # long term, there in interest in confirming this is a stateless pure function, which would make this easier to test
+  # long term, there in interest in confirming do_make_balance is a stateless pure function, which would make it easier to test
   @spec do_make_balance(non_neg_integer(), [T.client()], List.t()) :: map()
-  def do_make_balance(team_count, players, opts) do
+  defp do_make_balance(team_count, players, opts) do
     team_size = calculate_team_size(team_count, players)
 
     # Use Large Team ratings when balancing Team FFA
--- a/lib/teiserver/game/servers/balancer_server.ex
+++ b/lib/teiserver/game/servers/balancer_server.ex
@@ -95,7 +95,7 @@ defmodule Teiserver.Game.BalancerServer do
 
   @impl true
   def handle_cast(:reset_hashes, state) do
-    {:noreply, %{state | hashes: %{}}}
+    {:noreply, %{state | last_balance_hash: nil, last_balance_result: nil}}
   end
 
   def handle_cast({:set_algorithm, algorithm}, state) do
@@ -118,7 +118,8 @@ defmodule Teiserver.Game.BalancerServer do
         true ->
           state
           |> Map.put(key, value)
-          |> Map.put(:hashes, %{})
+          |> Map.put(:last_balance_hash, nil)
+          |> Map.put(:last_balance_result, nil)
 
         false ->
           state
--- a/lib/teiserver/game/servers/balancer_server.ex
+++ b/lib/teiserver/game/servers/balancer_server.ex
@@ -27,11 +27,7 @@ defmodule Teiserver.Game.BalancerServer do
           mean_diff_max: state.mean_diff_max,
           stddev_diff_max: state.stddev_diff_max,
           fuzz_multiplier: state.fuzz_multiplier,
-          shuffle_first_pick: state.shuffle_first_pick,
-          last_balance_hash_cache_hit: state.last_balance_hash_cache_hit,
-          last_balance_hash_cache_miss: state.last_balance_hash_cache_miss,
-          last_balance_hash: state.last_balance_hash,
-          last_balance_result: state.last_balance_result
+          shuffle_first_pick: state.shuffle_first_pick
         ]
 
     {balance, new_state} = make_balance(team_count, state, opts)
@@ -51,11 +47,7 @@ defmodule Teiserver.Game.BalancerServer do
           mean_diff_max: state.mean_diff_max,
           stddev_diff_max: state.stddev_diff_max,
           fuzz_multiplier: state.fuzz_multiplier,
-          shuffle_first_pick: state.shuffle_first_pick,
-          last_balance_hash_cache_hit: state.last_balance_hash_cache_hit,
-          last_balance_hash_cache_miss: state.last_balance_hash_cache_miss,
-          last_balance_hash: state.last_balance_hash,
-          last_balance_result: state.last_balance_result
+          shuffle_first_pick: state.shuffle_first_pick
         ]
 
     {balance, new_state} = make_balance(team_count, state, opts, players)
--- a/lib/teiserver/game/servers/balancer_server.ex
+++ b/lib/teiserver/game/servers/balancer_server.ex
@@ -9,7 +9,6 @@ defmodule Teiserver.Game.BalancerServer do
   @tick_interval 2_000
   # Balance algos that allow fuzz; randomness will be added to match rating before processing
   @algos_allowing_fuzz ~w(loser_picks force_party)
-
   @spec start_link(List.t()) :: :ignore | {:error, any} | {:ok, pid}
   def start_link(opts) do
     GenServer.start_link(__MODULE__, opts[:data], [])
@@ -28,7 +27,9 @@ defmodule Teiserver.Game.BalancerServer do
           mean_diff_max: state.mean_diff_max,
           stddev_diff_max: state.stddev_diff_max,
           fuzz_multiplier: state.fuzz_multiplier,
-          shuffle_first_pick: state.shuffle_first_pick
+          shuffle_first_pick: state.shuffle_first_pick,
+          last_balance_hash_cache_hit: state.last_balance_hash_cache_hit,
+          last_balance_hash_cache_miss: state.last_balance_hash_cache_miss
         ]
 
     {balance, new_state} = make_balance(team_count, state, opts)
@@ -48,7 +49,9 @@ defmodule Teiserver.Game.BalancerServer do
           mean_diff_max: state.mean_diff_max,
           stddev_diff_max: state.stddev_diff_max,
           fuzz_multiplier: state.fuzz_multiplier,
-          shuffle_first_pick: state.shuffle_first_pick
+          shuffle_first_pick: state.shuffle_first_pick,
+          last_balance_hash_cache_hit: state.last_balance_hash_cache_hit,
+          last_balance_hash_cache_miss: state.last_balance_hash_cache_miss
         ]
 
     {balance, new_state} = make_balance(team_count, state, opts, players)
@@ -74,7 +77,9 @@ defmodule Teiserver.Game.BalancerServer do
       mean_diff_max: state.mean_diff_max,
       fuzz_multiplier: state.fuzz_multiplier,
       stddev_diff_max: state.stddev_diff_max,
-      shuffle_first_pick: state.shuffle_first_pick
+      shuffle_first_pick: state.shuffle_first_pick,
+      last_balance_hash_cache_hit: state.last_balance_hash_cache_hit,
+      last_balance_hash_cache_miss: state.last_balance_hash_cache_miss
     }
 
     {:reply, result, state}
--- a/lib/teiserver/game/servers/balancer_server.ex
+++ b/lib/teiserver/game/servers/balancer_server.ex
@@ -35,6 +35,26 @@ defmodule Teiserver.Game.BalancerServer do
     {:reply, balance, new_state}
   end
 
+  @impl true
+  # http://planetspads.free.fr/spads/doc/spadsPluginApiDoc.html#balanceBattle-self-players-bots-clanMode-nbTeams-teamSize
+  def handle_call({:make_balance, team_count, call_opts, players}, _from, state) do
+    opts =
+      call_opts ++
+        [
+          algorithm: state.algorithm,
+          max_deviation: state.max_deviation,
+          rating_lower_boundary: state.rating_lower_boundary,
+          rating_upper_boundary: state.rating_upper_boundary,
+          mean_diff_max: state.mean_diff_max,
+          stddev_diff_max: state.stddev_diff_max,
+          fuzz_multiplier: state.fuzz_multiplier,
+          shuffle_first_pick: state.shuffle_first_pick
+        ]
+
+    {balance, new_state} = make_balance(team_count, state, opts, players)
+    {:reply, balance, new_state}
+  end
+
   def handle_call(:get_balance_mode, _from, %{last_balance_hash: hash} = state) do
     result = state.hashes[hash]
     {:reply, result.balance_mode, state}
@@ -109,28 +129,19 @@ defmodule Teiserver.Game.BalancerServer do
     {:noreply, state}
   end
 
-  # def handle_info({:lobby_update, :updated_client_battlestatus, _lobby_id, {_client, _reason}}, state) do
-  #   {:noreply, state}
-  # end
-
-  # def handle_info({:lobby_update, :add_user, _lobby_id, _userid}, state) do
-  #   {:noreply, state}
-  # end
-
-  # def handle_info({:lobby_update, _, _, _}, state), do: {:noreply, state}
-
-  # def handle_info({:host_update, _userid, _host_data}, state) do
-  #   {:noreply, state}
-  # end
-
-  # def handle_info(%{channel: "teiserver_server"}, state) do
-  #   {:noreply, state}
-  # end
-
   @spec make_balance(non_neg_integer(), T.balance_server_state(), list()) ::
           {map(), T.balance_server_state()}
   defp make_balance(team_count, state, opts) do
     players = Battle.list_lobby_players(state.lobby_id)
+
+    result = make_balance(team_count, state, opts, players)
+    result
+  end
+
+  # This function is public only for testing; it should not be called otherwise
+  @spec make_balance(non_neg_integer(), T.balance_server_state(), list(), list()) ::
+          {map(), T.balance_server_state()}
+  def make_balance(team_count, state, opts, players) do
     hash = make_player_hash(team_count, players, opts)
 
     if hash == state.last_balance_hash do
--- a/lib/teiserver/game/servers/balancer_server.ex
+++ b/lib/teiserver/game/servers/balancer_server.ex
@@ -29,7 +29,9 @@ defmodule Teiserver.Game.BalancerServer do
           fuzz_multiplier: state.fuzz_multiplier,
           shuffle_first_pick: state.shuffle_first_pick,
           last_balance_hash_cache_hit: state.last_balance_hash_cache_hit,
-          last_balance_hash_cache_miss: state.last_balance_hash_cache_miss
+          last_balance_hash_cache_miss: state.last_balance_hash_cache_miss,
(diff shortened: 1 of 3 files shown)


lib/teiserver/game/servers/balancer_server.ex
```diff
@@ -134,17 +133,31 @@ defmodule Teiserver.Game.BalancerServer do
     players = Battle.list_lobby_players(state.lobby_id)
```
NortySpock, line 133: I'm sort of stuck here, I don't know how to mock this to return a fake list of players, so I'm ending up trying to spin up a real lobby with real players in order to try to get the BalancerServer to actually operate.

Not trying to test this exhaustively, just trying to smoke-test it.

test/teiserver/game/balancer_server_test.exs
```diff
@@ -36,6 +37,54 @@ defmodule Teiserver.Game.BalancerServerTest do
...
+
+    {:ok, start_link_results} = BalancerServer.start_link(data: balance_server_init_results)
+
+    assert is_pid(start_link_results)
+    state = :sys.get_state(start_link_results)
+
+    state = GenServer.call(start_link_results, :report_state)
+    dbg(state)
+    # list the players in the lobby
+    players = Teiserver.Battle.list_lobby_players(lobby.id)
```
NortySpock, line 80: I thought I had managed to get the lobby to actually have players in it, but apparently not. (see my `assert Enum.any?(players)` below.)

Again, not trying to exhaustively test the BalancerServer, just trying to prove that I haven't obviously broken the BalancerServer in the changes I have made. I'm used to writing one or two unit tests or integration tests in order to do so, so that's what I'm attempting to do here.

I do not yet have Chobby or another client working with developer options yet; not sure if that would have made this integration testing easier or harder. I'm open to doing manual tests if necessary, but I'm also open to writing a workable scaffold for later automated tests.

geekingfrog: grmbl, this code is indeed quite couple with the lobby, which is sad.
You don't get any player because all tests run in their own isolated sandboxes, and the default state has no user.
You should instead use some [test fixtures](https://github.com/geekingfrog/teiserver/blob/0adfc477f5c1138f93a56821a90aa2c6597f2bfd/lib/central/general_test_lib.ex#L23-L43) to create the users you want, in the state you want.

Btw, I would say that spinning up a genserver with `start_link` is the correct way to go, don't call `init` directly.

A potential refactoring you could do is to make `make_balance` public, with a comment saying that it's public only for tests.
Then you can overload the function to either take a lobby id, in which case it queries the list of player, or a list of players, in which case it actually do the balancing and return the `{result, new_state}`. This way, you could test that function alone, without having to setup the genserver, and the associated lobby.

test/teiserver/game/balancer_server_test.exs
```diff
@@ -36,7 +37,49 @@ defmodule Teiserver.Game.BalancerServerTest do
...
+      )
+
+    dbg(second_balance_pass_hash_reuse)
+
+    assert second_balance_pass_hash_reuse.hash == first_balance_pass.hash
+
+    third_balance_pass_with_more_players =
+      GenServer.call(
+        start_link_results,
+        {:make_balance, team_count, [], create_fake_players(160)},
```
NortySpock, line 74: Currently failing because player 5 does not have a valid rank.

 I'll work on the test handler to create users as needed next, but refactoring the make_balance handler to be public and allow players to be passed in really made iterating through the test process easier.

geekingfrog: Sounds good.
Also, having test fixtures to create users in specific state is going to be helpful in general. This is an area I'm not too familiar with.

The general plan for balance is to reduce it to a pure function that takes a list of players and their attributes like OS, playtime and ranks, along with a specification of balancing, like the number of team and the number of player per team. This may be done through a network call to another service, or a pure elixir fallback. But I want to decouple all state concerns from the balancing algorithm itself. In that sense, focusing on a public `make_balance` is probably a good move.

test/teiserver/game/balancer_server_test.exs
```diff
@@ -1,9 +1,56 @@
...
+      "password" => params["password"] || "password",
+      "password_confirmation" => params["password"] || "password",
+      "data" => params["data"] || %{}
+    }
+
+    {:ok, u} =
+      Teiserver.Account.create_user(requested_user)
+  end
+
+  @spec make_users_with_ranks([non_neg_integer()]) ::
```
NortySpock, line 25: Here is a test fixture to create "a list of users with the indicated ranks".

test/teiserver/game/balancer_server_test.exs
```diff
@@ -36,7 +83,147 @@ defmodule Teiserver.Game.BalancerServerTest do
...
+      GenServer.call(
+        start_link_results,
+        {:make_balance, team_count, [], players},
+        5000
+      )
+
+    state3 = GenServer.call(start_link_results, :report_state, 5000)
+    dbg(state3)
+
+    assert second_balance_pass_hash_reuse.hash == first_balance_pass.hash
```
NortySpock, line 127: Still debugging why this fails when I passed in the same state and player list. :/

test/teiserver/game/balancer_server_test.exs
```diff
@@ -36,7 +83,147 @@ defmodule Teiserver.Game.BalancerServerTest do
     assert result == 4
   end
 
+  test "load some players in and run a balance pass or two" do
+    team_count = 2
+
+    arbitrary_rank_for_all_test_users = 10
```
NortySpock, line 89:
```suggestion

```

Oops, I'll remove this line.

test/teiserver/game/balancer_server_test.exs
```diff
@@ -36,7 +83,147 @@ defmodule Teiserver.Game.BalancerServerTest do
...
+        :get_current_balance,
+        5000
+      )
+
+    refute get_balance_mode_returns_value == nil
+  end
+
   defp create_fake_players(count) do
-    1..count |> Enum.map(fn _ -> %{} end)
+    1..count |> Enum.map(fn iter -> %{iter: iter} end)
```
NortySpock, line 227: I'll probably revert this function to the original form; I thought I could make it serve double-duty but the needs of the test harness evolved beyond simple tweaks

lib/teiserver/game/servers/balancer_server.ex
```diff
@@ -28,34 +27,75 @@ defmodule Teiserver.Game.BalancerServer do
...
+        state.last_balance_hash == hash -> state.last_balance_result
+        true -> nil
+      end
+
     {:reply, result, state}
   end
 
   def handle_call(:report_state, _from, state) do
     result = %{
-      hashes: Enum.count(state.hashes),
```
NortySpock, line 50: Oh ... Technically this becomes an API change. I guess I'll check if anything acts on the number of hashes? Does anything do any reporting based on that?

lib/teiserver/game/servers/balancer_server.ex
```diff
@@ -162,8 +206,10 @@ defmodule Teiserver.Game.BalancerServer do
     |> Base.encode64()
   end
 
+  # This function is public only for testing; it should not be called otherwise
+  # long term, there in interest in confirming this is a stateless pure function, which would make this easier to test
```
NortySpock, line 210: If I get the stateful tests working at all, it should be a snap to go back and write one or two tests of this pure function.

test/teiserver/game/balancer_server_test.exs
geekingfrog: This is because the hash is computed with, amongst other things, the following options:
```elixir
          last_balance_hash_cache_hit: state.last_balance_hash_cache_hit,
          last_balance_hash_cache_miss: state.last_balance_hash_cache_miss,
          last_balance_hash: state.last_balance_hash,
          last_balance_result: state.last_balance_result
```
that is incorrect. The only thing that should be included in the hash are the input parameters for making balance, because that's basically what memoization is. If you include other stuff for the hash, it loses its purpose.

NortySpock: Thanks for the tip, my mistake. Reverting to the original `opts` that were passed in to make_balance fixed the issue. (still making sure I've dotted my i's and crossed my t's before I mark this PR as ready for final review)

lib/teiserver/game/servers/balancer_server.ex
NortySpock: Looks like nothing inside Teiserver acts on the number of hashes, so I don't think this will be a breaking change. If desired, I can return a fixed value.

test/teiserver/game/balancer_server_test.exs
```diff
@@ -1,9 +1,56 @@
...
+          "email" => to_string(ExULID.ULID.generate()) <> "@example.com",
+          "permissions" => []
+        }
+      end)
+      |> Enum.map(fn x -> make_user(x) end)
+      # unwrap user creation response
+      |> Enum.map(fn {:ok, reply} -> reply end)
+      |> Enum.map(fn x -> Map.put(x, :userid, x.id) end)
+      # each user in their own party
+      |> Enum.map(fn x -> Map.put(x, :party_id, x.id + 1) end)
```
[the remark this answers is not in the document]
NortySpock: rewrote it to accept a list of ranks and a list of party integers

test/teiserver/game/balancer_server_test.exs
```diff
@@ -1,9 +1,56 @@
...
+
+    # set the user rank
+    Enum.map(Enum.with_index(users), fn {user, index} ->
+      Teiserver.Account.update_user_stat(user.id, %{
+        :rank => Enum.at(list_of_ranks_to_assign_to_users, index)
+      })
+    end)
```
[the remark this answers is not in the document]
NortySpock: Thanks, used this in at least one case.

test/teiserver/game/balancer_server_test.exs
```diff
@@ -36,7 +83,147 @@ defmodule Teiserver.Game.BalancerServerTest do
...
+
+    players2 =
+      players
+      |> Enum.map(fn x -> Map.put(x, :userid, x.id) end)
+
+    dbg(players2)
+
+    {:ok, start_link_results} = BalancerServer.start_link(data: %{lobby_id: 1})
+
+    starting_state = GenServer.call(start_link_results, :report_state, 5000)
```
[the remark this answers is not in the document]
NortySpock: Thanks, cleaned up

test/teiserver/game/balancer_server_test.exs
```diff
@@ -35,8 +40,176 @@ defmodule Teiserver.Game.BalancerServerTest do
...
+        %{
+          "name" => "user" <> to_string(index + 1) <> "_" <> to_string(ExULID.ULID.generate()),
+          "email" => to_string(ExULID.ULID.generate()) <> "@example.com",
+          "permissions" => []
+        }
+      end)
+      |> Enum.map(fn x -> make_user(x) end)
+      # unwrap user creation response
+      |> Enum.map(fn {:ok, reply} -> reply end)
+      |> Enum.map(fn x -> Map.put(x, :userid, x.id) end)
```
geekingfrog, line 89: I don't think you care about the username at all, so you shouldn't need the index to create the usernames and emails, the random ones from the fixture are enough.

Also, you don't need to chain multiple maps like that, you can have one single function, and also combine it with the loop just below:
```elixir
Enum.map(list_of_ranks, fn rank ->
  {:ok, user} = make_user()
  update_user_stat(user.id, %{:rank => rank})
end)
```

Also, why do you need to set `:userid` on the created user?

test/teiserver/game/balancer_server_test.exs
```diff
@@ -35,8 +40,176 @@ defmodule Teiserver.Game.BalancerServerTest do
...
+      Teiserver.Account.update_user_stat(user.id, %{
+        :rank => rank
+      })
+    end
+
+    # set the user party_id
+    users =
+      for {user, party_id} <- Enum.zip(users, list_of_parties) do
+        Map.put(user, :party_id, party_id)
+      end
```
geekingfrog, line 102: It seems you are always passing different party ids to this function, and they are all basically indexes.
So you could drop the argument and this loop, and set it when creating the user with the ranks. The entire function can become something like:

```elixir
for {rank, i} <- Enum.with_index(list_of_ranks) do
  {:ok, user} = create_user()
  user = Map.put(user, :party_id, i)
  update_user_stat(...)
  user
end
```

test/teiserver/game/balancer_server_test.exs
```diff
@@ -35,8 +40,176 @@ defmodule Teiserver.Game.BalancerServerTest do
...
+
+    users
+  end
+
+  test "load some players in and run a balance pass or two" do
+    team_count = 2
+
+    players =
+      make_users_with_ranks_and_parties([10, 20, 30, 40, 50, 60], [1, 2, 3, 4, 5, 6])
+      |> Enum.map(fn x -> Map.put(x, :userid, x.id) end)
```
geekingfrog, line 112: again, I'm not sure why you need that, and besides, the helper function already does that

test/teiserver/game/balancer_server_test.exs
```diff
@@ -35,8 +40,176 @@ defmodule Teiserver.Game.BalancerServerTest do
...
+        {:make_balance, team_count, [], players}
+      )
+
+    second_balance_pass_hash_reuse =
+      GenServer.call(
+        pid,
+        {:make_balance, team_count, [], players}
+      )
+
+    assert second_balance_pass_hash_reuse.hash == first_balance_pass.hash
```
geekingfrog, line 128: you should probably also test that the results are the same

test/teiserver/game/balancer_server_test.exs
```diff
@@ -35,8 +40,176 @@ defmodule Teiserver.Game.BalancerServerTest do
...
+    GenServer.cast(
+      pid,
+      :reset_hashes
+    )
+
+    poll_until_nil(fn -> GenServer.call(pid, :get_current_balance) end)
+
+    with {:ok, hash} <- Map.fetch(GenServer.call(pid, :report_state), :last_balance_hash) do
+      assert hash == nil
+    end
```
geekingfrog, line 183: you can do `assert Map.get(GenServer.call(pid, :report_state), :last_balance_hash, :not_found) == nil`

geekingfrog (changes requested):

test/teiserver/game/balancer_server_test.exs
NortySpock: Ah, yes, that is cleaner, thank you.

test/teiserver/game/balancer_server_test.exs
NortySpock: Ah, yes, good point. Done.

lib/teiserver/game/servers/balancer_server.ex
```diff
@@ -35,35 +34,68 @@ defmodule Teiserver.Game.BalancerServer do
...
+        [
+          algorithm: state.algorithm,
+          max_deviation: state.max_deviation,
+          rating_lower_boundary: state.rating_lower_boundary,
+          rating_upper_boundary: state.rating_upper_boundary,
+          mean_diff_max: state.mean_diff_max,
+          stddev_diff_max: state.stddev_diff_max,
+          fuzz_multiplier: state.fuzz_multiplier,
+          shuffle_first_pick: state.shuffle_first_pick
+        ]
```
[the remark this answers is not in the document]
NortySpock: Will do, was on the fence on if it was worth DRY-ing it up for a single duplication.

test/teiserver/game/balancer_server_test.exs
NortySpock: True, I guess it could be condensed on the assumption that everyone will be in different parties. I was trying to be explicit-rather-than-clever when trying to write testing code. I also wanted something that was easy to extend if we wanted to test cases with parties ("Oh, let's test this with Alice and Bob the same party, Charlie and Denise in different parties") which could be done with 

`make_users_with_ranks_and_parties([10, 20, 30, 40,], [1, 1, 2, 3])`

Is that valuable at all? 

Happy to throw it out in favor of using `Central.Helpers.GeneralTestLib.make_user()`; just thought I'd offer my reasoning for why I wrote the test helper that way.

test/teiserver/game/balancer_server_test.exs
NortySpock: required because BalancerServer expects `:userid` and `:party_id`

test/teiserver/game/balancer_server_test.exs
NortySpock: (resolved in the direction of the requested changes)

test/teiserver/game/balancer_server_test.exs
```diff
@@ -50,66 +50,31 @@ defmodule Teiserver.Game.BalancerServerTest do
...
-
-    # set the user party_id
+  def make_users_with_ranks(list_of_ranks) do
     users =
-      for {user, party_id} <- Enum.zip(users, list_of_parties) do
-        Map.put(user, :party_id, party_id)
+      for {rank, i} <- Enum.with_index(list_of_ranks) do
+        user = Central.Helpers.GeneralTestLib.make_user(%{})
+        user = Map.put(user, :party_id, i)
+        user = Map.put(user, :userid, user.id)
```
geekingfrog, line 61: bit of a nitpick, but you can use the pipe operator there:

```elixir
user = Central.Helpers.GeneralTestLib.make_user(%{})
  |> Map.put(:party_id, i)
  |> Map.put(:userid, user.id)
```

geekingfrog (approved):

geekingfrog:
I'll squash merge that tomorrow if you fancy the nitpick I mentioned in the meantime, not required though.

NortySpock:
> I'll squash merge that tomorrow if you fancy the nitpick I mentioned in the meantime, not required though.

I'll try to get the nitpick fix in tonight.

Thanks for taking the time to guide me through the process, learned a lot about Elixir syntax  along the way. :)

geekingfrog:
Good job with that, that was a really nasty issue

PR #597, opened 2025-02-15
Discord lib upgrade

geekingfrog:
This should be tested for safety, although that should work. I got the startup message for my local bot:
2025-02-15-382x100-scrot

This is in preparation for an elixir upgrade, the newer version addressed the warnings when bumping elixir version.


config/runtime.exs
```diff
@@ -162,3 +137,30 @@ config :logger, :error_log, path: "#{log_root_path}/error.log"
 config :logger, :notice_log, path: "#{log_root_path}/notice.log"
 
 config :logger, :info_log, path: "#{log_root_path}/info.log"
+
+enable_discord_bridge =
+  Teiserver.ConfigHelpers.get_env("TEI_ENABLE_DISCORD_BRIDGE", false, :bool)
```
geekingfrog, line 142: I changed the default from `true` to `false` so that it works out of the box locally. This should not impact anything on the real servers because [the env var](https://github.com/beyond-all-reason/ansible-teiserver/blob/b42a42da2726ba0ac9f64d3850ceb7df972a1657/roles/teiserver/templates/teiserver.env.j2#L40-L48) is always set correctly.

L-e-x-o-n:
I tried updating Nostrum in the past but ran into issue. It was able to send messages (discord bridge, server startup, moderation logs and similar) but not read them and as a result not respond to $text commands. I went through the documentation and code several times but couldn't find a solution so it didn't result in anything at the time. I am interesting if you will encounter the same issues.

geekingfrog:
I am encountering the same issues yes, and looking at how to update the code so that it still works.
The joy of libs without changelog or upgrade path :(

geekingfrog:
Alright, now with some additional changes, the text callbacks seems to work, and the `$gdt` command as well, but since I cannot have forum channels on my test discord server, I couldn't fully check. At least it is not raising any errors.

2025-02-17-2276x158-scrot
2025-02-17-1591x383-scrot

L-e-x-o-n:
Tested on integration, everything works: automated messages, $text and / commands and bridging. This is great!

PR #601, opened 2025-02-19
User map pool, games and engines in matchmaking

### LINKED ISSUE #591: [Feature]: Use map pool in matchmaking
### Describe the Feature

Now that teiserver knows about maps (https://github.com/beyond-all-reason/teiserver/pull/569) and their associated matchmaking queues, we can integrate this info to the matchmaking system.

* grab the relevant maps when a queue process starts
* return this list of map in the response to `matchmaking/list`: https://github.com/beyond-all-reason/tachyon/blob/master/docs/schema/matchmaking.md#list
* if the map list is empty, joining a queue should result in an `internal_error`
* pass the map info along to the pairing room process
* when a match is confirmed, the pairing room process should pick a map at random and create the start script data with that

This will require changes to the existing matchmaking tests: need to setup at least one map for all the existing tests, then add more tests:
* with multiple maps
* with no map (check that we get an `internal_error`)

L-e-x-o-n:
Resolves #591 
- [x] Grabs the  relevant maps when a queue process starts
- [x] Return this list of map in the response to `matchmaking/list`
- [x] If the map list is empty, joining a queue should result in an `internal_error`
- [x] Pass the map info along to the pairing room process
- [x] When a match is confirmed, the pairing room process picks a map at random and create the start script data
Additionally added games and engines to the queue process and their selection to pairing room, passing them to start_script and showing them in `matchmaking/list`

### DIFF
--- a/lib/teiserver/matchmaking/pairing_room.ex
+++ b/lib/teiserver/matchmaking/pairing_room.ex
@@ -117,40 +117,54 @@ defmodule Teiserver.Matchmaking.PairingRoom do
         {:stop, :normal, state}
 
       id ->
-        start_script = hardcoded_start_script(state)
+        engine = select_engine(state.queue.engines)
+        %{spring_game: game} = select_game(state.queue.games)
+        %{spring_name: map} = select_map(state.queue.maps)
 
-        case Teiserver.TachyonBattle.start_battle(id, start_script) do
-          {:error, reason} ->
-            QueueServer.disband_pairing(state.queue_id, self())
+        start_battle(state, id, engine, game, map)
+    end
+  end
 
-            for team <- state.teams, member <- team, p_id <- member.player_ids do
-              Teiserver.Player.matchmaking_notify_lost(p_id, {:server_error, reason})
-            end
+  defp start_battle(state, host_id, engine, game, map) do
+    start_script = start_script(state, engine, game, map)
 
-            {:stop, :normal, state}
+    case Teiserver.TachyonBattle.start_battle(host_id, start_script) do
+      {:error, reason} ->
+        QueueServer.disband_pairing(state.queue_id, self())
 
-          {:ok, host_data} ->
-            QueueServer.disband_pairing(state.queue_id, self())
+        state.teams
+        |> List.flatten()
+        |> Enum.flat_map(& &1.player_ids)
+        |> Enum.each(fn p_id ->
+          Teiserver.Player.matchmaking_notify_lost(p_id, {:server_error, reason})
+        end)
 
-            ids =
-              for team <- state.teams, member <- team, p_id <- member.player_ids do
-                p_id
-              end
+        {:stop, :normal, state}
 
-            Logger.debug("Pairing completed for players " <> Enum.join(ids, ","))
+      {:ok, host_data} ->
+        QueueServer.disband_pairing(state.queue_id, self())
 
-            battle_start_data =
-              host_data
-              |> Map.put(:engine, %{version: start_script.engineVersion})
-              |> Map.put(:game, %{springName: start_script.gameName})
-              |> Map.put(:map, %{springName: start_script.mapName})
+        ids =
+          for team <- state.teams, member <- team, p_id <- member.player_ids do
+            p_id
+          end
 
-            for team <- state.teams, member <- team, p_id <- member.player_ids do
-              Teiserver.Player.battle_start(p_id, battle_start_data)
-            end
+        Logger.debug("Pairing completed for players " <> Enum.join(ids, ","))
 
-            {:stop, :normal, state}
-        end
+        battle_start_data =
+          host_data
+          |> Map.put(:engine, engine)
+          |> Map.put(:game, %{springName: game})
+          |> Map.put(:map, %{springName: map})
+
+        state.teams
+        |> List.flatten()
+        |> Enum.flat_map(& &1.player_ids)
+        |> Enum.each(fn p_id ->
+          Teiserver.Player.battle_start(p_id, battle_start_data)
+        end)
+
+        {:stop, :normal, state}
     end
   end
 
@@ -217,12 +231,13 @@ defmodule Teiserver.Matchmaking.PairingRoom do
     {:stop, :normal, state}
   end
 
-  @spec hardcoded_start_script(state()) :: Teiserver.TachyonBattle.start_script()
-  defp hardcoded_start_script(state) do
+  @spec start_script(state(), %{version: String.t()}, String.t(), String.t()) ::
+          Teiserver.TachyonBattle.start_script()
+  defp start_script(state, engine, game, map) do
     %{
-      engineVersion: "105.1.1-2590-gb9462a0 bar",
-      gameName: "Beyond All Reason test-26929-d709d32",
-      mapName: "Red Comet Remake 1.8",
+      engineVersion: engine.version,
+      gameName: game,
+      mapName: map,
       startPosType: :ingame,
       allyTeams: get_ally_teams(state)
     }
@@ -239,5 +254,21 @@ defmodule Teiserver.Matchmaking.PairingRoom do
 
       %{teams: teams}
     end
+  end
+
+  # TODO implement some smarter engine/game selection logic here in the future, get first for now
+  @spec select_engine([%{version: String.t()}]) :: %{version: String.t()}
+  def select_engine(engines) do
+    Enum.at(engines, 0)
+  end
+
+  @spec select_game([%{spring_game: String.t()}]) :: %{spring_game: String.t()}
+  def select_game(games) do
+    Enum.at(games, 0)
+  end
+
+  @spec select_map([Teiserver.Asset.Map.t()]) :: Teiserver.Asset.Map.t()
+  def select_map(maps) do
+    Enum.random(maps)
   end
 end
--- a/lib/teiserver/matchmaking/pairing_room.ex
+++ b/lib/teiserver/matchmaking/pairing_room.ex
@@ -132,12 +132,9 @@ defmodule Teiserver.Matchmaking.PairingRoom do
       {:error, reason} ->
         QueueServer.disband_pairing(state.queue_id, self())
 
-        state.teams
-        |> List.flatten()
-        |> Enum.flat_map(& &1.player_ids)
-        |> Enum.each(fn p_id ->
+        for team <- state.teams, member <- team, p_id <- member.player_ids do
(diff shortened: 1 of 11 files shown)


test/teiserver_web/tachyon/matchmaking_test.exs
```diff
@@ -3,52 +3,111 @@ defmodule Teiserver.Tachyon.MatchmakingTest do
...
+    queue_attrs(id, team_size)
+    |> put_in([:settings, :pairing_timeout], timeout)
+    |> mk_queue()
+  end
+
+  def setup_mapless_queue(id, team_size) do
+    queue_attrs(id, team_size)
+    |> Map.put(:maps, [])
+    |> mk_queue()
+  end
```
geekingfrog, line 84: I am not the biggest fan of these very simple function, especially since they are used only once. I think it's clearer in the tests to construct the right attrs directly, basically, inline these functions.

test/teiserver_web/tachyon/matchmaking_test.exs
```diff
@@ -3,52 +3,111 @@ defmodule Teiserver.Tachyon.MatchmakingTest do
...
 
-    test "works", %{client: client} do
-      resp = Tachyon.list_queues!(client)
+  def altair_attr(id),
+    do: %{
+      spring_name: "Altair Crossing Remake 1.2.3",
+      display_name: "Altair Crossing",
+      thumbnail_url: "https://www.beyondallreason.info/map/altair-crossing",
+      matchmaking_queues: ["1v1", "2v2", "3v3", id]
```
geekingfrog, line 18: why do you need some defaults for the `matchmaking_queues`?

test/teiserver_web/tachyon/matchmaking_test.exs
```diff
@@ -59,8 +118,64 @@ defmodule Teiserver.Tachyon.MatchmakingTest do
...
+            "teamSize" => 2,
+            "ranked" => true,
+            "engines" => engine_names(),
+            "games" => game_names(),
+            "maps" => map_names(q2v2_id)
+          }
+        ])
+
+      # Checking for subset because the response also contains default queues
+      assert MapSet.subset?(expected_playlists, MapSet.new(resp["data"]["playlists"]))
```
geekingfrog, line 155: Alternatively you can use [Enum.member?/2](https://hexdocs.pm/elixir/Enum.html#member?/2) for each map you're expecting.

test/teiserver_web/tachyon/matchmaking_test.exs
```diff
@@ -110,13 +225,73 @@ defmodule Teiserver.Tachyon.MatchmakingTest do
...
+        Teiserver.Matchmaking.QueueServer.init_state(%{
+          id: id,
+          name: id,
+          team_size: 1,
+          team_count: 2,
+          engines: [],
+          games: game_versions(),
+          maps: map_names(id)
+        })
+        |> Teiserver.Matchmaking.QueueServer.start_link()
```
geekingfrog, line 252: Why not use `QueueSupervisor.start_queue!(state)`? This way you guarantee that this setup is faithful to the real thing.

geekingfrog (commented):
I think the failing tests are because the async tests don't work well with the sql sandbox. And now that starting a queue requires DB access, the tests fail because of the sandbox.

lib/teiserver/matchmaking/pairing_room.ex
```diff
@@ -117,40 +117,54 @@ defmodule Teiserver.Matchmaking.PairingRoom do
...
+        QueueServer.disband_pairing(state.queue_id, self())
 
-          {:ok, host_data} ->
-            QueueServer.disband_pairing(state.queue_id, self())
+        state.teams
+        |> List.flatten()
+        |> Enum.flat_map(& &1.player_ids)
+        |> Enum.each(fn p_id ->
+          Teiserver.Player.matchmaking_notify_lost(p_id, {:server_error, reason})
+        end)
```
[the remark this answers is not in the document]
L-e-x-o-n: I had some warnings that `for` did not return anything, but I can't reproduce them anymore. I think the `for`s are a bit cleaner and easier to understand here so will revert.

lib/teiserver/matchmaking/pairing_room.ex
```diff
@@ -117,40 +117,54 @@ defmodule Teiserver.Matchmaking.PairingRoom do
...
+          |> Map.put(:map, %{springName: map})
+
+        state.teams
+        |> List.flatten()
+        |> Enum.flat_map(& &1.player_ids)
+        |> Enum.each(fn p_id ->
+          Teiserver.Player.battle_start(p_id, battle_start_data)
+        end)
+
+        {:stop, :normal, state}
```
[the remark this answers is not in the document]
L-e-x-o-n: I will try, unfortunately I was jumping all over the place and changing things multiple times in different order.

lib/teiserver/matchmaking/queue_supervisor.ex
```diff
@@ -6,6 +6,30 @@ defmodule Teiserver.Matchmaking.QueueSupervisor do
...
+              team_count: any(),
+              team_size: any()
+            },
+            settings: %{
+              :max_distance => any(),
+              :pairing_timeout => any(),
+              :tick_interval_ms => any(),
+              optional(any()) => any()
+            }
+          },
```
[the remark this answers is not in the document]
L-e-x-o-n: Oops, I think I clicked the spec generator here which did this.

test/teiserver_web/tachyon/matchmaking_test.exs
L-e-x-o-n: I don't, but thought there was no harm in putting some as an example.

test/teiserver_web/tachyon/matchmaking_test.exs
L-e-x-o-n: Copied from the "too many players" test above, will change.

lib/teiserver/matchmaking/pairing_room.ex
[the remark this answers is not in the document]
geekingfrog: Yeah, no worry, it's definitely a habit that takes a bit of practice, and when you don't know where you're going it's extra hard. Just something to aim for to improve bit by bit.

lib/teiserver/matchmaking/queue_supervisor.ex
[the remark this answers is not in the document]
geekingfrog: There's a spec generator??? How does that work?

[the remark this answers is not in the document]
L-e-x-o-n: With the ElixirLS extension it will shows it like this:
image
You can click on it and it will generate it like you saw in this example above, it can be useful sometimes but usually generates a lot of `ąny`s or other more generalised types so I don't really use it.

test/teiserver_web/tachyon/matchmaking_test.exs
```diff
@@ -3,52 +3,111 @@ defmodule Teiserver.Tachyon.MatchmakingTest do
...
+  end
+
+  def setup_mapless_queue(id, team_size) do
+    queue_attrs(id, team_size)
+    |> Map.put(:maps, [])
+    |> mk_queue()
+  end
+
+  defp setup_maps(id) do
+    Asset.delete_all_maps()
```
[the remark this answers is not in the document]
L-e-x-o-n: Without it I was running into unique constraint error, but after thinking about it now that's probably because some tests setup multiple queues which then try adding the same map multiple times.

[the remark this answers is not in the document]
L-e-x-o-n: Reworked this a bit

test/teiserver_web/tachyon/matchmaking_test.exs
L-e-x-o-n: Moved to tests where possible

test/teiserver_web/tachyon/matchmaking_test.exs
L-e-x-o-n: Removed to not cause confusion

test/teiserver_web/tachyon/matchmaking_test.exs
L-e-x-o-n: Changed.

test/support/fixtures/asset.ex
```diff
@@ -1,8 +1,28 @@
...
+
+  def create_or_update_map(attrs) do
+    case Asset.get_map(attrs.spring_name) do
+      nil -> create_map(attrs)
+      map -> update_map(map, attrs)
+    end
+  end
+
+  def update_map(map, attrs) do
+    Logger.warning(
```
geekingfrog, line 18: that may be a bit noisy, especially since it's just for testing.

geekingfrog (approved):

geekingfrog:
Just need to add a commit to fix the matchmaking tests, remove the async tag, and make sure there are maps, engines and games setup for the test to still pass.

test/support/fixtures/asset.ex
```diff
@@ -1,8 +1,28 @@
...
 
   def create_map(attrs) do
     %Asset.Map{} |> Asset.Map.changeset(attrs) |> Repo.insert!()
   end
+
+  def create_or_update_map(attrs) do
+    case Asset.get_map(attrs.spring_name) do
+      nil -> create_map(attrs)
+      map -> update_map(map, attrs)
+    end
```
geekingfrog, line 14: Thinking a bit more about that, I think that can work, but may create some interferences with other tests. For example, the ones like `teiserver/matchmaking/queue_test.exs`.
Another approach could be to have the name be randomly generated. So that with random queue_name, there's basically no chance of conflicts. Your call

L-e-x-o-n: I was actually using randomised names but at the time it was a mess so I reverted the change and had a single map for a while.
Changed to use random map names again, simpler than updating existing ones and probably with less potential issues.

test/teiserver/matchmaking/queue_test.exs
```diff
@@ -2,13 +2,31 @@ defmodule Teiserver.Matchmaking.QueueTest do
   use Teiserver.DataCase
   alias Teiserver.Matchmaking
   alias Teiserver.Matchmaking.QueueServer
+  alias Teiserver.AssetFixtures
+
+  defp stg_attr(id),
+    do: %{
+      spring_name: "Supreme that glitters",
```
geekingfrog, line 9: 🤣

geekingfrog (approved):

PR #605, opened 2025-02-21
Reduce s.d. dev diff penalty to split_noobs

jauggy:
Follow up from here: https://discord.com/channels/549281623154229250/1338898797387972714/1342615307868700734
BLodir's party was split because of very high sd diff. However, he was not in an op party and had the party been kept the ratings would have been closer. If we reduce the sd diff importance alone, this would cause one of our tests to fail (line 105 in split_noobs_test). There is a test there to ensure top three players don't end up on same team.

In order to keep Blodir's party but also split op parties the following changes are made:

- party importance reduced to 6 from 7
- s.d. importance reduced to 1
- captain importance set to 1


L-e-x-o-n:
What about `rating_diff_importance`, would increasing it as well be beneficial?

jauggy:
Increasing rating diff importance is identical to reducing party diff importance and reducing s.d. diff importance. I did test that. But one of the tests broke a party when it didn't need to if I do just that.

jauggy:
OK tried some other tweaks. These settings also work to pass tests:

- party importance reduced to 6 (from 7)
- s.d. importance reduced to 1 (from 4)
- captain importance set to 1 (from 0)

PR #608, opened 2025-02-25
Make tau configurable by admin

jauggy:
Tau is configurable by admin in Site Config -> Lobby section

The default value of tau is unchanged (which is 1/3)

Also adding in prevent_sigma_increase to true

### DIFF
--- a/lib/teiserver/config.ex
+++ b/lib/teiserver/config.ex
@@ -458,6 +458,7 @@ defmodule Teiserver.Config do
 
     case type.type do
       "integer" -> Teiserver.Helper.NumberHelper.int_parse(value)
+      "float" -> Teiserver.Helper.NumberHelper.float_parse(value)
       "boolean" -> if value == "true" or value == true, do: true, else: false
       "select" -> value
       "string" -> value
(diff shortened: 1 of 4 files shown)


L-e-x-o-n:
Can you please put it under a new Openskill section?

jauggy:
I've added it to a Rating section. Calling it more generic just in case we change from openskill to another system.


Also I've added in prevent_sigma_increase to true so that we don't have stuff like this where uncertainty increases:

image

This occurs when tau is very high. The way it works is that before a match is rated, the value of tau is added to uncertainty then the match is rated, so if tau is really high you can gain uncertainty.
```
    new_rating_groups =
      Enum.map(rating_groups, fn team ->
        Enum.map(team, fn {mu, sigma} ->
          {mu, Math.sqrt(sigma ** 2 + options.tau ** 2)}
        end)
      end)

    output = options.model.rate(new_rating_groups, options)
    ```
    
    So your uncertainty becomes `Math.sqrt(sigma ** 2 + options.tau ** 2)`
    Then the match gets rated, and there's a chance uncertainty rises if it doesn't fall enough after rating.
    
    The prevent_sigma_increase is already in the openskill lib.

PR #612, opened 2025-02-28
Elixir upgrade 1.18

geekingfrog:


lib/teiserver/account/reports/retention_report.ex
```diff
@@ -70,7 +70,7 @@ defmodule Teiserver.Account.RetentionReport do
             end
           end)
 
-        if last_login != nil and last_played != nil do
```
L-e-x-o-n, line 73: Why this?

geekingfrog: I’m following the compiler. And `Timex.from_unix` raises an error if it fails, it never returns nil. The newer compiler is now able to spot that kind of things.

L-e-x-o-n (approved):

Issue #619, opened 2025-03-13
[Feature]: implement tachyon messages for friends


geekingfrog:
### Describe the Feature

spec there: https://github.com/beyond-all-reason/tachyon/blob/master/docs/schema/friend.md

Issue #573, opened 2025-02-02
[Feature]: Limit relation lists size


geekingfrog:
### Describe the Feature

It is unreasonable to allow unlimited number of friends, blocked, ignored and avoided players.
We should limit all of these list to some arbitrary number, let's say, 300.

For people already over the limit, I suggest we remove items based on their creation date. So oldest additions to the list get removed until it's at 300.

### Expected Behavior

When trying to add a new friend/block/ignore/avoid should get an error.

There should be a mix task to trim down all existing relationship lists to their new maximum.

L-e-x-o-n:
I agree, a limit makes sense, but is it necessary to delete existing ones to get them below 300?
Should probably run some queries and check how many users actually have over 300 relationships.

PR #620, opened 2025-03-13
Tachyon friends

geekingfrog:
This should cover the messages to [handle friends](https://github.com/beyond-all-reason/teiserver/issues/619).

I noticed it's getting a bit verbose sending error responses, I'll refactor that and put it as part of the [handler](https://github.com/beyond-all-reason/teiserver/blob/master/lib/teiserver/tachyon/handler.ex) behaviour, but that can be done in another PR.

### DIFF
--- a/lib/teiserver/account/libs/friend_request_lib.ex
+++ b/lib/teiserver/account/libs/friend_request_lib.ex
@@ -144,33 +144,30 @@ defmodule Teiserver.Account.FriendRequestLib do
     :ok
   end
 
+  @doc """
+  Returns all friend requests for the given user
+  """
+  @spec list_requests_for_user(T.userid()) :: {outgoing :: [map()], incoming :: [map()]}
+  def list_requests_for_user(userid) do
+    Account.list_friend_requests(where: [to_or_from_is: userid])
+    |> Enum.reduce({[], []}, fn req, {outgoing, incoming} ->
+      if req.from_user_id == userid do
+        {[req | outgoing], incoming}
+      else
+        {outgoing, [req | incoming]}
+      end
+    end)
+  end
+
   @spec list_incoming_friend_requests_of_userid(T.userid()) :: [T.userid()]
   def list_incoming_friend_requests_of_userid(userid) do
-    Teiserver.cache_get_or_store(:account_incoming_friend_request_cache, userid, fn ->
-      Account.list_friend_requests(
-        where: [
-          to_user_id: userid
-        ],
-        select: [:from_user_id]
-      )
-      |> Enum.map(fn %{from_user_id: from_user_id} ->
-        from_user_id
-      end)
-    end)
+    {_, incoming} = list_requests_for_user(userid)
+    Enum.map(incoming, fn incoming -> incoming.from_user_id end)
   end
 
   @spec list_outgoing_friend_requests_of_userid(T.userid()) :: [T.userid()]
   def list_outgoing_friend_requests_of_userid(userid) do
-    Teiserver.cache_get_or_store(:account_outgoing_friend_request_cache, userid, fn ->
-      Account.list_friend_requests(
-        where: [
-          from_user_id: userid
-        ],
-        select: [:to_user_id]
-      )
-      |> Enum.map(fn %{to_user_id: to_user_id} ->
-        to_user_id
-      end)
-    end)
+    {outgoing, _} = list_requests_for_user(userid)
+    Enum.map(outgoing, fn outgoing -> outgoing.to_user_id end)
   end
 end
--- a/lib/teiserver/player/tachyon_handler.ex
+++ b/lib/teiserver/player/tachyon_handler.ex
@@ -400,6 +400,36 @@ defmodule Teiserver.Player.TachyonHandler do
     end
   end
 
+  def handle_command("friend/rejectRequest" = cmd_id, "request", message_id, msg, state) do
+    with {:ok, originator_id} <- parse_user_id(msg["data"]["from"]),
+         :ok <- Account.decline_friend_request(originator_id, state.user.id) do
+      {:response, cmd_id, nil, state}
+    else
+      {:error, :invalid_id} ->
+        resp =
+          Schema.error_response(cmd_id, message_id, :invalid_user)
+          |> Jason.encode!()
+
+        {:reply, :ok, {:text, resp}, state}
+
+      {:error, "no request"} ->
+        resp =
+          Schema.error_response(cmd_id, message_id, :no_pending_request)
+          |> Jason.encode!()
+
+        {:reply, :ok, {:text, resp}, state}
+
+      err ->
+        Logger.error("Unhandled error in rejectRequest: #{inspect(err)}")
+
+        resp =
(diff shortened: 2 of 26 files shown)


lib/teiserver/player/tachyon_handler.ex
```diff
@@ -301,6 +321,172 @@ defmodule Teiserver.Player.TachyonHandler do
...
+      outgoingPendingRequests: outgoing
+    }
+
+    {:response, cmd_id, resp, state}
+  end
+
+  def handle_command("friend/sendRequest" = cmd_id, "request", message_id, msg, state) do
+    with {:ok, target} <- get_user(msg["data"]["to"]),
+         {:ok, %Account.FriendRequest{}} <-
+           Account.create_friend_request(state.user.id, target.id) do
```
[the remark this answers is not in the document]
L-e-x-o-n: Unrelated to your code but relevant to the feature... I think I discovered a bug, we are not checking if a friend request exists where the other person is the sender. It is possible for 2 persons to send each other a friend requests at the same time. This will create 2 different requests in the db and is the explanation for some errors in the log that pop up occasionally . 
The way to solve this is to check if the other person sent a request first and if so accept it instead of sending a new one.

lib/teiserver/player/tachyon_handler.ex
```diff
@@ -301,6 +321,172 @@ defmodule Teiserver.Player.TachyonHandler do
...
+
+        {:reply, :ok, {:text, resp}, state}
+    end
+  end
+
+  def handle_command("friend/rejectRequest" = cmd_id, "request", message_id, msg, state) do
+    with {:ok, originator_id} <- parse_user_id(msg["data"]["from"]),
+         %Account.FriendRequest{} = req <-
+           Account.get_friend_request(originator_id, state.user.id),
+         {:ok, _changeset} <- Account.delete_friend_request(req) do
```
L-e-x-o-n, line 415: Why not use `Account.decline_friend_request`?

lib/teiserver/player/tachyon_handler.ex
```diff
@@ -301,6 +321,172 @@ defmodule Teiserver.Player.TachyonHandler do
...
+          Schema.error_response(cmd_id, message_id, :internal_error, inspect(err))
+          |> Jason.encode!()
+
+        {:reply, :ok, {:text, resp}, state}
+    end
+  end
+
+  def handle_command("friend/cancelRequest" = cmd_id, "request", message_id, msg, state) do
+    with {:ok, target_id} <- parse_user_id(msg["data"]["to"]),
+         %Account.FriendRequest{} = req <- Account.get_friend_request(state.user.id, target_id),
```
L-e-x-o-n, line 446: Why not use `Account.rescind_friend_request`?

lib/teiserver/player/tachyon_handler.ex
```diff
@@ -301,6 +321,172 @@ defmodule Teiserver.Player.TachyonHandler do
...
+         {:ok, _changeset} <- Account.delete_friend(friend) do
+      Player.Session.friend_removed(target_id, state.user.id)
+      {:response, cmd_id, nil, state}
+    else
+      nil ->
+        resp =
+          Schema.error_response(cmd_id, message_id, :not_in_friendlist)
+          |> Jason.encode!()
+
+        {:reply, :ok, {:text, resp}, state}
```
L-e-x-o-n, line 472: I know the schema for this is already done but it does remind me of our discussion on party spec::
> party/leave should probably have not_in_party fail reason.
>> I don't think it makes sense. What would a client do with that information? most likely, it'll ignore it since the goal is achieved. Making it a no-op also makes it idempotent which is always a nice property to have.

I think the same applies here, we even have `Account.delete_friend/2` which returns :ok if the users are not friends.

lib/teiserver/account/libs/friend_request_lib.ex
```diff
@@ -173,4 +186,17 @@ defmodule Teiserver.Account.FriendRequestLib do
...
+  and returns the entire records
+  """
+  @spec list_outgoing_friend_requests(T.userid()) :: [map()]
+  def list_outgoing_friend_requests(userid) do
+    Account.list_friend_requests(
+      where: [
+        from_user_id: userid
+      ]
+    )
+  end
```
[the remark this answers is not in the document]
geekingfrog: I reworked the logic in fed8619c4271504052061155dd05c869a96e5b91 and the next commit.
Dropped the cache altogether because it shouldn't be something high traffic that requires caching and it'll simplify the code.

I am not sure about what you mean about the "mapping inside the lib rather than the handler". The mapping in the handler is to convert internal structures to the tachyon message, and that definitely belongs to the handler.

lib/teiserver/player/tachyon_handler.ex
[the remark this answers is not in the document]
geekingfrog: I have created https://github.com/beyond-all-reason/teiserver/issues/622 to fix that.
I'd rather do it separately, this is focused more on the tachyon messages and less about the underlying features.

lib/teiserver/player/tachyon_handler.ex
```diff
@@ -301,6 +321,172 @@ defmodule Teiserver.Player.TachyonHandler do
...
+          Schema.error_response(cmd_id, message_id, :internal_error)
+          |> Jason.encode!()
+
+        {:reply, :ok, {:text, resp}, state}
+    end
+  end
+
+  def handle_command("friend/acceptRequest" = cmd_id, "request", message_id, msg, state) do
+    with {:ok, originator_id} <- parse_user_id(msg["data"]["from"]),
+         :ok <- Account.accept_friend_request(originator_id, state.user.id) do
```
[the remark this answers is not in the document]
geekingfrog: Let's do that in https://github.com/beyond-all-reason/teiserver/issues/573, also need amend the schema to add this error reason.

lib/teiserver/player/tachyon_handler.ex
geekingfrog: didn't see it, amended the commit to use that function.

lib/teiserver/player/tachyon_handler.ex
geekingfrog: same, didn't see it, amended the commit to use that function.

lib/teiserver/player/tachyon_handler.ex
geekingfrog: amended the commit to make it a no-op when there is no friend request.

lib/teiserver/account/libs/friend_request_lib.ex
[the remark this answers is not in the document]
L-e-x-o-n: I was thinking of this part:
```ex
Account.list_friends_for_user(state.user.id)
      |> Enum.map(fn friend ->
        id = if friend.user1_id == state.user.id, do: friend.user2_id, else: friend.user1_id
```
it seems like finding which user is the friend will be common logic that would have to be used whenever you use `Account.list_friends_for_user` and so I was thinking of changing the list_friends_for_user to return a list a list of maps or similar containing friend and inserted_at.
I don't have any strong opinions on it, just an idea.

PR #642, opened 2025-04-06
Add total matches col

jauggy:
# Purpose
Currently num_matches and num_wins represent the number of matches/wins for the season. We need similar data for total (all seasons) for the purpose of potentially allowing provisional ratings in the future and for doing analysis. 

If we want to calculate this to add data to the match rating logs, we would have to do a group by and sum and it requires an additional query. The number of rows in the rating table will increase each season too making this additional query more costly as time goes by.

Adding a total_matches and total_wins column means we don't have to do an additional query.

# Changes
Added migration to add total_matches and total_wins column
Added migration to populate these columns for season 1 and season 2.


priv/repo/migrations/20250406001309_populate_total_matches.exs
```diff
@@ -0,0 +1,28 @@
...
+    total_wins = temp_table.total_wins
+    from (
+    select user_id, rating_type_id, sum(num_matches) as total_matches, sum(num_wins) as total_wins  from teiserver_account_ratings tar
+    where season <= #{season_number}
+    group by user_id, rating_type_id
+    ) as temp_table
+    where tar.user_id = temp_table.user_id
+    and tar.rating_type_id = temp_table.rating_type_id
+    and tar.season = #{season_number}
+    """
```
geekingfrog, line 26: How long does that query takes? Making it part of a migration means that the table is locked until this query is done, which may take a while.
A safer way to do so is to:
* decouple the populating step from the migration
* modify the code so that if it encounter a null, it then does the query before proceeding
* run this in the background, ignoring already updated rows

It is massively more complex, but doesn't block everything until it is done.
So ultimately, it all comes down to "how long does this migration take"

jauggy:
@geekingfrog I'm getting a test failure:
```
1) test server stats works (TeiserverWeb.Tachyon.SystemTest)
Error:      test/teiserver_web/tachyon/system_test.exs:27
     match (=) failed
     code:  assert %{"data" => %{"userCount" => 2}} = Tachyon.server_stats!(client1)
     left:  %{"data" => %{"userCount" => 2}}
     right: %{
              "commandId" => "system/serverStats",
              "data" => %{"userCount" => 3},
              "messageId" => "9728eff3-803a-4b4b-9323-5926e0b77735",
              "status" => "success",
              "type" => "response"
            }
     stacktrace:
       test/teiserver_web/tachyon/system_test.exs:31: (test)
```
Is this anything to do with this PR?

geekingfrog:
This looks like a flaky test. It should be fixed after https://github.com/beyond-all-reason/teiserver/pull/644 is merged.
You can rebase after the that merge, but in the meantime you can also ignore this error, re-running the test would probably fix it.

priv/repo/migrations/20250406001309_populate_total_matches.exs
jauggy: @L-e-x-o-n can run it on integration to test the time it takes. However, we did run something similar in the past to populate the num_matches column. This time it should be faster as it is summing the account ratings table. In the past it would go through the matches table.

L-e-x-o-n: I don't have the exact time but it's quite fast, shouldn't be an issue.

L-e-x-o-n:
As discussed on Discord, I am was a bit uncertain (from the db design perspective) about adding the counter columns as opposed to running the counter queries, however the performance gains should be worth it considering that rating lookups are used quite often.

PR #644, opened 2025-04-06
Test improvements

geekingfrog:
As I was working on something else I encountered an issue where restarting a supervisor during a test would cause some issues. I tried to make it so that each test would run with a brand new supervision tree but encountered some errors with that, so I reduced it to only the tachyon system which is fairly self contained.

And also added a few other minor improvements, running the tests hundreds of time tend to unearth some issues.


test/teiserver/protocols/spring/spring_telemetry_test.exs
L-e-x-o-n, line 1: Why comment?

geekingfrog: list_client_properties doesn't exist

L-e-x-o-n: Why not delete instead?

PR #645, opened 2025-04-19
Blog improvements

Teifion:
I was discussing improvements to the blog system with Lexon and he remarked it would be nice to see a poll and image upload system. I've added them to my blog and as a favour to Lexon have back-ported them to the BAR system.

The styling is lack-lusture but at the same time I don't know what styling you're after so have left it as is with the idea you can improve it and tweak it to suit your desires.


Teifion:
I've addressed the dialyzer concerns as best as possible. On my machine I still get:
```
::warning file=lib/teiserver_web/live/microblog/admin/uploads/index_live.ex,line=28,col=7,title=no_return::Function handle_event/3 has no local return.
::warning file=lib/teiserver_web/live/microblog/admin/uploads/index_live.ex,line=32,col=31,title=call::The function call delete_upload will not succeed.
```

These refer to the `handle_event` function for deleting a file upload:
```elixir
  def handle_event("delete", %{"id" => id}, socket) do
    upload = Microblog.get_upload!(id)

    if upload.uploader_id == socket.assigns.current_user.id do
      Microblog.delete_upload(upload)
    else
      raise "Not your upload"
    end

    socket
    |> stream_delete(:user_uploads, %{id: id})
    |> noreply
  end
  ```

I've tested it manually and it works fine, I am unable to ascertain why Dialyzer is throwing said error but suffice to say it is incorrect.

File: lib/teiserver/microblog.ex [hunk 1]
```diff
@@ -159,4 +162,65 @@ defmodule Teiserver.Microblog do
...
+  defdelegate update_upload(upload, attrs), to: UploadLib
+
+  @spec delete_upload(Upload) :: {:ok, Upload} | {:error, Ecto.Changeset}
+  defdelegate delete_upload(upload), to: UploadLib
+
+  @spec change_upload(Upload) :: Ecto.Changeset
+  defdelegate change_upload(upload), to: UploadLib
+
+  @spec change_upload(Upload, map) :: Ecto.Changeset
+  defdelegate change_upload(upload, attrs), to: UploadLib
```
geekingfrog, line 225: In general, the specs should refer to types directly with the `.t()`, or calling functions like `map()`.

```elixir
  @spec list_uploads() :: [Upload.t()]
  defdelegate list_uploads(), to: UploadLib

  @spec list_uploads(list) :: [Upload.t()]
  defdelegate list_uploads(args), to: UploadLib

  @spec get_upload(Upload.id()) :: Upload.t() | nil
  defdelegate get_upload(upload_id), to: UploadLib

  @spec get_upload!(Upload.id()) :: Upload.t()
  defdelegate get_upload!(upload_id), to: UploadLib

  @spec create_upload() :: {:ok, Upload.t()} | {:error, Ecto.Changeset.t()}
  defdelegate create_upload(), to: UploadLib

  @spec create_upload(map()) :: {:ok, Upload.t()} | {:error, Ecto.Changeset.t()}
  defdelegate create_upload(attrs), to: UploadLib

  @spec update_upload(Upload.t(), map()) :: {:ok, Upload.t()} | {:error, Ecto.Changeset.t()}
  defdelegate update_upload(upload, attrs), to: UploadLib

  @spec delete_upload(Upload.t()) :: {:ok, Upload.t()} | {:error, Ecto.Changeset.t()}
  defdelegate delete_upload(upload), to: UploadLib

  @spec change_upload(Upload.t()) :: Ecto.Changeset.t()
  defdelegate change_upload(upload), to: UploadLib

  @spec change_upload(Upload.t(), map()) :: Ecto.Changeset.t()
  defdelegate change_upload(upload, attrs), to: UploadLib
```

lib/teiserver/microblog.ex
```diff
@@ -159,4 +162,65 @@ defmodule Teiserver.Microblog do
...
+  @spec list_uploads() :: [Upload]
+  defdelegate list_uploads(), to: UploadLib
+
+  @spec list_uploads(list) :: [Upload]
+  defdelegate list_uploads(args), to: UploadLib
+
+  @spec get_upload(Upload.id()) :: Upload.t() | nil
+  defdelegate get_upload(upload_id), to: UploadLib
+
+  @spec get_upload!(Upload.id()) :: Upload.t() | nil
```
geekingfrog, line 206: this function cannot return `nil`

lib/teiserver/microblog/libs/poll_response_lib.ex
```diff
@@ -0,0 +1,163 @@
...
+    %PollResponse{}
+    |> PollResponse.changeset(attrs)
+    |> Repo.insert()
+    |> broadcast_create_poll_response
+  end
+
+  defp broadcast_create_poll_response({:ok, %PollResponse{} = poll_response}) do
+    spawn(fn ->
+      # We sleep this because sometimes the message is seen fast enough the database doesn't
+      # show as having the new data (row lock maybe?)
```
geekingfrog, line 58: This is deeply suspicious. `Repo.insert` should immediately commit, and any subsequent read should see the latest value, that's the C of ACID.

lib/teiserver/microblog/libs/poll_response_lib.ex
```diff
@@ -0,0 +1,163 @@
...
+      :timer.sleep(1000)
+
+      PubSub.broadcast(
+        Teiserver.PubSub,
+        "microblog_poll_responses",
+        %{
+          channel: "microblog_poll_responses",
```
geekingfrog, line 65: Where is this subscribed to? I didn't find anything grepping through the code.

lib/teiserver/microblog/libs/poll_response_lib.ex
```diff
@@ -0,0 +1,163 @@
...
+      {:error, %Ecto.Changeset{}}
+
+  """
+  def delete_poll_response(%PollResponse{} = poll_response) do
+    Repo.delete(poll_response)
+    |> broadcast_delete_poll_response
+  end
+
+  defp broadcast_delete_poll_response({:ok, %PollResponse{} = poll_response}) do
+    PubSub.broadcast(
```
geekingfrog, line 136: there's no sleep there?

lib/teiserver/microblog/libs/poll_response_lib.ex
```diff
@@ -0,0 +1,163 @@
...
+      "microblog_poll_responses",
+      %{
+        channel: "microblog_poll_responses",
+        event: :poll_response_deleted,
+        poll_response: poll_response
+      }
+    )
+
+    {:ok, poll_response}
+  end
```
geekingfrog, line 147: Maybe a nitpick, but you could refactor the various broadcast stuff with a function that takes the event and the tuple as argument and does the rest, something like:

```elixir

  defp broadcast_poll_response({:ok, %PollResponse{} = poll_response}, event) do
    PubSub.broadcast(
      Teiserver.PubSub,
      "microblog_poll_responses",
      %{
        channel: "microblog_poll_responses",
        event: event,
        poll_response: poll_response
      }
    )

    {:ok, poll_response}
  end

# and call with
|> broadcast_poll_response(:poll_response_updated)
```

lib/teiserver/microblog/libs/post_lib.ex
```diff
@@ -197,6 +213,9 @@ defmodule Teiserver.Microblog.PostLib do
     query = "DELETE FROM microblog_post_tags WHERE post_id = $1;"
     {:ok, _} = Ecto.Adapters.SQL.query(Repo, query, [post.id])
 
+    query = "DELETE FROM microblog_poll_responses WHERE post_id = $1;"
+    {:ok, _} = Ecto.Adapters.SQL.query(Repo, query, [post.id])
```
geekingfrog, line 217: Slightly out of scope, but since you're touching this function, it would be nice to make all the deletion transactional, to avoid deleting half of the stuff if a later error comes.

lib/teiserver/microblog/schemas/poll_response.ex
```diff
@@ -0,0 +1,27 @@
+defmodule Teiserver.Microblog.PollResponse do
+  @moduledoc false
+  use TeiserverWeb, :schema
+
```
geekingfrog, line 4: This module should ideally define the types:
```elixir
  @type id :: Ecto.UUID.t()
  @type t :: %__MODULE__{
          id: id(),
          uploader: Teiserver.Account.User.t(),
          filename: String.t(),
          type: String.t(),
          file_size: integer(),
          inserted_at: DateTime.t(),
          updated_at: DateTime.t()
        }
```

This defines the type `Upload.id()`, and correctly map it to an ecto uuid. This also fixes the error you were having with dialyzer and `delete_upload`. Because it was expecting a `Ecto.UUID.t()` but was receiving a `Upload.id()` and it's not smart enough without the additional type definition that these things are the same.

mix.exs
```diff
@@ -103,7 +103,7 @@ defmodule Teiserver.MixProject do
       {:cowboy, "~> 2.9"},
       {:statistics, "~> 0.6.2"},
       {:csv, "~> 2.4"},
-      {:earmark, "~> 1.4"},
+      {:mdex, "~> 0.2"},
```
geekingfrog, line 106: Why this change? I know nothing about any of these 2 libs.

test/teiserver/microblog/microblog_test.exs
```diff
@@ -293,4 +293,142 @@ defmodule Teiserver.MicroblogTest do
...
+        filename: "some updated contents",
+        type: "some updated title",
+        file_size: 43,
+        uploader_id: user.id
+      }
+
+      assert {:ok, %Upload{} = upload} = Microblog.update_upload(upload, update_attrs)
+      assert upload.filename == "some updated contents"
+      assert upload.type == "some updated title"
+      assert upload.file_size == 43
```
geekingfrog, line 414: I don't think the last 3 asserts are of any value. Because `Repo.insert()` more or less returns the struct passed in, except for fields with `read_after_writes` or specifying `returning: xxx`.
If you want to test that the stuff in the db is indeed correct, you should query the record from the db, and perform asserts on that thing instead (or use the `:returning` feature)

lib/teiserver/microblog.ex
Teifion: Most of them do, which ones were you not happy with?

lib/teiserver/microblog/libs/poll_response_lib.ex
Teifion: In my experiments that was not the case. Happy to remove it if you want though.

lib/teiserver/microblog/libs/poll_response_lib.ex
Teifion: It's not but it's part of the way I'd used other things and wanted to provide an option for it to be subscribed to. Happy to remove it if you're not happy with it.

lib/teiserver/microblog/libs/poll_response_lib.ex
Teifion: Didn't seem to be needed here.

lib/teiserver/microblog/libs/poll_response_lib.ex
Teifion: Happy to port something else back for you guys if you want - https://github.com/durandal4x/durandal-server/blob/main/lib/durandal/helpers/pubsub_helper.ex#L33

It's not exactly what you've asked for though.

lib/teiserver/microblog.ex
Teifion: Good spot, thank you.

lib/teiserver/microblog/libs/post_lib.ex
Teifion: Good idea, updating.

lib/teiserver/microblog/schemas/poll_response.ex
Teifion: Explains the error perfectly, thank you.

test/teiserver/microblog/microblog_test.exs
Teifion: Fair

Teifion:
I've fixed a number of your concerns directly, others I expect a little more discussion to take place.

lib/teiserver/microblog.ex
geekingfrog: apart from `get_upload` and `get_upload!`, they are all using `Upload` or `Ecto.Changeset`
I would need to check my config, but the dialyzer integration with my lsp is complaining about unknown type, not sure why the CI one doesn't.

lib/teiserver/microblog/libs/poll_response_lib.ex
geekingfrog: What kind of experiments? How did you test that?
Did you have a local genserver subscribed to the topic and observed the race there?
I’m quite suspicious because that kind of artificial delay usually mask a true race condition, and I’d rather fix the underlying race than waiting and praying.

But also, alternatively, since there's no consumer for these topics, we could also remove all of this code and call it a day, it's not particularly hard to put it back either.

Teifion: I ran it locally, if I had a separate tab open to the index page of the blog it'd error saying it got a message there was a new post but that post didn't exist in the database. That said, at that stage I'd not put in the functionality of saying what type of update it was so in hindsight we can remove this part.

lib/teiserver/microblog.ex
Teifion: The existing code further up the file does it the way I've done it:

```elixir
@spec create_user_preference() :: {:ok, UserPreference} | {:error, Ecto.Changeset}
defdelegate create_user_preference(), to: UserPreferenceLib
```

Do you want me to start doing it differently or do you want to keep it consistent?

lib/teiserver/microblog/libs/poll_response_lib.ex
Teifion: Latest changes address this

lib/teiserver/helpers/pubsub_helper.ex
```diff
@@ -0,0 +1,74 @@
...
+    # Function as topic with an explicit key to pass to the topic function
+    insert_object()
+    |> broadcast_on_ok({&topic_func/1, :foreign_key_id}, :object, %{event: :created_object})
+  """
+  @spec broadcast_on_ok(
+          {:ok, any} | {:error, any},
+          String.t() | function() | {function(), atom()},
+          atom(),
+          map()
+        ) :: :ok
```
geekingfrog, line 38: The return type here should be `{:ok, any()}`

lib/teiserver/microblog.ex
```diff
@@ -159,4 +162,65 @@ defmodule Teiserver.Microblog do
...
+  @spec list_uploads() :: [Upload.t()]
+  defdelegate list_uploads(), to: UploadLib
+
+  @spec list_uploads(list) :: [Upload.t()]
+  defdelegate list_uploads(args), to: UploadLib
+
+  @spec get_upload(Upload.id()) :: Upload.t() | nil
+  defdelegate get_upload(upload_id), to: UploadLib
+
+  @spec get_upload(Upload.id()) :: Upload.t()
```
geekingfrog, line 206: The spec here should be for `get_upload!`, with a `!` at the end.

lib/teiserver/microblog/libs/upload_lib.ex
```diff
@@ -0,0 +1,106 @@
...
+      ** (Ecto.NoResultsError)
+  """
+  @spec get_upload!(Ecto.UUID.t()) :: Upload
+  def get_upload!(upload_id) do
+    [id: upload_id]
+    |> UploadQueries.query_uploads()
+    |> Repo.one!()
+  end
+
+  @spec get_upload!(Ecto.UUID.t(), list) :: Upload
```
geekingfrog, line 35: the return type should be `Upload.t()`

lib/teiserver/microblog/libs/upload_lib.ex
```diff
@@ -0,0 +1,106 @@
...
+  @doc """
+  Gets a single upload.
+  Raises `Ecto.NoResultsError` if the Upload does not exist.
+  ## Examples
+      iex> get_upload!(123)
+      %Upload{}
+      iex> get_upload!(456)
+      ** (Ecto.NoResultsError)
+  """
+  @spec get_upload!(Ecto.UUID.t()) :: Upload
```
geekingfrog, line 28: the return type should be `Upload.t()`

lib/teiserver/microblog/libs/post_lib.ex
```diff
@@ -194,11 +210,16 @@ defmodule Teiserver.Microblog.PostLib do
...
+      {:ok, _} = Ecto.Adapters.SQL.query(Repo, query, [post.id])
+
+      query = "DELETE FROM microblog_poll_responses WHERE post_id = $1;"
+      {:ok, _} = Ecto.Adapters.SQL.query(Repo, query, [post.id])
 
-    Repo.delete(post)
-    |> broadcast_delete_post
+      Repo.delete(post)
+      |> broadcast_delete_post
+    end)
```
geekingfrog, line 222: The return of the transaction is `{:ok, stuff} | {:error, tx_error}`, where stuff is whatever is returned from the tx block. So when that works all fine, you get `{:ok, {:ok, post}}`. You need to pattern match on that to unwrap the extra `:ok, ...` tuple.
This is the source of the test failure at `test/teiserver/microblog/microblog_test.exs:133`

geekingfrog (commented):
Looking good. I posted a few comments to help you fix the dialyzer problems (these were a bit weird) and fix the test.
Should be good to go once the CI is green

lib/teiserver/microblog.ex
geekingfrog: Hum, ideally keep it consistent, but I'm not too precious about it really, up to you, as long as CI passes that's the important bit.

lib/teiserver/microblog/libs/upload_lib.ex
Teifion: Surprised that didn't error.

geekingfrog (approved):
maybe squash merge that, only the first 3 commits make sense to be separate, but not sure it's worth the hastle to neatly rewrite the history.

L-e-x-o-n:
Thanks

Issue #237, opened 2024-03-27
Application can't start when discord is unavailable


p2004a:
When Discord times out, is unavailable etc, teiserver can't start, crashes, and *of course* disabling Discord integration is non trivial and requires recompilation, thus Discord is a critical startup dependency for teiserver which is pretty much unacceptable.

I'm not sure what is the behavior when discord suddenly has outage when teiserver is already running.

StanczakDominik:
Closed by #281

p2004a:
@StanczakDominik Is it actually fixing this issue? Is the code rewritten so that when discord is enabled in configuration, application doesn't crash when discord service is unavailable?

StanczakDominik:
Ah, fair - I thought that was within the scope of that PR. Thanks for keeping me honest!

NortySpock:
Is this still a high-priority issue? 

1) it looks like relatively recent changes (https://github.com/beyond-all-reason/teiserver/pull/597/files#diff-e43c5cbf91db7a8062b6cb860cbf118296c1b4c7ee32fdcf702e54234ba38092) turned Discord integration into an environment variable, so that (presumably) manual recovery during a discord outage is "just" (a) flip the ENV_VAR to false (b) restart teiserver, which sounds less invasive than (a) edit config (b) recompile (c) restart teiserver.

2) Not that I know anything about Discord or its APIs or how to set up some "local" developer testing, but I did notice that https://github.com/Kraigie/nostrum/releases/tag/v0.10.4 mentions a fix to the ratelimiter that fixed a hanging bug when the HTTP connection died abnormally... wonder if that also works for connection failures on application startup?

3) Provided I can find a discord server to test in (again, a noob at Discord but willing to plug away at the problem slowly on the weekends), maybe I can figure out how to shim or tweak the discord bridge to degrade gracefully under API failures?

p2004a:
> manual recovery during a discord outage is "just" (a) flip the ENV_VAR to false

Server crashing for some other reason (e.g we had ooms in the past) and not being able to start up afterwards e.g. during EU night when everyone is asleep would be bad. There is no time box when someone would do manual intervention and figure out env variable to flip.

It's an unlikely scenario so I'm not sure about priority but it would be good to fix it.

PR #649, opened 2025-04-26
address "Application can't start when discord is unavailable"

NortySpock:
Should address the startup issue in #237 

 - Moves Nostrum.Application / API into a DynamicSupervisor that basically just logs any Nostrum errors or crashes
 - This allows Teiserver to start despite Discord API failure
    - (I simulated that by manually induced failure by using an invalid access token)
    - This would make Discord integration a one-shot attempt. After teiserver startup Discord connection would not be re-attempted until the next teiserver restart. 
 - Connection appears to succeed as expected when Nostrum is provided correct Discord credential information: 
    >  image

 - Does not retain a deliberate queue of messages to eventually send to discord, but _could_ be a source of slow memory leaks or something as messages for Nostrum pile up in a process message queue. Nonetheless, I assume this could limp for several hours. possibly days (unsure of memory growth rate), until someone is online to investigate or restart the server to allow another discord connection attempt.

 - _note that this code was written by me in heavy consultation with Microsoft Copilot. I think I understand it, I would vouch that it appears to be straightforward, but (a) it lacks true tests and (b) I am a novice in the world of Elixir, so it could have subtle bugs that I am just blind to._

---------------------------------

Example output on invalid Discord token (which previously crashed Teiserver):

> 2025-04-26 00:24:21.997 [error] pid=<0.612.0>  Nostrum failed: {%RuntimeError{message: "Invalid token format. Copy it again from the \"Bot\" tab of your Application in the Discord Developer Portal."}, [{Nostrum.Token, :check_token!, 1, [file: ~c"lib/nostrum/token.ex", line: 46, error_info: %{module: Exception}]}, {Nostrum.Application, :start, 2, [file: ~c"lib/nostrum/application.ex", line: 24]}, {DynamicSupervisor, :start_child, 3, [file: ~c"lib/dynamic_supervisor.ex", line: 800]}, {DynamicSupervisor, :handle_start_child, 2, [file: ~c"lib/dynamic_supervisor.ex", line: 786]}, {:gen_server, :try_handle_call, 4, [file: ~c"gen_server.erl", line: 1131]}, {:gen_server, :handle_msg, 6, [file: ~c"gen_server.erl", line: 1160]}, {:proc_lib, :init_p_do_apply, 3, [file: ~c"proc_lib.erl", line: 241]}]}. DiscordBridgeBot will NOT start.

----------------------------------

Example internal log output on failing to send a chat message from `#main` to Discord `#main` (because Nostrum Application / API is unavailable): 

```
 
2025-04-26 00:30:40.854 [error] request_id=BridgeServer pid=<0.3802.0>  GenServer #PID<0.3802.0> terminating
** (ArgumentError) errors were found at the given arguments:

  * 1st argument: out of range

    (erts 14.2.5.2) :erlang.phash2("/channels/1361893231796813955/messages", 0)
    (nostrum 0.10.1) lib/nostrum/api/ratelimiter_group.ex:44: Nostrum.Api.RatelimiterGroup.limiter_for_bucket/1
    (nostrum 0.10.1) lib/nostrum/api/ratelimiter.ex:982: Nostrum.Api.Ratelimiter.queue/1
    (nostrum 0.10.1) lib/nostrum/api/message.ex:110: Nostrum.Api.Message.create/2
    (teiserver 0.1.0) lib/teiserver/bridge/bridge_server.ex:188: Teiserver.Bridge.BridgeServer.handle_info/2
    (stdlib 5.2.3.1) gen_server.erl:1095: :gen_server.try_handle_info/3
    (stdlib 5.2.3.1) gen_server.erl:1183: :gen_server.handle_msg/6
    (stdlib 5.2.3.1) proc_lib.erl:241: :proc_lib.init_p_do_apply/3
Last message: {:new_message, 16, "main", "tharp test"}
State: %{user: %{rank: 0, id: 792, lobby_hash: "IC", last_played: nil, email: "bridge@teiserver", clan_id: nil, discord_id: nil, icon: "fa-brands fa-discord", colour: "#0066AA", last_login_timex: ~U[2025-04-26 05:30:37.551000Z], last_login_mins: 29094091, inserted_at: ~N[2025-04-24 02:49:59], spring_password: false, restricted_until: nil, last_login: 29094091, chobby_hash: nil, permissions: [], name: "DiscordBridgeBot", verified: true, country: "GB", email_change_code: nil, moderator: false, bot: true, shadowbanned: false, steam_id: nil, print_client_messages: false, hw_hash: nil, restrictions: [], discord_dm_channel: nil, lobby_client: "Teiserver Internal Client", password_hash: nil, discord_dm_channel_id: nil, smurf_of_id: nil, roles: [], print_server_messages: false, last_logout: nil}, ip: "127.0.0.1", username: "DiscordBridgeBot", client: %{rank: 0, team_colour: "0", in_game: false, unready_at: nil, lobby_host: false, role: "spectator", ip: "127.0.0.1", ready: false, temp_mute_count: 0, awaiting_warn_ack: false, muted: false, player_number: 0, sync: %{map: 0, engine: 0, game: 0}, token_id: nil, userid: 792, lobby_id: nil, clan_tag: nil, party_id: nil, name: "DiscordBridgeBot", warned: false, country: "GB", moderator: false, player: false, bot: false, queues: [], shadowbanned: false, print_client_messages: false, tcp_pid: #PID<0.3802.0>, chat_times: [], protocol: :internal, lobby_client: "Teiserver Internal Client", handicap: 0, away: false, side: 0, connected: true, team_number: 0, app_status: nil, print_server_messages: false, restricted: false}, userid: 792, channel_lookup: %{"gdt-discussion" => 0, "gdt-voting" => 0, "main" => 1361893231796813955, "moderation-actions" => 0, "moderation-reports" => 0, "newbies" => 0, "promote" => 0, "server-updates" => 1361895967955353721, "telemetry-infologs" => 0}, room_lookup: %{0 => "promote", 1361893231796813955 => "main"}, lobby_host: false, recent_bridged_messages: %{}}

```

### DIFF
--- a/lib/teiserver/bridge/discord_supervisor.ex
+++ b/lib/teiserver/bridge/discord_supervisor.ex
@@ -6,8 +6,7 @@ defmodule Teiserver.Bridge.DiscordSupervisor do
   end
 
   def init(_) do
-    Supervisor.init([{Nostrum.Application, []}, {Teiserver.Bridge.DiscordBridgeBot, []}],
-      restart: :temporary,
+    Supervisor.init([Nostrum.Application, Teiserver.Bridge.DiscordBridgeBot],
       strategy: :rest_for_one
     )
   end
--- a/lib/teiserver/bridge/discord_supervisor.ex
+++ b/lib/teiserver/bridge/discord_supervisor.ex
@@ -6,7 +6,8 @@ defmodule Teiserver.Bridge.DiscordSupervisor do
   end
 
   def init(_) do
-    Supervisor.init([Nostrum.Application, Teiserver.Bridge.DiscordBridgeBot],
+    Supervisor.init(
+      [Nostrum.Application, Teiserver.Bridge.BridgeServer, Teiserver.Bridge.DiscordBridgeBot],
       strategy: :rest_for_one
     )
   end
--- a/lib/teiserver/bridge/discord_supervisor.ex
+++ b/lib/teiserver/bridge/discord_supervisor.ex
@@ -1,0 +1,14 @@
+defmodule Teiserver.Bridge.DiscordSupervisor do
+  use Supervisor
+
+  def start_link(init_arg) do
+    Supervisor.start_link(__MODULE__, init_arg, name: __MODULE__)
+  end
+
+  def init(_) do
+    Supervisor.init([{Nostrum.Application, []}, {Teiserver.Bridge.DiscordBridgeBot, []}],
+      restart: :temporary,
+      strategy: :rest_for_one
+    )
+  end
+end
--- a/lib/teiserver/bridge/discord_system.ex
+++ b/lib/teiserver/bridge/discord_system.ex
@@ -10,12 +10,14 @@ defmodule Teiserver.Bridge.DiscordSystem do
   def init(_) do
     pid = DynamicSupervisor.init(strategy: :one_for_one)
 
-    Task.async(fn ->
-      DynamicSupervisor.start_child(
-        Teiserver.Bridge.DiscordSystem,
-        Supervisor.child_spec(Teiserver.Bridge.DiscordSupervisor, restart: :temporary)
-      )
-    end)
+    if Teiserver.Communication.use_discord?() do
+      Task.async(fn ->
+        DynamicSupervisor.start_child(
+          __MODULE__,
+          Supervisor.child_spec(Teiserver.Bridge.DiscordSupervisor, restart: :temporary)
+        )
+      end)
+    end
 
     pid
   end
--- a/lib/teiserver/bridge/discord_system.ex
+++ b/lib/teiserver/bridge/discord_system.ex
@@ -8,7 +8,7 @@ defmodule Teiserver.Bridge.DiscordSystem do
 
   @impl true
   def init(_) do
-    pid = DynamicSupervisor.init(strategy: :one_for_one)
+    {:ok, sup_flags} = DynamicSupervisor.init(strategy: :one_for_one)
 
     if Teiserver.Communication.use_discord?() do
       Task.async(fn ->
@@ -19,6 +19,6 @@ defmodule Teiserver.Bridge.DiscordSystem do
       end)
     end
 
-    pid
+    {:ok, sup_flags}
   end
 end
--- a/lib/teiserver/bridge/discord_system.ex
+++ b/lib/teiserver/bridge/discord_system.ex
@@ -1,22 +1,22 @@
 defmodule Teiserver.Bridge.DiscordSystem do
-  use Supervisor
+  use DynamicSupervisor
+  use Task
 
-  def start_link(init_arg) do
-    Supervisor.start_link(__MODULE__, init_arg, name: __MODULE__)
+  def start_link(_) do
+    DynamicSupervisor.start_link(__MODULE__, [], name: __MODULE__)
   end
 
   @impl true
   def init(_) do
-    children =
-      if Teiserver.Communication.use_discord?() do
-        [Teiserver.Bridge.DiscordSupervisor]
-      else
-        []
-      end
+    pid = DynamicSupervisor.init(strategy: :one_for_one)
 
-    Supervisor.init(children,
-      restart: :temporary,
-      strategy: :rest_for_one
-    )
+    Task.async(fn ->
+      DynamicSupervisor.start_child(
+        Teiserver.Bridge.DiscordSystem,
+        Supervisor.child_spec(Teiserver.Bridge.DiscordSupervisor, restart: :temporary)
+      )
+    end)
+
+    pid
   end
 end
--- a/lib/teiserver/bridge/discord_system.ex
+++ b/lib/teiserver/bridge/discord_system.ex
@@ -1,0 +1,22 @@
+defmodule Teiserver.Bridge.DiscordSystem do
+  use Supervisor
+
+  def start_link(init_arg) do
+    Supervisor.start_link(__MODULE__, init_arg, name: __MODULE__)
+  end
+
+  @impl true
+  def init(_) do
+    children =
+      if Teiserver.Communication.use_discord?() do
+        [Teiserver.Bridge.DiscordSupervisor]
+      else
+        []
+      end
+
+    Supervisor.init(children,
+      restart: :temporary,
+      strategy: :rest_for_one
+    )
+  end
+end
--- a/lib/teiserver/application.ex
+++ b/lib/teiserver/application.ex
@@ -126,7 +126,7 @@ defmodule Teiserver.Application do
         {DynamicSupervisor, strategy: :one_for_one, name: Teiserver.Throttles.Supervisor},
 
         # Bridge
-        Teiserver.Bridge.BridgeServer,
+        Teiserver.Bridge.DiscordSystem,
         concache_sup(:discord_bridge_dm_cache),
         concache_perm_sup(:discord_channel_cache),
         concache_sup(:discord_bridge_account_codes, global_ttl: 300_000),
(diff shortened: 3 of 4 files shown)


geekingfrog (commented):
I think the core idea: making the discord stuff `:temporary` is solid, but there are 2 main issues:

1. the biggest one: I haven't fully understood the error you posted, but a genserver terminating because discord isn't there is not great. I would prefer doing nothing when discord is disabled or down.
2. The init is a but sus with some code all over the place.

What I would recommend is to unconditionally add a new supervisor to the main tree: `DiscordSystem`. The init function should contain the logic about enabled/disabled and add children to itself accordingly.
Then, I would also link nostrum with the bridge, so that if nostrum goes down, then the bridge as well. You can do that with `restart: rest_for_one`. This may fix the issue with the genserver exiting btw, because in this case it would behave as if everything was disabled.

What I have in mind is something like that:
```
DiscordSystem
  based on config, add one child:
  DiscordSupervisor restart: :temporary, strategy: :rest_for_one
    ├ Nostrum
    └ Discordbridge
```

Also this way, to restart the bridge, one can open a repl and call `restart_child` on the `DiscordSupervisor` module to restart everything at once.

Finally, I do not understand what you wrote about message piling up. Where would these messages be?

NortySpock:
Thanks for the review, it may be a week before I can get time take a swing at implementing the DiscordSystem you suggest. That definitely sounds like a more clear architecture than I had cobbled together. 

As for "messages piling up" I had just assumed that messages sent to a pid that dies went into a fifo queue (e.g. Kafka, NATS JetStream, and other persistent message queues),  but I guess they are just dropped. My mistake.

lib/teiserver/bridge/discord_supervisor.ex
```diff
@@ -0,0 +1,14 @@
+defmodule Teiserver.Bridge.DiscordSupervisor do
+  use Supervisor
+
+  def start_link(init_arg) do
+    Supervisor.start_link(__MODULE__, init_arg, name: __MODULE__)
+  end
+
+  def init(_) do
+    Supervisor.init([{Nostrum.Application, []}, {Teiserver.Bridge.DiscordBridgeBot, []}],
+      restart: :temporary,
```
geekingfrog: the warning is because there's no `restart` option for this function.

lib/teiserver/bridge/discord_system.ex
```diff
@@ -0,0 +1,22 @@
...
+  def init(_) do
+    children =
+      if Teiserver.Communication.use_discord?() do
+        [Teiserver.Bridge.DiscordSupervisor]
+      else
+        []
+      end
+
+    Supervisor.init(children,
+      restart: :temporary,
```
geekingfrog: same as above

lib/teiserver/bridge/discord_supervisor.ex
```diff
@@ -0,0 +1,14 @@
+defmodule Teiserver.Bridge.DiscordSupervisor do
+  use Supervisor
+
+  def start_link(init_arg) do
+    Supervisor.start_link(__MODULE__, init_arg, name: __MODULE__)
+  end
+
+  def init(_) do
+    Supervisor.init([{Nostrum.Application, []}, {Teiserver.Bridge.DiscordBridgeBot, []}],
```
geekingfrog: `{Module, []}` is the same as `Module`.

lib/teiserver/bridge/discord_system.ex
```diff
@@ -0,0 +1,22 @@
...
+    DynamicSupervisor.start_link(__MODULE__, [], name: __MODULE__)
+  end
+
+  @impl true
+  def init(_) do
+    pid = DynamicSupervisor.init(strategy: :one_for_one)
+
+    Task.async(fn ->
+      DynamicSupervisor.start_child(
+        Teiserver.Bridge.DiscordSystem,
```
geekingfrog: you can use `__MODULE__` here

lib/teiserver/bridge/discord_system.ex
```diff
@@ -0,0 +1,22 @@
...
+  def start_link(_) do
+    DynamicSupervisor.start_link(__MODULE__, [], name: __MODULE__)
+  end
+
+  @impl true
+  def init(_) do
+    pid = DynamicSupervisor.init(strategy: :one_for_one)
+
+    Task.async(fn ->
+      DynamicSupervisor.start_child(
```
geekingfrog: this should still be conditional on `use_discord?()`

lib/teiserver/bridge/discord_supervisor.ex
```diff
@@ -0,0 +1,13 @@
+defmodule Teiserver.Bridge.DiscordSupervisor do
+  use Supervisor
+
+  def start_link(init_arg) do
+    Supervisor.start_link(__MODULE__, init_arg, name: __MODULE__)
+  end
+
+  def init(_) do
+    Supervisor.init([Nostrum.Application, Teiserver.Bridge.DiscordBridgeBot],
```
geekingfrog: I would move `Teiserver.Bridge.BridgeServer` here as well, it must be started before DiscordBridgeBot it seems. Although looking at the code there, it seems a bit cursed. Maybe needs some logic/catch there on the various `{call,cast,send}_bridge` functions to handle the case where the bridge is just not there.

lib/teiserver/bridge/discord_system.ex
NortySpock: Doh! Yep, will fix that...

lib/teiserver/bridge/discord_supervisor.ex
NortySpock: Moved the Teiserver.Bridge.BridgeServer here. I agree it's at least slightly cursed and probably prone to data races, but at least moving BridgeServer to be part of the supervision tree means the things it depends on are probably going to be up in time for it to be able to use them.

NortySpock:
I rebased and re-tested on top of master. Currently Chobby can send to Discord just fine, but when a message is placed in discord (to be bridged to Chobby), I get the following error. Seems pretty clear that an array is being passed to the Regex rather than a string. I'll try to dig into git blame to figure out what is going on...

```
2025-06-13 07:54:45.125 [error] pid=<0.1646.0>  Error in event handler: ** (FunctionClauseError) no function clause matching in Regex.match?/2
    (elixir 1.18.2) lib/regex.ex:319: Regex.match?(~r/superinnapropriateword/, ["nortyspock: lets see if this runs"])
    (elixir 1.18.2) lib/enum.ex:2546: Enum."-reduce/3-lists^foldl/2-0-"/3
    (teiserver 0.1.0) lib/teiserver/chat/chat_room_cache.ex:176: Teiserver.Room.send_message/3
    (teiserver 0.1.0) deps/nostrum/lib/nostrum/consumer.ex:448: anonymous fn/1 in Teiserver.Bridge.DiscordBridgeBot.handle_info/2
    (elixir 1.18.2) lib/task/supervised.ex:101: Task.Supervised.invoke_mfa/2

```

NortySpock:
Fixed the above error with a small shim to unwrap the case when send_message was passed a list of messages, rather than a `msg` string.

Verified bidirectional bridging manually and took a screenshot:

Screenshot_20250625_061853_showing_discord_bridge_working_bidirectionally

lib/teiserver/bridge/discord_system.ex
```diff
@@ -0,0 +1,24 @@
...
+  use DynamicSupervisor
+  use Task
+
+  def start_link(_) do
+    DynamicSupervisor.start_link(__MODULE__, [], name: __MODULE__)
+  end
+
+  @impl true
+  def init(_) do
+    pid = DynamicSupervisor.init(strategy: :one_for_one)
```
geekingfrog: the variable name is misleading, this is not a `pid`, but `{:ok, sup_flags()}` as per [the doc](https://hexdocs.pm/elixir/DynamicSupervisor.html#init/1)

NortySpock: Thanks, still learning Elixir syntax and conventions. 

Updated the response tuple; how does it look now?

PR #660, opened 2025-05-18
Add max heap size to SpringTcpServer

L-e-x-o-n:
Sets [max_heap_size](https://www.erlang.org/docs/23/man/erlang#process_flag_max_heap_size) for SpringTcpServer to ~20 MB

For info on how Elixir's GC works https://www.erlang.org/doc/apps/erts/garbagecollection. 
If my understanding is correct, GC will attempt to collect when the process reaches the `max_heap_size` and if it isn't successful in reducing the heap size, the process will be killed (configurable as well, default true).

### DIFF
--- a/lib/teiserver/tcp/spring/spring_tcp_server.ex
+++ b/lib/teiserver/tcp/spring/spring_tcp_server.ex
@@ -42,7 +42,8 @@ defmodule Teiserver.SpringTcpServer do
     mode = if opts[:ssl], do: :ranch_ssl, else: :ranch_tcp
 
     # https://www.erlang.org/docs/23/man/erlang#process_flag_max_heap_size
-    max_heap_size = 2_620_000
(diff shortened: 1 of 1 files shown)


lib/teiserver/tcp/spring/spring_tcp_server.ex
```diff
@@ -41,6 +41,9 @@ defmodule Teiserver.SpringTcpServer do
   def start_link(opts) do
     mode = if opts[:ssl], do: :ranch_ssl, else: :ranch_tcp
 
+    # https://www.erlang.org/docs/23/man/erlang#process_flag_max_heap_size
+    max_heap_size = 2_620_000
```
geekingfrog, line 45: maybe a comment as to what is this number, or replace it with an expression like `1024 * 1024 * 20`

geekingfrog (approved):

PR #664, opened 2025-05-23
Web registration fixes

geekingfrog:
* Handle the verification. This part is still quite cursed. There is something in `CacheUser` that does similar thing, but take the password as md5. None of it makes sense in term of code organisation (why in cache user????). So I opted for the hack of copy/pasting the stuff D:
* fix the lack of styling

### DIFF
--- a/lib/teiserver/account.ex
+++ b/lib/teiserver/account.ex
@@ -2200,8 +2200,8 @@ defmodule Teiserver.Account do
     })
   end
 
-  @spec encrypt_password(any) :: binary | {binary, binary, {any, any, any, any, any}}
-  def encrypt_password(password) do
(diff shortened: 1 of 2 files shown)


L-e-x-o-n:
Might be nice to unify the 2 registration flows and use the same functions considering the only difference should be in password being in MD5 or plan text format, I can revisit this later while also fixing an annoying verification bug.

geekingfrog:
Alright, take 2. I refactored the chobby registration to extract the validation and verification, and reuse that for the web registration.
I'm still uneasy about the whole "cache user" thing, but that's another story.
I did test registering a new user with local chobby and could connect so it should be fine.

geekingfrog:
More changes, with more code sharing between the 2 paths now.
Asides the existing tests, I did register with chobby -> can login with both chobby and web. And same for a web registered account.

PR #667, opened 2025-05-28
Set startboxes for matchmaking

geekingfrog:
make use of the startbox set from the maps when creating the startscript

### DIFF
--- a/lib/teiserver/asset/libs/map_lib.ex
+++ b/lib/teiserver/asset/libs/map_lib.ex
@@ -24,4 +24,41 @@ defmodule Teiserver.Asset.MapLib do
       err -> err
     end
   end
+
+  @type startbox :: %{top: integer(), bottom: integer(), left: integer(), right: integer()}
+
+  @doc """
+  Return the suitable startboxes for the given match and the number of (ally) teams
+  nil if nothing matches
+  """
+  @spec get_startboxes(Asset.Map.t(), number_of_teams :: non_neg_integer()) :: [startbox()] | nil
+  def get_startboxes(%Asset.Map{} = map, number_of_teams) do
+    # By construction, there's only one startbox definition in the set for a given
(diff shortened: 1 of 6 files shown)


lib/teiserver/asset/libs/map_lib.ex
```diff
@@ -24,4 +24,41 @@ defmodule Teiserver.Asset.MapLib do
...
+    # wants [0;1], so rescale here
+    # originally, the spring protocol defines ADDSTARTRECT for the startboxes
+    # these are using [0;200]x[0;200] coordinates
+    # The map metadata service is using the same coordinate and this is what's
+    # saved in teiserver
+    # However, tachyon and engine require the startboxes to be within [0;1] so
+    # scale that here. This may not stay this way if/when the map metadata
+    # service changes the coordinates
+    # https://discord.com/channels/549281623154229250/927564746104905728/1303302250713583727
+    %{bottom: br["y"] / 200, left: tl["x"] / 200, top: tl["y"] / 200, right: br["x"] / 200}
```
NortySpock, line 62: (nit) Seems like this long comment about the assumptions this code makes (which is an excellent comment) also means it would be a good place to put an `assert`, `refute`, or a unit test to exercise the code.

test/teiserver_web/tachyon/matchmaking_test.exs
```diff
@@ -88,6 +89,22 @@ defmodule Teiserver.Tachyon.MatchmakingTest do
     }
   end
 
+  defp add_startboxes(attrs, n_teams) do
```
NortySpock, line 92: (nit) It seems to me that it would be valuable to make this a public function and add a single unit test demonstrating and testing the behavior. 

Not that I think it is super risky, but just to satisfy the curious that it was tested and can easily have tests added to it if edge cases are found.

PR #668, opened 2025-05-30
Fix moderator role management

L-e-x-o-n:
Fixes the bug where moderators would lose their own `Moderator` role when editing their other roles or properties.

### DIFF
--- a/lib/teiserver/account/libs/role_lib.ex
+++ b/lib/teiserver/account/libs/role_lib.ex
@@ -278,21 +278,16 @@ defmodule Teiserver.Account.RoleLib do
 
   def allowed_role_management("Admin") do
     staff_roles() ++
-      community_roles() ++ privileged_roles() ++ allowed_role_management("Moderator")
+      community_roles() ++
+      privileged_roles() ++ moderation_roles() ++ allowed_role_management("Moderator")
   end
 
   def allowed_role_management("Moderator") do
-    global_roles() ++ moderation_roles() ++ property_roles()
+    global_roles() ++ property_roles()
   end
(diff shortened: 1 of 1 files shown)


L-e-x-o-n:
For context.
From `TeiserverWeb.Admin.UserController`:
```ex
changeable_roles =
      cond do
        Enum.member?(current_user.roles, "Server") ->
          RoleLib.allowed_role_management("Server")

        Enum.member?(current_user.roles, "Admin") ->
          RoleLib.allowed_role_management("Admin")

        Enum.member?(current_user.roles, "Moderator") ->
          RoleLib.allowed_role_management("Moderator")

        true ->
          []
      end

    # Go through all roles, any we're not allowed to change
    # we leave as they are, any we are we do stuff with
    new_roles =
      RoleLib.all_role_names()
      |> Enum.map(fn role_name ->
        selected =
          if Enum.member?(changeable_roles, role_name) do
            user_params[role_name] == "true"
          else
            Enum.member?(user.roles, role_name)
          end

        if selected do
          role_name
        end
      end)
      |> Enum.reject(&(&1 == nil))
      |> Enum.uniq()
```

This code only updates the roles the user is allowed to change. No role should be able to edit themselves, however due to a bug Moderators had their own role in the list of roles they were allowed to manage (defined in `RoleLib.allowed_role_management("Moderator")`) , the form was still hidden with a different check so they couldn't actually change these.
So because "Moderator" was a changeable role and they couldn't see the form, the ``user_params["Moderator]` == true"` check would always return false and consequently their moderator role would be removed when updating the user.

PR #670, opened 2025-06-05
Fix bruteforce matchmaking

geekingfrog:
Fix a bug with the existing bruteforce matchmaking algo. Also introduce the basis for a later algo, still based on bruteforce, where we can check for things like OS and wait time.


test/teiserver/matchmaking/algos_test.exs
```diff
@@ -0,0 +1,74 @@
...
+      members = [mk_member(0), mk_member([1, 2]), mk_member(3)]
+      assert {:match, [team]} = Algos.ignore_os(2, 2, members)
+      counts = Enum.map(team, &Enum.count/1)
+      assert MapSet.new(counts) == MapSet.new([1, 2])
+    end
+
+    test "ordering doesn't matter" do
+      members = [mk_member(0), mk_member(1), mk_member([2, 3]), mk_member([4, 5])]
+      assert {:match, _} = Algos.ignore_os(3, 2, members)
+    end
```
geekingfrog, line 60: that test would fail with the previous `greedy_match` function

L-e-x-o-n:
We already have `Teiserver.Helper.CombinationsHelper` as well as the [Combination] (https://hexdocs.pm/combination/Combination.html)library. Maybe some of those could be reused, if they can't I would still move the new functions to the existing helper.

lib/teiserver/matchmaking/algos.ex
```diff
@@ -0,0 +1,73 @@
+defmodule Teiserver.Matchmaking.Algos do
+  @moduledoc """
+  The simplest possible matchmaking algoright, ignore OS values, and simply
+  put players together, respecting parties and team size
+  """
```
L-e-x-o-n: The module name suggests it contains multiple algos while the doc sounds like it is intended to have only one algo (even if only one exists at the moment).
If the intention is for each algorithm to have it's own module (which I think should be the case) then a better module name would be `Teiserver.Matchmaking.Algorithms.IgnoreRating` or similar.

L-e-x-o-n: There may not be as many MM algos and they may be handled in a different way in the future but I like the way multiple balance algos are handled. Seems very modular and easy to extend and add new algos.
```ex
@spec algorithm_modules() :: %{String.t() => module}
def algorithm_modules() do
  %{
    "loser_picks" => Teiserver.Battle.Balance.LoserPicks,
    "force_party" => Teiserver.Battle.Balance.ForceParty,
    "brute_force" => Teiserver.Battle.Balance.BruteForce,
    "split_noobs" => Teiserver.Battle.Balance.SplitNoobs,
    "auto" => Teiserver.Battle.Balance.AutoBalance,
    "respect_avoids" => Teiserver.Battle.Balance.RespectAvoids
  }
end
  ```
each module having a `perform` function so they can be called like this

```ex
balance_result =
  case algorithm_modules()[algo_name] do
    nil ->
      raise "No balance module by the name of '#{algo_name}'"

    m ->
      m.perform(expanded_groups, team_count, opts)
  end
 ```

geekingfrog: I have reworded the comments to be less misleading.
I'm going to keep the unique module for now, and perhaps split once there is another algo. I am a bit on the fence about the approach, because that restrict all algos to have the same signature, but perhaps that's the point. We'll see, I don't expect to ever have more than 4 algo though (ignore os for testing, 1 fallback, 1 calling an external service for very fancy logic, and perhaps one for general testing if we want to evolve the algos in a gradual manner).

geekingfrog:
> We already have `Teiserver.Helper.CombinationsHelper` as well as the [Combination](https://hexdocs.pm/combination/Combination.html) library. Maybe some of those could be reused and if they can't I would move the new functions to the existing helper.

Oh, somehow I missed that. I reworked the existing code and rolled it into the combination module

lib/teiserver/matchmaking/algos.ex
L-e-x-o-n: That is the point. I also think having different algos in different would be more organised as they could get quite complex. Doesn't matter much for now when there is only one.

PR #671, opened 2025-06-08
Allow !cv joinas when bots/AI are present?

aaronmorgan1989:
Here is my suggested change to allow !joinas for AI players. This checks the lobby for AI and does not alias "!cv joinas" and "!callvote joinas" when AI are present.


aaronmorgan1989:
Ok I see that was not correct. I will take a better look at this when I am not travelling. Thank you for your time

aaronmorgan1989:
it is not clear to me what the purpose is of aliasing joinas with send_direct_message. So I am not sure if this latest attempt is correct. I assumed it was only to prevent workarounds? 
I also noticed chat_lib uses String.split, but this is missing from cache_user. Was this an oversight or intentional?

PR #680, opened 2025-06-13
More generic logger metadata

geekingfrog:
To avoid having to add more and more keywords to the list of metadata to log, unify them under a tuple (type, id) that can be used across different systems.

@L-e-x-o-n do you think I should also modify the other metadata across the code? There are about 20 total, things like:
```elixir
Logger.metadata(request_id: "PartyServer##{id}")
# could be turned into
Logger.metadata(actor_type: :party, actor_id: id)
```

I don't use or even have access to these logs, so I haven't touched anything outside tachyon.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -98,7 +98,7 @@ config :teiserver, Teiserver,
 
 config :logger, :default_handler, false
 
-metadata = [:request_id, :user_id, :queue_id, :pid]
+metadata = [:request_id, :user_id, :pid, :actor_type, :actor_id]
 
 config :logger, LoggerBackends.Console,
   format: "$date $time [$level] $metadata $message\n",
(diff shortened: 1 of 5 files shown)


L-e-x-o-n:
If you could that would be nice. It would also mean we can remove `request_id` completely in that case, right?
I would prefer for the `actor_type` value to be the same as the file name where possible (it would make it easier and faster to find the relevant file) so `actor_type: :party_server` in your example.

geekingfrog:
> If you could that would be nice. It would also mean we can remove `request_id` completely in that case, right? I would prefer for the `actor_type` value to be the same as the file name where possible (it would make it easier and faster to find the relevant file) so `actor_type: :party_server` in your example.

I'd keep `request_id`, it's used by phoenix, and it makes sense in a web context.
As for the file, we can enable `:file, :line` in the logger config and these would be automatically added like:

```
2025-06-13 10:47:35.708 [info] request_id=MatchMonitorServer pid=<0.2850.0> file=lib/teiserver/data/client.ex line=167  Client connected: AutohostMonitor using Teiserver Internal Client
```

PR #686, opened 2025-06-21
Allow voting for joinas when bots are in lobby

aaronmorgan1989:
This will allow !cv joinas and !callvote joinas when bots are in the lobby.
I closed [PR#671](https://github.com/beyond-all-reason/teiserver/pull/671) because github is confusing and I needed a fresh start. I believe I am doing this properly now.
Changed the ai check from pr671 to only happen when using !cv joinas and !callvote joinas and not every message


L-e-x-o-n:
Tested it and works well.
Thanks.

PR #687, opened 2025-06-23
fix_id: Allow shared player IDs for unranked games using shareId

aaronmorgan1989:
### Summary

This updates the fix_id logic to allow multiple players to have the same player_number in unranked games, provided they set matching shareId values.

### Implementation Details

- Updated `lib/teiserver/coordinator/consul_server.ex`:
 - Modified `fix_ids/1` to allow duplicate player_numbers for unranked games if shareId values match.
 - For ranked games, the original enforcement of unique player_numbers remains unchanged.

**How should player configs (shareId) be retrieved?**
- I thought it would be beneficial to use SPADS's current shareId functionality to enable player ID sharing as expected.
- The current approach attempts to fetch shareId from the user config via a DB call, **but the DB call causes testing to fail.**
- What is the best way to access shareId values for all players in the lobby? The shareId would by default be blank and then updated by the player in lobby before a game starts.
- Is there recommended method for syncing or listening for SPADS related config changes in Teiserver?

Overall thoughts on this method? Should I go a different route with this?
**Looking for guidance**

### DIFF
--- a/lib/teiserver/coordinator/consul_server.ex
+++ b/lib/teiserver/coordinator/consul_server.ex
@@ -1034,6 +1034,7 @@ defmodule Teiserver.Coordinator.ConsulServer do
   end
 
   # Ensure no two players have the same player_number
+  # unless the game is unranked and players have set matching shareId values
   defp fix_ids(state) do
     players = list_players(state)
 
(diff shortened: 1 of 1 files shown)


aaronmorgan1989:
Interesting, my local testing failed because of the DB call but it passed here. Still looking for input though

PR #690, opened 2025-06-25
Explore fix for 250 player balance memory consumption

NortySpock:
Just tooling around to see if I could reproduce the BalancerServer memory consumption issue and fix it using BlackMelon's suggestion.

I did not manage to reproduce the GC killing the process for being too big, so I could be very wide of the mark.

I don't feel like I have a strong understanding of how setting `GenServer.start_link(__MODULE__, opts[:data], max_heap_size: max_heap_size)` "works", I'm just copy-pasting code from Lexon's previous work.


lib/teiserver/game/servers/balancer_server.ex
```diff
@@ -11,7 +11,9 @@ defmodule Teiserver.Game.BalancerServer do
   @algos_allowing_fuzz ~w(loser_picks force_party)
   @spec start_link(List.t()) :: :ignore | {:error, any} | {:ok, pid}
   def start_link(opts) do
-    GenServer.start_link(__MODULE__, opts[:data], [])
+    # https://www.erlang.org/docs/23/man/erlang#process_flag_max_heap_size
+    max_heap_size = 50 * 1024 * 1024 / 8
+    GenServer.start_link(__MODULE__, opts[:data], max_heap_size: max_heap_size)
```
geekingfrog, line 16: the dialyzer warnings show that this is incorrect, there is no such argument `max_heap_size` for `start_link/3`. The doc you linked shows that it is another function entirely: `Process.flag/2` (the [elixir version](https://hexdocs.pm/elixir/Process.html#flag/2) points to the erlang one since it's the same).

test/teiserver/game/balancer_server_test.exs
```diff
@@ -174,4 +174,46 @@ defmodule Teiserver.Game.BalancerServerTest do
     poll_until_nil(fn -> GenServer.call(pid, :get_current_balance) end)
     assert Map.get(GenServer.call(pid, :report_state), :last_balance_hash, :not_found) == nil
   end
+
+  test "load 250 players in and run a balance pass or three, for perf testing" do
+    team_count = 2
+
+    players =
+      make_users_with_ranks(Enum.to_list(1..250))
```
geekingfrog, line 182: this isn't going to trigger the problematic behaviour, that test doesn't bring much value.
If you want to trigger the memory explosion, you need to have some groups of users in parties, and some without.

L-e-x-o-n:
> I don't feel like I have a strong understanding of how setting GenServer.start_link(__MODULE__, opts[:data], max_heap_size: max_heap_size) "works", I'm just copy-pasting code from Lexon's previous work.

It doesn't :(
Blackmelon's approach with `Process.flag` seems to be the correct way to do this. Working on this at the moment actually.

NortySpock:
Yeah, I'd say let's kill this PR in favor of Lexon's #692 , as I assume they know far more Elixir than I do.

If I get time I'll see if I can write a better working test case, just to understand how it all works under the hood.

PR #695, opened 2025-06-25
change curve algorihm for rating graph to curveCatmullRom to fix visual glitches

Kry-Vosa:
`curveCatmullRom` was chosen over `curveMonotoneX`.
`curveMonotoneX` give better looking curve, however it introduces a minor visual glitch where the "middle of uncertainty" skill line doesn't follow the edge of "bottom part of confidence interval" and leaves a little gap, see image.
If we wanna use the nicer looking `curveMonotoneX`, we could also omit the "middle of uncertainty" skill line, since it is barely visible anyway and the edge of the two areas accomplishes the same visual feat.

gap

### DIFF
--- a/lib/teiserver_web/templates/battle/match/ratings_graph.html.heex
+++ b/lib/teiserver_web/templates/battle/match/ratings_graph.html.heex
@@ -137,7 +137,7 @@
         .x(function(d) { return x(dateFormat(d.date)) })
         .y0(function(d) { return y(d.skill) })
         .y1(function(d) { return y(d.skill + d.uncertainty) })
-        .curve(d3.curveCardinal)
+        .curve(d3.curveMonotoneX)
         )
 
     // Show bottom part of confidence interval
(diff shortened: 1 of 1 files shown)


L-e-x-o-n:
What does `curveCatmullRom` look like?

NortySpock:
Not to open a can of worms, but:  Why a curve vs using a straight line segment, like a "normal" line chart?

Kry-Vosa:
`curveCatmullRom` would look like this:
image

NortySpock:
To me, the artificial curves looked wrong, like someone was trying to paint their own lines on a chart just to hide the fact that their OS sucked. ("See, right there, just for a moment I was totally above 17OS!") 

I prefer line segments, but then I am far too literal and matter-of-fact about things. Also I am unfun at parties, and am trending below 10 OS. 😛 🙃

L-e-x-o-n:
I prefer the flatness of `curveMonotoneX` and it undershooting less.
The uncertainty gap is not as bad and the line could be removed entirely as you said.
In any case, all of these options fix the very obvious visual glitches of the previous curve and are an improvement.

Kry-Vosa:
@NortySpock `curveMonotoneX` preserves local minima and maxima (which includes global minima and maxima), meaning the curve never overshoots (or undershoots) on the Y axis.

Kry-Vosa:
Updated the commit - switched to `curveMonotoneX`, removed the "middle of uncertainty" line. This is how it looks with in the updated pull request: 
image

PR #699, opened 2025-06-29
Os in matchmaking

geekingfrog:
Switch matchmaking to use a basic matchmaking algorithm that uses OS and the win prediction to filter out unsatisfactory matches. It is rather basic with some problems, but the change is getting big enough so far, and it's now in a good place to stop.

I've added a [livebook](https://livebook.dev/) to simulate the matching process and visualise how that works.
The main issue is that for 1v1, players outside the mean (something around `2.5σ`) will basically never be matched.
2025-06-29-817x273-scrot

This is much better with teams since it's easier to balance:
2025-06-29-860x263-scrot

A further improvement would be to adjust the algorithm on each player's OS: be more lenient for the highest and lowest OS players.
Another improvement is to be more lenient for player the longer they are in the queue.

### DIFF
--- a/matchmaking-os.livemd
+++ b/matchmaking-os.livemd
@@ -1,0 +1,183 @@
+# matchmaking with OS
+
+```elixir
+Mix.install([
+  {:vega_lite, "~> 0.1.6"},
+  {:kino_vega_lite, "~> 0.1.11"},
+  {:openskill, git: "https://github.com/beyond-all-reason/openskill.ex.git", branch: "master"},
+  {:elixir_uuid, "~> 1.2"},
+  {:teiserver, path: __DIR__}
+])
+```
+
+## Parameters
+
+```elixir
+alias VegaLite, as: Vl
+alias Teiserver.Matchmaking.{Algo, Member}
+```
+
+```elixir
+# how many tick to run the simulation for
+number_of_ticks = 400
+
+# match parameters
+team_size = 1
+team_count = 2
+
+# See https://github.com/beyond-all-reason/teiserver/issues/679
+# to get the "correct" value for the normal distribution of rating
+{mu, sigma} = {18.964, 8.484}
+
+
+# which algo to use for matchmaking
+alg_module = Algo.BruteforceFilter
+```
+
+```elixir
+gen_rating = fn() ->
+  rating = Statistics.Distributions.Normal.rand(mu, sigma)
+  # completely arbitrary
+  uncertainty = :rand.uniform() * (7.0 - 2.0) + 2.0
+  %{rating: rating, uncertainty: uncertainty}
+end
+
+gen_player = fn(p_id) ->
+  %Member{
+    id: UUID.uuid4(),
+    player_ids: [p_id],
+    rating: gen_rating.(),
+    joined_at: DateTime.utc_now()
+  }
+end
+
+game_type = Teiserver.Battle.MatchLib.game_type(team_size, team_count)
+
+# eyeballing the graphs of matches, there are roughly, per day
+# 650 duels
+# 500 small team game
+# 950 large team
+# 110 ffa
+# it's probably not evenly distributed, so let's say this happens over
+# 20h (arbitrary).
+# the arrival time is then very roughly, per minute:
+# 1.08 for duels
+# 3.3 for small team (assuming 4 players per team)
+# 0.9 for ffa (assuming 10 players per match)
+# 12.6 for large team (16 players per match)
+
+# expectation that λ events happen during the time interval
+# here, it's the number of player joining the queue between two ticks
+# Assuming one tick every 30 seconds.
+lambda = 1.5
+
+id_gen = Stream.iterate(1000, &(&1 + 1))
+
+# This is a stream where each element is an array of %Member that joined during this tick.
+members_per_tick =
+  Stream.unfold({0, id_gen}, fn {tick, id_gen} ->
+    # number of player arriving this tick, it's always an integer but
+    # rand() returns a float
+    n = trunc(Statistics.Distributions.Poisson.rand(lambda))
+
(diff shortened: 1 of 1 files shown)


lib/teiserver/matchmaking/algos.ex
```diff
@@ -1,35 +1,42 @@
...
+          team_size :: pos_integer(),
+          team_count :: pos_integer(),
+          pred :: ([[Member.t()]] -> boolean),
+          acc :: term()
+        ) :: [[[Member.t()]]]
+  def match_members(members, team_size, team_count, pred, acc \\ []) do
     res =
-      match_stream(team_size, team_count, members)
+      match_stream(members, team_size, team_count)
+      |> Stream.filter(pred)
```
geekingfrog, line 39: the `take(1)` could be improved a bit as well later. We could `take` more and then get the best according to some metric, but that's future improvements.

matchmaking-os.livemd
geekingfrog, line 1: This is the livebook, it's better to see it with a livebook environment than just the code.

lib/teiserver/account.ex
```diff
@@ -942,6 +942,12 @@ defmodule Teiserver.Account do
     end)
   end
 
+  # def get_ratings(user_id, season \\ nil) do
+  #   season = if season == nil, do: MatchRatingLib.active_season(), else: season
+  #   q = rating_query(search: [user_id: user_id, season: season])
+  #   Teiserver.Game.get_ratings_for_user(preload: [ratings: q])
+  # end
+
```
L-e-x-o-n: What's this for?

lib/teiserver/battle/libs/match_lib.ex
```diff
@@ -60,6 +60,19 @@ defmodule Teiserver.Battle.MatchLib do
...
+  @spec game_type_for_team(team_size :: pos_integer(), team_count :: pos_integer()) :: binary()
+  def game_type_for_team(team_size, team_count \\ 2) do
+    cond do
+      team_count > 2 -> if team_size == 1, do: "FFA", else: "Team FFA"
+      team_size <= 1 -> "Duel"
+      team_size <= 5 -> "Small Team"
+      true -> "Large Team"
+    end
+  end
+
```
L-e-x-o-n: Please use `Teiserver.Battle.MatchLib.game_type/2`.

lib/teiserver/matchmaking/member.ex
```diff
@@ -21,14 +29,49 @@ defmodule Teiserver.Matchmaking.Member do
...
+        ratings = rt.ratings
+        n = Enum.count(ratings)
+        rating = Enum.sum_by(ratings, & &1.rating_value)
+        uncertainty = Enum.sum_by(ratings, & &1.uncertainty)
+
+        if n == 0,
+          do: default,
+          else: %{rating: rating / n, uncertainty: uncertainty / n}
+    end
+  end
```
L-e-x-o-n: `Openskill.predict_win` expects skill, uncertainty pairs, not rating uncertainty. (`rating_value` is `max(skill - uncertainty, 0)`).

lib/teiserver/matchmaking/algo/bruteforce_filter.ex
```diff
@@ -0,0 +1,56 @@
...
+  end
+
+  defp filter_within_bounds(match) do
+    team_skills =
+      Enum.map(match, fn team ->
+        Enum.map(team, fn m ->
+          %{rating: rating, uncertainty: uncertainty} = m.rating
+          {rating, uncertainty}
+        end)
+      end)
```
L-e-x-o-n: `Openskill.predict_win` expects skill, uncertainty pairs, not rating uncertainty. (`rating_value` is `max(skill - uncertainty, 0)`).

lib/teiserver/game.ex
```diff
@@ -375,6 +375,24 @@ defmodule Teiserver.Game do
...
+  @spec get_ratings_for_users(user_ids :: [integer()]) :: [RatingType.t()]
+  def get_ratings_for_users(user_ids),
+    do: get_ratings_for_users(user_ids, MatchRatingLib.active_season())
+
+  @spec get_ratings_for_users(user_ids :: [integer()], season :: integer()) :: [RatingType.t()]
+  def get_ratings_for_users(user_ids, season) do
+    query = Teiserver.Account.rating_query(search: [user_id_in: user_ids, season: season])
+
+    rating_type_query(preload: [ratings: query])
+    |> Repo.all()
```
L-e-x-o-n, line 393: Is this necessary? Preloading and returning all ratings this way only to later later filter them by `game_type`?
Exising `Account.list_ratings` used should work here as well:
```ex
Account.list_ratings(
  search: [
    rating_type_id: rating_type_id,
    user_id_in: user_ids,
    season: season
  ]
)
```
You can get `rating_type_id` from `game_type` / name like this:
```ex
rating_type_id = get_or_add_rating_type(.game_type)
```
(already in the same module, `Teiserver.Game`)

lib/teiserver/account.ex
geekingfrog: Forgot to remove that while I was figuring stuff out

lib/teiserver/game.ex
geekingfrog: I wanted to have this function because it's going to be used for lobbies, where we'll want to know all possible ratings upfront, so that switching game type doesn't require additional back & forth to get missing data.

geekingfrog:
> Not a major issue, but I don't agree with the pluralisation in [2a7e7b7](https://github.com/beyond-all-reason/teiserver/commit/2a7e7b79d827b1a2c5de6db0cdd5875085071bef). It is still holds one pair of numbers representing the rating of that member, I don't see why it should be pluralised, imo ratings suggests it's a list.

Yeah, you're right, I reworked the commit to mention that instead it holds only the rating pair for the gametype of the queue.

PR #705, opened 2025-07-18
ADD s.battle.teams & s.battle.teams_update custom messages

wrath625:
First attempt at adding a custom message sent from teiserver to CHOBBY to inform of teamSize and nbTeams before loading into a specific lobby.

this will allow for the lobby list to display the actual lobby sizes instead of /16 every time.

When a user logs in they receive a single bulk message of all all lobbyids and their respective teamSize and nbTeams values in base64 encoded json

when :update is called on a lobby it also triggers a s.battle.teams_update message for that specific lobby sending teamSize and nbTeams

There is CHOBBY code implementing this here:

https://github.com/beyond-all-reason/BYAR-Chobby/pull/1029


wrath625:
this relates to this thread:

https://discord.com/channels/549281623154229250/1395735550698328084

lib/teiserver/protocols/spring/spring_out.ex
```diff
@@ -737,6 +737,17 @@ defmodule Teiserver.Protocols.SpringOut do
       end
     end)
 
+    # Send BATTLETEAMS batch message
+    teams_data = 
+      Lobby.list_lobby_ids()
+      |> Enum.map(fn lobby_id ->
+        consul_state = Teiserver.Coordinator.call_consul(lobby_id, :get_all)
```
geekingfrog: This is going to do a call to every single lobby and I'm worried about the performance implications of that.
There are 2 issues:

1) this is going to be a bit slow when the number of lobbies is high. You can mitigate that by doing these call in tasks and then use [Task.await_many/2](https://hexdocs.pm/elixir/Task.html#await_many/2)

2) the consul genservers are going to get flooded by these requests since these are done by each connected players, only to return the same thing. One quick win is to return only what you need instead of the full lobby state. Create a new message and handler for the teamsize and teamcount, that'll limit the amount of bytes copied.

As for the thundering herd effect, I am unsure what's the best approach. You could cache these calls, but then, you'll have to deal with invalidation of said cache. The easy way is to set a very low ttl, like 1s, so it can be stale, but only for a short while, and 1s should cover all the clients asking for the same thing. You can either cache per lobby, or for all lobbies at once. I think I have a preference for all lobbies.

Another option, would be to invalidate the cache in [consul_server](https://github.com/geekingfrog/teiserver/blob/b5e55c49a56047e468352a261a139f89be5d548e/lib/teiserver/coordinator/consul_server.ex) itself when the teamsize or teamcount changes. There seems to be some central logic about [player count change](https://github.com/geekingfrog/teiserver/blob/b5e55c49a56047e468352a261a139f89be5d548e/lib/teiserver/coordinator/consul_server.ex#L1108). Probably good to have a look at this file and see how these properties you're interested in are managed, maybe it's not much work to invalidate a cache.

geekingfrog (changes requested):
I've tested it locally and it works but I'm worried about performances.
I left some comments about one point.

I would also like you to put this behind a feature flag. We already have a bunch of "site configs", and I think this could be under [lobby_configs](https://github.com/geekingfrog/teiserver/blob/b5e55c49a56047e468352a261a139f89be5d548e/lib/teiserver/libs/teiserver_configs.ex#L313).
This way, if things go haywire, we can swiftly turn off the codepath without having a full redeploy (that kills all lobbies).

lib/teiserver/protocols/spring/spring_out.ex
wrath625: Hey thanks for the comments.

I've added a feature flag, optimized and parallelized the consul requests on user auth, and added a cache with a 1s TTL as you suggested.

I opted to avoid the cache invalidation route as I'm not confident I would be able to find all the edge cases given my familiarity with the project so far.

lib/teiserver/protocols/spring/spring_out.ex
```diff
@@ -737,6 +737,34 @@ defmodule Teiserver.Protocols.SpringOut do
       end
     end)
 
+    # Send BATTLETEAMS batch message
+    if Config.get_site_config_cache("lobby.Broadcast Battle Teams Information") do
+      ttl_ms = 1_000
+      now = System.system_time(:millisecond)
```
geekingfrog: Prefer using [monotonic_time](https://www.erlang.org/doc/apps/erts/erlang#monotonic_time/0) since this is only going to be used for time deltas.

lib/teiserver/protocols/spring/spring_out.ex
```diff
@@ -737,6 +737,34 @@ defmodule Teiserver.Protocols.SpringOut do
...
+                timeout: 2_000
+              )
+              |> Enum.map(fn {:ok, result} -> result end)
+              |> Map.new()
+            Teiserver.cache_put(:application_temp_cache, :battle_teams, {data, now})
+            data
+        end
+
+      send(self(), {:battle_teams, teams_data})
+    end
```
geekingfrog, line 745: In term of architecture and separation of concern it's not great to have this block of code inside `spring_out`. This file is responsible to translate some data from an internal format to whatever spring requires. Bundling logic to fetch and cache this data break this abstraction boundary.

I would move the entire content of the if block into `teiserver/lobby/libs/lobby_lib.ex`, and expose that in `Teiserver.Battle`, like it's done for other functions.

lib/teiserver/tcp/spring/spring_tcp_server.ex
```diff
@@ -409,6 +409,17 @@ defmodule Teiserver.SpringTcpServer do
         end
       end
 
+    # Check for team size/count changes
+    new_state = 
+      if Config.get_site_config_cache("lobby.Broadcast Battle Teams Information") and
+         (Map.has_key?(new_values, :host_teamsize) or Map.has_key?(new_values, :host_teamcount)) do
+        consul_state = Teiserver.Coordinator.call_consul(lobby_id, :get_all)
```
geekingfrog: why not use the new `:get_team_config` message?

lib/teiserver/coordinator/consul_server.ex
```diff
@@ -82,6 +82,10 @@ defmodule Teiserver.Coordinator.ConsulServer do
     {:reply, result, state}
   end
 
+  def handle_call(:get_team_config, _from, state) do
+    {:reply, %{host_teamsize: state.host_teamsize, host_teamcount: state.host_teamcount}, state}
+  end
```
geekingfrog, line 87: One comment here, but don't feel forced to change the code, is to create an API to hide away the actual messages. For example, in [the newer code](https://github.com/beyond-all-reason/teiserver/blob/master/lib/teiserver/matchmaking/queue_server.ex#L150-L156), there's [join_queue](https://github.com/beyond-all-reason/teiserver/blob/master/lib/teiserver/matchmaking.ex#L34-L37) that merely sends the correct message to the correct genserver. This way, if one needs to modify the message, it's only an internal refactoring and caller to the API don't need to care.

Though I appreciate that this style of coding isn't the norm for the spring related bits of teiserver, so feel free to ignore this suggestion and keep it consistent with the rest of the module.

lib/teiserver/protocols/spring/spring_out.ex
```diff
@@ -737,6 +737,34 @@ defmodule Teiserver.Protocols.SpringOut do
...
+      ttl_ms = 1_000
+      now = System.system_time(:millisecond)
+      teams_data =
+        case Teiserver.cache_get(:application_temp_cache, :battle_teams) do
+          {cached, ts} when now - ts < ttl_ms ->
+            cached
+          _ ->
+            data =
+              Lobby.list_lobby_ids()
+              |> Task.async_stream(
```
geekingfrog: oh, didn't know about `async_stream`. In this case, I think I prefer `await_many`, since you're going to send messages to different genserver anyway, the scheduler will take care of the rest, and you don't have to come up with some arbitrary `max_concurrency` number.

wrath625: sounds good, changing this to await_many

lib/teiserver/coordinator/consul_server.ex
wrath625: I can definitely do this, I would prefer to do it clean where possible!

lib/teiserver/tcp/spring/spring_tcp_server.ex
wrath625: whoops!

lib/teiserver/protocols/spring/spring_out.ex
wrath625: totally makes sense, moving this

lib/teiserver/protocols/spring/spring_out.ex
wrath625: easy fix

geekingfrog:
Tested and works (though need the patch I submitted on chobby for it to actually work). Server side seems fine, we can turn it on and off.

PR #707, opened 2025-07-24
Added reaction to "Spring crashed !"

### LINKED ISSUE #706: [Bug]: Balance frozen after spring crash
### Describe the Bug

This bug is observerd for at least about 2 years now on hosts with ffa preset teamsize 1.
The source may not be teiserver itself, could also be sth. in the communication between spads and teiserver.

After a game is started and spring/recoil crashes on server side, it won't balance the users anymore until next regular game has finished.

From the chobby view it behaves exactly the same way as if the game was running:
- If another user joins the playerbase, he will be forced to ally team 1
- If this user tries to change its ally nb., it gets forced back to ally team 1
- !balance doesn't work

The community solves this situation by voting a boss, setting balancemode off, switching each user manually to a different ally team nb. and then start the next game. This works always.

### Reproduce the bug

We could reproduce the bug on test server5 env with Host[EU4][000].

1. Setup a host with ffa preset
2. Join some players
3. Start the game
4. Kill the specific spring/recoil process on server side (done by Beherith in the test)

To get the correct spring process:
`ps aux | grep 'recoil_2025.04.08_amd64-linux/spring' | grep 'Host[EU4][000]'`
and then kill with
`kill xxxx`

I did the same with my local teiserver+spads-cluster (WSL-Ubuntu) **and the bug didn't happen.**
Either it's a different configuration or it's sth. specific to the BAR environment.

In server4 it happens always, when the host spring process crashes.

### Screenshots

Image
Image

Image

Image

Image

Image

Image

Image

Image

Image

### Additional context

[infolog_Fireball_Test3.txt](https://github.com/user-attachments/files/21394073/infolog_Fireball_Test3.txt)

FIr3baL:
Resolves #706

### DIFF
--- a/lib/teiserver/battle/servers/match_monitor_server.ex
+++ b/lib/teiserver/battle/servers/match_monitor_server.ex
@@ -88,6 +88,20 @@ defmodule Teiserver.Battle.MatchMonitorServer do
     {:noreply, state}
   end
 
+  # Spring crashed
+  def handle_info(
+        {:new_message, from_id, "autohosts", "* Spring crashed ! (running time" <> rest},
+        state
+      ) do
+    client = Client.get_client_by_id(from_id)
+    Battle.stop_match(client.lobby_id)
+
+    match_id = Battle.get_match_id_from_userid(from_id)
+    Telemetry.log_simple_lobby_event(nil, match_id, "lobby.spring_crashed")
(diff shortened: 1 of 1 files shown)


L-e-x-o-n:
Thanks

PR #708, opened 2025-07-25
Friend Request code cleanup

### LINKED ISSUE #622: [Feature]: tighter create requests
### Describe the Feature

There are 2 issues with [create_friend_request](https://github.com/beyond-all-reason/teiserver/blob/master/lib/teiserver/account.ex#L1857):

1. it doesn't validate anything using [can_send_friend_request_with_reason?](https://github.com/beyond-all-reason/teiserver/blob/master/lib/teiserver/account/libs/friend_request_lib.ex#L22)
2. it doesn't check that there isn't already a pending friend request, either with the same originator, or from the other direction

The function should be refactored and the callsites changed so that it returns an error when the request can't be created. This check must be done in a transaction to avoid any TOCTOU issue.

### Expected Behavior

_No response_

### Screenshots

_No response_

### Additional context

_No response_

wrath625:
This code tightens up the friend request creation code, centralizing validation into account.ex, and adding transactionality(is that a word?) to the creation of friend requests.

It standardizes error response handling, which is also compatible with tachyon_handler.ex's handling of create_friend_request

Closes #622

### DIFF
--- a/lib/teiserver/account.ex
+++ b/lib/teiserver/account.ex
@@ -1907,56 +1907,53 @@ defmodule Teiserver.Account do
 
         existing_request ->
           # Auto-accept the existing request
-          result = FriendRequestLib.accept_friend_request(existing_request)
+          with :ok <- FriendRequestLib.accept_friend_request(existing_request) do
+            # Send pubsub broadcasts for both users to update UI
+            PubSub.broadcast(
+              Teiserver.PubSub,
+              "account_user_relationships:#{existing_request.from_user_id}",
+              %{
+                channel: "account_user_relationships:#{existing_request.from_user_id}",
+                event: :friend_request_accepted,
+                userid: existing_request.from_user_id,
+                accepter_id: existing_request.to_user_id
+              }
+            )
 
-          case result do
-            :ok ->
-              # Send pubsub broadcasts for both users to update UI
-              PubSub.broadcast(
-                Teiserver.PubSub,
-                "account_user_relationships:#{existing_request.from_user_id}",
-                %{
-                  channel: "account_user_relationships:#{existing_request.from_user_id}",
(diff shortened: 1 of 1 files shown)


lib/teiserver/account.ex
```diff
@@ -1887,42 +1887,73 @@ defmodule Teiserver.Account do
...
-
-    result
+    # NOTE: All call sites must now handle {:error, reason} returns from create_friend_request/2
+    from_user_id = Map.get(attrs, :from_user_id)
+    to_user_id = Map.get(attrs, :to_user_id)
+    create_friend_request(from_user_id, to_user_id)
   end
 
   def create_friend_request(from_user_id, to_user_id) do
     if from_user_id == to_user_id do
```
L-e-x-o-n: Maybe this check would be better inside `can_send_friend_request_with_reason?`?

lib/teiserver/account.ex
```diff
@@ -1887,42 +1887,73 @@ defmodule Teiserver.Account do
...
+            Repo.one(
+              from fr in FriendRequest,
+                where:
+                  (fr.from_user_id == ^from_user_id and fr.to_user_id == ^to_user_id) or
+                    (fr.from_user_id == ^to_user_id and fr.to_user_id == ^from_user_id)
+            )
+
+          if existing do
+            Repo.rollback("Pending request exists")
+          else
```
L-e-x-o-n: `can_send_friend_request_with_reason?` already checks if an outgoing friend request exists.

Additionally, it would be nice if an incoming friend request could be auto accepted if you are trying to send a new request to the user that sent you a friend request.
So for example,
- User A sends friend request to user B
- User B attempts to send friend request to user A
- User A's friend request is accepted

PR #709, opened 2025-07-25
Extend admin api to create accounts via REST

### LINKED ISSUE #676: [Feature]: Extend admin api to create accounts
### Describe the Feature

For load testing at least, it's going to be useful to have a way to generate users in a specific state in a programmatic manner.

### Expected Behavior

With a token with the scope `admin.user` a POST request to `/teiserver/api/admin/users` should be able to create a user with a specific:
* email
* name
* skill (mu and sigma)
* play time + spec time + lobby time

The result should be in json format and also contain credentials: an access and refresh token. The access token may be set to expire later than usual maybe so that load testing tooling doesn't need to care about the oauth flow.

Optionally there could also be another endpoint to get an existing user with a given email + fresh credentials. This would allow us to reuse test accounts when doing load testing repeatedly.

wrath625:
Added POST request to /teiserver/api/admin/users to create users with specified payload:

- email
- name
- skill (mu and sigma)
- play time + spec time + lobby time

Added GET request to /teiserver/api/admin/users/:email to retrieve a user and generate new auth keys

Added a new Oauth application scoped to admin.user, created when you run `mix teiserver.tachyon_setup`

Closes #676

Testing can be done as follows:

1. Re-run `mix teiserver.tachyon_setup` to create the new admin_user oauth application

2. Get an oauth token

3. Create new user:
```
curl -X POST http://localhost:4000/teiserver/api/admin/users \
  -H "Authorization: Bearer YOUR_OAUTH_TOKEN" \
  -H "Content-Type: application/json" \
  -d '{
    "email": "test@example.com",
    "name": "TestUser",
    "skill": {"mu": 1500, "sigma": 500},
    "play_time": 3600,
    "spec_time": 1800,
    "lobby_time": 900
  }'
```

4. Get new auth key for existing user:
```
curl -X GET http://localhost:4000/teiserver/api/admin/users/test@example.com \
  -H "Authorization: Bearer YOUR_OAUTH_TOKEN" \
  -v
```

### DIFF
--- a/lib/teiserver_web/controllers/api/admin/user_controller.ex
+++ b/lib/teiserver_web/controllers/api/admin/user_controller.ex
@@ -61,7 +61,7 @@ defmodule TeiserverWeb.API.Admin.UserController do
   end
 
   defp update_stats_if_needed(user_id, stat_fields) do
-    if Map.size(stat_fields) > 0 do
+    if map_size(stat_fields) > 0 do
       Account.update_user_stat(user_id, stat_fields)
     else
       {:ok, nil}
@@ -84,23 +84,17 @@ defmodule TeiserverWeb.API.Admin.UserController do
     end
   end
 
-  @dialyzer {:no_return, create_user_token: 2}
   defp create_user_token(user, app) do
-    with {:ok, token} <-
-           OAuth.create_token(
-             user,
-             %{
-               id: app.id,
-               scopes: app.scopes
-             },
-             create_refresh: true,
-             scopes: app.scopes
-           ) do
-      {:ok, token}
-    end
+    OAuth.create_token(
+      user,
+      %{
+        id: app.id,
+        scopes: app.scopes
+      },
+      create_refresh: true
+    )
   end
 
-  @dialyzer {:nowarn_function, build_user_response: 2}
   defp build_user_response(user, token) do
     credentials = %{
       access_token: token.value,
@@ -121,7 +115,7 @@ defmodule TeiserverWeb.API.Admin.UserController do
     %{user: Map.merge(user_data, timestamps), credentials: credentials}
   end
 
-  defp handle_error(conn, {:error, changeset}) when is_struct(changeset, Ecto.Changeset) do
+  defp handle_error(conn, {:error, %Ecto.Changeset{} = changeset}) do
     conn |> put_status(400) |> json(%{error: format_changeset_errors(changeset)})
   end
 
--- a/test/teiserver_web/controllers/api/admin/user_controller_test.exs
+++ b/test/teiserver_web/controllers/api/admin/user_controller_test.exs
@@ -76,6 +76,13 @@ defmodule TeiserverWeb.API.Admin.UserControllerTest do
 
       assert resp["user"]["name"] == "testuser2"
       assert resp["credentials"]["access_token"]
+
+      # Verify the created user has the correct stats
+      user = Teiserver.Account.get_user_by_email("testuser2@example.com")
+      user_stats = Teiserver.Account.get_user_stat_data(user.id)
+      assert user_stats["mu"] == 1500
+      assert user_stats["sigma"] == 100
+      assert user_stats["play_time"] == 3600
     end
 
     test "handles missing required fields", %{authed_conn: conn} do
@@ -99,18 +106,6 @@ defmodule TeiserverWeb.API.Admin.UserControllerTest do
       assert resp["error"] =~ "invalid email"
     end
 
-    test "accepts valid email with plus sign", %{authed_conn: conn} do
-      user_data = %{
-        "name" => "testuser",
-        "email" => "test+123@example.com",
-        "password" => "testpassword123"
-      }
-
-      resp = conn |> post(create_user_path(), user_data) |> json_response(200)
-      assert resp["user"]["email"] == "test+123@example.com"
-      assert resp["credentials"]["access_token"]
-    end
-
     test "handles duplicate email", %{authed_conn: conn} do
       # First create a user
       user_data = %{
@@ -130,49 +125,6 @@ defmodule TeiserverWeb.API.Admin.UserControllerTest do
 
       resp = conn |> post(create_user_path(), duplicate_data) |> json_response(400)
       assert resp["error"] =~ "Email already attached to a user"
-    end
-
-    test "handles empty email", %{authed_conn: conn} do
-      user_data = %{
-        "name" => "testuser",
-        "email" => "",
-        "password" => "testpassword123"
-      }
-
-      resp = conn |> post(create_user_path(), user_data) |> json_response(400)
-      assert resp["error"] =~ "can't be blank"
-    end
-
-    test "handles nil email", %{authed_conn: conn} do
-      user_data = %{
-        "name" => "testuser",
-        "password" => "testpassword123"
-      }
-
-      resp = conn |> post(create_user_path(), user_data) |> json_response(400)
-      assert resp["error"] =~ "can't be blank"
-    end
-
-    test "handles email with only @", %{authed_conn: conn} do
-      user_data = %{
-        "name" => "testuser",
-        "email" => "test@",
-        "password" => "testpassword123"
-      }
-
-      resp = conn |> post(create_user_path(), user_data) |> json_response(400)
-      assert resp["error"] =~ "invalid email"
-    end
-
-    test "handles email with only .", %{authed_conn: conn} do
-      user_data = %{
-        "name" => "testuser",
-        "email" => "test.com",
-        "password" => "testpassword123"
-      }
-
-      resp = conn |> post(create_user_path(), user_data) |> json_response(400)
-      assert resp["error"] =~ "invalid email"
     end
 
     test "creates user with all optional fields", %{authed_conn: conn} do
@@ -249,25 +201,6 @@ defmodule TeiserverWeb.API.Admin.UserControllerTest do
       refresh_data = %{"email" => ""}
       resp = conn |> post(refresh_token_path(), refresh_data) |> json_response(404)
       assert resp["error"] == "User not found"
-    end
-
-    test "handles email with plus sign", %{authed_conn: conn} do
-      # First create a user with plus sign
-      user_data = %{
-        "name" => "plususer",
-        "email" => "user+test@example.com",
-        "password" => "testpassword123"
-      }
-
-      conn |> post(create_user_path(), user_data) |> json_response(200)
-
-      # Then refresh their token
-      refresh_data = %{"email" => "user+test@example.com"}
-      resp = conn |> post(refresh_token_path(), refresh_data) |> json_response(200)
-
-      assert resp["user"]["name"] == "plususer"
-      assert resp["user"]["email"] == "user+test@example.com"
-      assert resp["credentials"]["access_token"]
     end
   end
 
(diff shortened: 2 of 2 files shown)


lib/teiserver_web/controllers/api/admin/user_controller.ex
```diff
@@ -0,0 +1,102 @@
+defmodule TeiserverWeb.API.Admin.UserController do
+  use TeiserverWeb, :controller
+  alias Teiserver.{Account, OAuth}
+
+  plug Teiserver.OAuth.Plug.EnsureAuthenticated, scopes: ["admin.user"]
+
+  @stat_fields ["mu", "sigma", "play_time", "spec_time", "lobby_time"]
+
+  # POST /teiserver/api/admin/users
```
geekingfrog: Not the biggest fan of these comments, as they are bound to fall out of sync. Usually, it's easy enough to grep the module name in `router.ex`, or run `mix phx.routes` to have this info.

lib/teiserver_web/controllers/api/admin/user_controller.ex
```diff
@@ -0,0 +1,152 @@
...
+          "data" => Account.User.default_data()
+        },
+        params
+      )
+
+    case Account.script_create_user(params) do
+      {:ok, user} ->
+        update_user_stats(user.id, params)
+
+        case OAuth.get_application_by_uid("user_admin") do
```
geekingfrog: This is not the correct application. Creating a token bound to this app means it will have permissions like `admin.user`, which is definitely not what we want.
You need to grab the `generic_lobby` app so that the token can be used with the new client and regular tachyon operations.

lib/teiserver_web/controllers/api/admin/user_controller.ex
```diff
@@ -0,0 +1,152 @@
...
+            |> put_status(400)
+            |> json(%{
+              error:
+                "user_admin OAuth application not found. Please run 'mix teiserver.tachyon_setup' to create it."
+            })
+        end
+
+      {:error, changeset} ->
+        conn |> put_status(400) |> json(%{error: format_changeset_errors(changeset)})
+    end
```
geekingfrog, line 40: To avoid many level of nesting, you can use the [with](https://hexdocs.pm/elixir/Kernel.SpecialForms.html#with/1) construct.

lib/teiserver_web/controllers/api/admin/user_controller.ex
```diff
@@ -0,0 +1,152 @@
...
+                     id: app.id,
+                     scopes: app.scopes
+                   },
+                   create_refresh: true,
+                   scopes: app.scopes
+                 ) do
+              {:ok, token} ->
+                credentials = %{access_token: token.value}
+
+                if token.refresh_token,
```
geekingfrog: I don't think you need the `if` there since you're explicitly asking to create a refresh token 4 lines above.

lib/teiserver_web/controllers/api/admin/user_controller.ex
```diff
@@ -0,0 +1,152 @@
...
+                "user_admin OAuth application not found. Please run 'mix teiserver.tachyon_setup' to create it."
+            })
+        end
+
+      {:error, changeset} ->
+        conn |> put_status(400) |> json(%{error: format_changeset_errors(changeset)})
+    end
+  end
+
+  # GET /teiserver/api/admin/users/:email
```
geekingfrog: similar comment

lib/teiserver_web/controllers/api/admin/user_controller.ex
```diff
@@ -0,0 +1,152 @@
...
+  def show(conn, %{"email" => email}) do
+    case Account.get_user_by_email(email) do
+      nil ->
+        conn |> put_status(404) |> json(%{error: "User not found"})
+
+      user ->
+        case OAuth.get_application_by_uid("user_admin") do
```
geekingfrog: Two comments on this block:
* there is **a lot** of duplication to create the token and then convert the result into the correct json form, that should be extracted into a `defp`
* I don't like mutating `GET` requests. In the http world, `GET` requests are assumed to be cacheable, and often idempotent. This is neither. I would remove the `GET` and instead do a `POST` on an endpoint like `.../refresh_token` or something like that.

lib/teiserver_web/controllers/api/admin/user_controller.ex
```diff
@@ -0,0 +1,152 @@
...
+        end
+    end
+  end
+
+  # Private helpers
+
+  defp update_user_stats(user_id, params) do
+    stat_fields = Map.take(params, @stat_fields)
+    if stat_fields != %{}, do: Account.update_user_stat(user_id, stat_fields)
+    :ok
```
geekingfrog: you shouldn't throw away the result of `update_user_stat`.

geekingfrog (commented):
This is going to be quite useful thanks.
I left a few minor comments.

But more importantly, this need tests. You can have a look for example at the [asset api tests](https://github.com/geekingfrog/teiserver/blob/ce0acea486c53304df8f82df54a821ea031d3770/test/teiserver_web/controllers/api/admin/asset_controller_test.exs) for inspiration.

The spring related code is undertested, but since it's getting replaced, I'm not too concerned about it. This code however is brand new for upcoming feature, and I want reasonable test coverage.

lib/teiserver_web/controllers/api/admin/user_controller.ex
wrath625: I've reverted this back to `generic_lobby`, however I'm not sure I'm understanding how we want this set up.  The original issue wanted this scoped to `admin.user`, did you change your mind or am I not understanding how this works?

NortySpock:
Worked in my testing, great to see some tests in there too. 

I'm no expert, just a mere occasional contributor, but it looks good to me :)

test/teiserver_web/controllers/api/admin/user_controller_test.exs
```diff
@@ -0,0 +1,405 @@
...
+        "password" => "testpassword123",
+        "mu" => 1500,
+        "sigma" => 100,
+        "play_time" => 3600
+      }
+
+      resp = conn |> post(create_user_path(), user_data) |> json_response(200)
+
+      assert resp["user"]["name"] == "testuser2"
+      assert resp["credentials"]["access_token"]
```
geekingfrog, line 78: You should also test that the created user has indeed the correct stats.

test/teiserver_web/controllers/api/admin/user_controller_test.exs
```diff
@@ -0,0 +1,405 @@
...
+        "name" => "testuser",
+        "email" => "not-an-email-at-all",
+        "password" => "testpassword123"
+      }
+
+      resp = conn |> post(create_user_path(), user_data) |> json_response(400)
+      assert resp["error"] =~ "invalid email"
+    end
+
+    test "accepts valid email with plus sign", %{authed_conn: conn} do
```
geekingfrog: ooooh, these are a rarity, but it's nice to see all of that works.

test/teiserver_web/controllers/api/admin/user_controller_test.exs
```diff
@@ -0,0 +1,405 @@
...
+    test "handles email with only .", %{authed_conn: conn} do
+      user_data = %{
+        "name" => "testuser",
+        "email" => "test.com",
+        "password" => "testpassword123"
+      }
+
+      resp = conn |> post(create_user_path(), user_data) |> json_response(400)
+      assert resp["error"] =~ "invalid email"
+    end
```
geekingfrog: I would remove all the tests around email validation but one. Like all the ones around email format. Because you are merely using an existing `valid_email?` function, that you can assume [has been tested elsewhere](https://github.com/geekingfrog/teiserver/blob/6d7618f0312a5ec32574e66c940bf0ebe8ef1931/test/teiserver/data/user_test.exs#L113-L126), though it seems it's lacking a few cases.

maybe keep one and the duplicate email test, the other should go.

test/teiserver_web/controllers/api/admin/user_controller_test.exs
```diff
@@ -0,0 +1,405 @@
...
+      conn |> post(create_user_path(), user_data) |> json_response(200)
+
+      # Then refresh their token
+      refresh_data = %{"email" => "user+test@example.com"}
+      resp = conn |> post(refresh_token_path(), refresh_data) |> json_response(200)
+
+      assert resp["user"]["name"] == "plususer"
+      assert resp["user"]["email"] == "user+test@example.com"
+      assert resp["credentials"]["access_token"]
+    end
```
geekingfrog: similar, I don't think this test is necessary.

test/teiserver_web/controllers/api/admin/user_controller_test.exs
```diff
@@ -0,0 +1,405 @@
...
+        "email" => "negative@example.com",
+        "password" => "testpassword123",
+        "mu" => -100,
+        "sigma" => -50,
+        "play_time" => -3600
+      }
+
+      resp = conn |> post(create_user_path(), user_data) |> json_response(200)
+      assert resp["user"]["name"] == "negativeuser"
+    end
```
geekingfrog: why did you add these 2 tests? They don't seem to be very useful to me but maybe I'm missing something.

lib/teiserver_web/controllers/api/admin/user_controller.ex
```diff
@@ -0,0 +1,139 @@
...
+
+  defp get_user_by_email(email) do
+    with user when not is_nil(user) <- Account.get_user_by_email(email) do
+      {:ok, user}
+    else
+      nil -> {:error, :user_not_found}
+    end
+  end
+
+  @dialyzer {:no_return, create_user_token: 2}
```
geekingfrog: what are these annotations? You shouldn't just silence dialyzer without at the very least a comment explaining why you're doing that.
In this case, you get this error because there is no `scopes` keyword argument to this function.

Also, doing `with {:ok, x} <- foo(), do: {:ok, x}` is useless, you can directly call `foo()` without `with`.
`with` is useful when you want to chain several calls, otherwise you can use `case` to manipulate the results, or nothing at all if the return types are fine (this is the case here).

lib/teiserver_web/controllers/api/admin/user_controller.ex
```diff
@@ -0,0 +1,139 @@
...
+  defp update_user_stats(user_id, params) do
+    stat_fields = Map.take(params, @stat_fields)
+
+    with {:ok, _} <- update_stats_if_needed(user_id, stat_fields) do
+      :ok
+    end
+  end
+
+  defp update_stats_if_needed(user_id, stat_fields) do
+    if Map.size(stat_fields) > 0 do
```
geekingfrog: I'm getting a warning here: "Map.size/1 is deprecated. Use Kernel.map_size/1 instead". Note that `Kernel` is automatically imported so you can simply use `map_size`.

lib/teiserver_web/controllers/api/admin/user_controller.ex
```diff
@@ -0,0 +1,139 @@
...
+               scopes: app.scopes
+             },
+             create_refresh: true,
+             scopes: app.scopes
+           ) do
+      {:ok, token}
+    end
+  end
+
+  @dialyzer {:nowarn_function, build_user_response: 2}
```
geekingfrog: Same here, silencing dialyzer should come with a comment explaining why. In this case, I don't think there is any need to silence it.

lib/teiserver_web/controllers/api/admin/user_controller.ex
```diff
@@ -0,0 +1,139 @@
...
+
+    timestamps = %{
+      inserted_at: Map.get(user, :inserted_at),
+      updated_at: Map.get(user, :updated_at)
+    }
+
+    %{user: Map.merge(user_data, timestamps), credentials: credentials}
+  end
+
+  defp handle_error(conn, {:error, changeset}) when is_struct(changeset, Ecto.Changeset) do
```
geekingfrog: Instead of using `is_struct`, the idiomatic way of doing that is:
`handle_error(conn, {:error, %Ecto.changeset{} = changeset}) do`

lib/teiserver_web/controllers/api/admin/user_controller.ex
wrath625: Thanks for the explanation for what was going on here.  I couldn't for the life of my figure out why dialyzer was mad.  This fixed both dialyzer errors.

geekingfrog: yeah, dialyzer errors are not great, it's better than nothing, but often it's super confusing.

PR #711, opened 2025-07-26
Kill battle with `!stop`

geekingfrog:
This is the start of handling commands from players and relay them to autohost.
It requires https://github.com/beyond-all-reason/recoil-autohost/issues/21 to be fixed to actually work.

Couple of points here:

* I've refactored the way to issue request from server to client and block until client responds. I think it'll require a bit more work to handle error response, but we can see when we start copy/pasting too much code.
* I changed the order of startup between autohost and battles. That means battle can freely `call` into autohosts, but autohosts should only `cast` to battle. Following this rule (based on init order in the supervision tree) should prevent deadlocks.
* I added the `send_message` logic. Turn out it's irrelevant when it comes to stopping a battle, but we'll need that anyway, so might as well leave it there.

The next big piece is going to figure out voting (and tracking players in battle).

### DIFF
--- a/lib/teiserver/autohost.ex
+++ b/lib/teiserver/autohost.ex
@@ -82,4 +82,8 @@ defmodule Teiserver.Autohost do
           :ok | {:error, reason :: term()}
   defdelegate send_message(autohost, payload),
     to: Teiserver.Autohost.TachyonHandler
+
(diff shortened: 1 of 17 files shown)


geekingfrog:
The test failure seems flaky, although it's on the things I'm touching. I'm going to look a bit more into it but that should be good to start having a look.

PR #713, opened 2025-07-31
Simple matchmaking queue statistics

### LINKED ISSUE #683: [Feature]: store simple stats for each matchmaking queue
### Describe the Feature

For very simple monitoring purpose, store in each matchmaking queue state a few numbers:
* total number of join
* total number of leaver without match
* average waiting time before match

for parties, mutate the state once the entire party has joined.

### Expected Behavior

The state of each queue should have a new key with the following:
```
stats: %{
  total_joined: xxx,
  total_left: xxx,
  total_wait_time_s: xxx # only store the seconds, should be enough
  total_matched: xxx # the number of time the queue spawned a pairing room
}
```

Add these stats to the queue registry.

The tests can directly query the registry to get these.

wrath625:
Added simple incremental matchmaking stats which closes #683 

Added some basic unit tests to make sure it increments as intended. I didn't go too deep into trying to increment large party sizes and just implemented the logic for a "party of 1"

`total_wait_time_s` is a total aggregate of time waited, not an average.  If this is incorrect I can swap it.

### DIFF
--- a/lib/teiserver/matchmaking.ex
+++ b/lib/teiserver/matchmaking.ex
@@ -14,6 +14,7 @@ defmodule Teiserver.Matchmaking do
   @type lost_reason :: Matchmaking.PairingRoom.lost_reason()
   @type cancelled_reason :: Matchmaking.QueueServer.cancelled_reason()
   @type ready_data :: Matchmaking.PairingRoom.ready_data()
+  @type stats :: Matchmaking.QueueServer.stats()
 
   @spec lookup_queue(Matchmaking.QueueServer.id()) :: pid() | nil
   def lookup_queue(queue_id) do
@@ -53,6 +54,12 @@ defmodule Teiserver.Matchmaking do
   """
   @spec ready(pid(), ready_data()) :: :ok | {:error, term()}
   defdelegate ready(room_pid, user_id), to: Matchmaking.PairingRoom
+
+  @doc """
+  Get statistics for a specific queue
+  """
+  @spec get_stats(queue_id :: String.t()) :: {:ok, stats()} | {:error, :not_found}
+  defdelegate get_stats(queue_id), to: Matchmaking.QueueServer
 
   @doc """
   Kill and restart all matchmaking queues. This can be used to reset the
--- a/lib/teiserver/matchmaking/queue_server.ex
+++ b/lib/teiserver/matchmaking/queue_server.ex
@@ -23,6 +23,16 @@ defmodule Teiserver.Matchmaking.QueueServer do
   @type id :: String.t()
 
   @typedoc """
+  Queue statistics for monitoring
+  """
+  @type stats :: %{
+          total_joined: non_neg_integer(),
+          total_left: non_neg_integer(),
+          total_wait_time_s: non_neg_integer(),
+          total_matched: non_neg_integer()
+        }
+
+  @typedoc """
   Internal settings for the queue so it's easier to control the frequency of
   ticks and whatnot
   """
@@ -66,10 +76,10 @@ defmodule Teiserver.Matchmaking.QueueServer do
           # committed to joining this queue
           pending_parties: %{
             Party.id() => %{waiting_for: [T.userid()], joined: [T.userid()], tref: :timer.tref()}
-          }
+          },
 
-          # TODO: add some bits for telemetry (see QueueWaitServer) like avg
-          # wait time and join count
+          # queue statistics for monitoring
+          stats: stats()
         }
 
   @type cancelled_reason ::
@@ -120,7 +130,13 @@ defmodule Teiserver.Matchmaking.QueueServer do
       members: Map.get(attrs, :members, []),
       monitors: MC.new(),
       pairings: [],
-      pending_parties: %{}
+      pending_parties: %{},
+      stats: %{
+        total_joined: 0,
+        total_left: 0,
+        total_wait_time_s: 0,
+        total_matched: 0
+      }
     }
   end
 
@@ -236,9 +252,19 @@ defmodule Teiserver.Matchmaking.QueueServer do
       :ok ->
         game_type = MatchLib.game_type(state.queue.team_size, state.queue.team_count)
         new_member = Member.new([player_id], game_type)
-        new_state = add_member_to_queue(state, new_member)
+
+        new_state =
+          state
+          |> add_member_to_queue(new_member)
+          |> update_in([:stats, :total_joined], &(&1 + 1))
+
         {:reply, {:ok, self()}, new_state}
     end
+  end
+
+  @impl true
+  def handle_call(:get_stats, _from, state) do
+    {:reply, {:ok, state.stats}, state}
   end
 
   @impl true
@@ -260,6 +286,7 @@ defmodule Teiserver.Matchmaking.QueueServer do
             state
             |> Map.update!(:pending_parties, &Map.delete(&1, party_id))
             |> add_member_to_queue(member)
+            |> update_in([:stats, :total_joined], &(&1 + length(member.player_ids)))
 
           Logger.info("Party #{party_id} with members #{inspect(member.player_ids)} joined queue")
 
@@ -293,7 +320,8 @@ defmodule Teiserver.Matchmaking.QueueServer do
         {:reply, {:error, :not_queued}, state}
 
       {:ok, state} ->
-        {:reply, :ok, state}
+        new_state = update_in(state, [:stats, :total_left], &(&1 + 1))
+        {:reply, :ok, new_state}
     end
   end
 
@@ -372,7 +400,9 @@ defmodule Teiserver.Matchmaking.QueueServer do
   end
 
   @impl true
-  def handle_info(:tick, state) do
+  def handle_info(:tick, state), do: handle_info({:tick, DateTime.utc_now()}, state)
+
+  def handle_info({:tick, now}, state) do
     new_state =
       case match_members(state) do
         :no_match ->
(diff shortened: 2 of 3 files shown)


lib/teiserver/matchmaking/member.ex
```diff
@@ -64,8 +64,8 @@ defmodule Teiserver.Matchmaking.Member do
       rt ->
         ratings = rt.ratings
         n = Enum.count(ratings)
-        skill = Enum.sum_by(ratings, & &1.skill)
-        uncertainty = Enum.sum_by(ratings, & &1.uncertainty)
+        skill = Enum.sum(Enum.map(ratings, & &1.skill))
+        uncertainty = Enum.sum(Enum.map(ratings, & &1.uncertainty))
```
geekingfrog: what is the dialyzer error you were getting? Because I get nothing, and the 2 expressions are the same, I'd rather keep sum_by

lib/teiserver/matchmaking/queue_server.ex
```diff
@@ -237,10 +248,17 @@ defmodule Teiserver.Matchmaking.QueueServer do
         game_type = MatchLib.game_type(state.queue.team_size, state.queue.team_count)
         new_member = Member.new([player_id], game_type)
         new_state = add_member_to_queue(state, new_member)
+        # Update stats
```
geekingfrog: useless comment.
Also, you can use the pipe operator:
```elixir
new_state = state
  |> add_member_to_queue(new_member)
  |> update_stats(:total_joined, 1)
```

lib/teiserver/matchmaking/queue_server.ex
```diff
@@ -293,7 +312,9 @@ defmodule Teiserver.Matchmaking.QueueServer do
         {:reply, {:error, :not_queued}, state}
 
       {:ok, state} ->
-        {:reply, :ok, state}
+        # Update stats for leaving
```
geekingfrog: same, useless comment

lib/teiserver/matchmaking/queue_server.ex
```diff
@@ -551,6 +587,12 @@ defmodule Teiserver.Matchmaking.QueueServer do
     apply(alg_module, :get_matches, [state.members, alg_state])
   end
 
+  defp update_stats(state, stat_key, increment) do
+    Map.update!(state, :stats, fn stats ->
+      Map.update!(stats, stat_key, &(&1 + increment))
+    end)
+  end
```
geekingfrog: I am not convinced by this function. You should be able to remove it and call a deep update inline where required, it would make the code easier to follow.
```elixir
new_state = update_in(state, [:stats, :total_matched], &(&1 + 1))
```
and so on

lib/teiserver/matchmaking/queue_server.ex
```diff
@@ -405,10 +426,25 @@ defmodule Teiserver.Matchmaking.QueueServer do
...
 
+          # Calculate total wait time for all matched members
+          total_wait_time =
+            matches
+            |> Enum.flat_map(fn teams ->
+              for team <- teams, member <- team do
+                DateTime.diff(DateTime.utc_now(), member.joined_at, :second)
+              end
+            end)
+            |> Enum.sum()
```
geekingfrog, line 443: It's a bit odd to have both `Enum.flat_map` with a for comprehension. I suggest doing just the comprehension, you can chain things:
```elixir
for teams <- matches, team <- teams, member <- team do
  DateTime.diff(DateTime.utc_now(), member.joined_at, :second)
end
|> Enum.sum()
```
haven't tested that but that should work, the flat_map is redundant there.

lib/teiserver/matchmaking/queue_server.ex
```diff
@@ -237,10 +248,17 @@ defmodule Teiserver.Matchmaking.QueueServer do
...
+        # Update stats
+        new_state = update_stats(new_state, :total_joined, 1)
         {:reply, {:ok, self()}, new_state}
     end
   end
 
+  @impl true
+  def handle_call(:get_state, _from, state) do
+    {:reply, {:ok, state}, state}
+  end
```
geekingfrog, line 268: There is `:sys.get_state(pid)` which does the same thing.

test/teiserver/matchmaking/queue_test.exs
geekingfrog, line 1: I am not a fan of tests that inspect the internal state of a genserver.
Instead, I would suggest you implement a function `get_stats(queue_id)` that does the call and returns the stats. This way, you keep the genserver state an implementation details, the test will be more robust and you can freely refactor things.

test/teiserver/matchmaking/queue_test.exs
```diff
@@ -66,4 +66,84 @@ defmodule Teiserver.Matchmaking.QueueTest do
...
+      user2 = Central.Helpers.GeneralTestLib.make_user()
+
+      # Join first user
+      {:ok, _pid} = Matchmaking.join_queue(queue_id, user1.id)
+      {:ok, state} = GenServer.call(queue_pid, :get_state)
+      assert state.stats.total_joined == 1
+      assert state.stats.total_wait_time_s == 0
+
+      # Wait for 1 second
+      Process.sleep(1000)
```
geekingfrog: This is not great, it makes the tests slower, and it's often brittle (though here it doesn't matter).
Instead, what I suggest is to overload the handler for the `:tick` event so that you can pass a time, plus a comment saying that it's useful for tests.

```elixir
  @impl true
def handle_info(:tick, state), do: handle_info(:tick, DateTime.utc_now(), state) # useful for testing

def handle_info(:tick, now, state) do
  # same as usual.
end
```

And in the test, you can pass whatever clock you want.

test/teiserver/matchmaking/queue_test.exs
```diff
@@ -66,4 +66,84 @@ defmodule Teiserver.Matchmaking.QueueTest do
...
+
+      # Join second user (this should trigger a match)
+      {:ok, _pid} = Matchmaking.join_queue(queue_id, user2.id)
+
+      # Trigger the tick to process matches
+      send(queue_pid, :tick)
+
+      # Give it a moment to process
+      Process.sleep(100)
```
geekingfrog: There are a few helper functions inside `Teiserver.Support.Polling` that you should use instead of arbitrary sleep.
But I find that curious you need it at all tbh. Because messages are processed in the order they are received. So by the time the genserver deal with the `call`, the `:tick` message should've been processed already.
Why did you need this sleep?

geekingfrog:
also, why so many flaky tests on CI??? That's getting super annoying.

wrath625:
> ~also, why so many flaky tests on CI??? That's getting super annoying.~
> 
> This is not a flake!!!

Yeah when I modified the email validation in #709 I think that now doesn't play well with the hardcoded bot users

I fixed coordinator@teiserver -> coordinator@teiserver.local

but I guess every time this runs its going to hit a new account until they're all corrected

lib/teiserver/matchmaking/member.ex
wrath625: sigh. I had the wrong elixir version.  1.17 doesnt have sum_by apparently.  Apparently I didnt re-do the elixir setup properly when i swapped into WSL last week.

wrath625:
Uh oh, I think Ill need to go back to #709 and update a few other tests that dont follow the email rules..

wrath625:
So it looks like the change that added email validation to the user schema in #709 broke some tests, I've moved all the fixes to a separate PR: #714

lib/teiserver/matchmaking.ex
```diff
@@ -54,6 +54,23 @@ defmodule Teiserver.Matchmaking do
...
+    case Matchmaking.QueueRegistry.lookup(queue_id) do
+      nil ->
+        {:error, :not_found}
+
+      pid ->
+        case GenServer.call(pid, :get_stats, 1000) do
+          {:ok, stats} -> {:ok, stats}
+          _ -> {:error, :not_found}
+        end
+    end
```
geekingfrog: This may seems to work, but is racy. What if the genserver dies between the lookup in the registry and this call?
Instead, you should use [this pattern](https://github.com/geekingfrog/teiserver/blob/04715a7c43cbb03d7a4b96dd6367b57d0c661624/lib/teiserver/player/session.ex#L128-L134) which doesn't suffer from race condition.

Also, this is something that is basically only relevant to the queue server, it doesn't really mix other concepts together. As such, it should live inside the `queue_server.ex` file, and this context module should re-export it with a defdelegate, as it's done around.

In the queue_server.ex file, note the use of `via_tuple` to automatically do the lookup with a given registry. 
This is handled by the internal `GenServer` abstraction.

and finally, the return type isn't super useful. I suggest you define a type in queue_server.ex like
```elixir
@type stats :: %{            
            total_joined: non_neg_integer(),
            total_left: non_neg_integer(),
            total_wait_time_s: non_neg_integer(),
            total_matched: non_neg_integer()
}
```
then you can use this type in the genserver state, re-export this type in the `matchmaking.ex` context, and use this type for the `get_stats` function.

test/teiserver/matchmaking/queue_test.exs
```diff
@@ -66,4 +66,80 @@ defmodule Teiserver.Matchmaking.QueueTest do
...
   end
+
+  describe "queue statistics" do
+    test "initial stats are zero", %{queue_id: queue_id} do
+      # Initially stats should be zero
+      {:ok, stats} = Matchmaking.get_stats(queue_id)
+      assert stats.total_joined == 0
+      assert stats.total_left == 0
+      assert stats.total_matched == 0
+      assert stats.total_wait_time_s == 0
```
geekingfrog: perhaps a nitpick, but you can do all of these asserts at once:
```
assert stats == %{total_joined: 0, total_left: 0, ...}
```

geekingfrog (commented):
With fresh eyes, I realised there is a missing piece:

The count of player that left the queue without being paired. This can be tracked in the `remove_member` function I believe. Their wait time should also be tallied at that point, perhaps under a different key? Don't need something too sophisticated. But I think knowing how many players bounce off is a useful thing to have.

PR #714, opened 2025-08-02
Fix email validation in tests and bot accounts after PR #709

wrath625:
Working on #713, @geekingfrog noticed a bunch of non-flakey test failures.  I _think_ this resolves that issue without rolling back the change introduced to add email validation to the user schema in #709 

- Update AccountTestLib.user_fixture() to use valid email addresses with @test.local domain
- Fix admin user controller tests to use valid email addresses
- Fix all bot account emails to use valid domains:
  - Coordinator: coordinator@teiserver.local
  - Bridge: bridge@teiserver.local
  - Match Monitor: match_monitor@teiserver.local
  - Accolade: accolades_bot@teiserver.local
- This resolves test failures caused by email validation introduced in PR #709

### DIFF
--- a/lib/teiserver/account/accolades/accolade_bot_server.ex
+++ b/lib/teiserver/account/accolades/accolade_bot_server.ex
@@ -172,7 +172,7 @@ defmodule Teiserver.Account.AccoladeBotServer do
     user =
       Account.get_user(nil,
         search: [
-          email: "accolades_bot@teiserver"
+          email: "accolades_bot@teiserver.local"
         ]
       )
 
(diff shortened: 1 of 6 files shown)


geekingfrog:
Nice.
But also, wtf is going on with the CI? It seems it's using some cache somewhere somehow, because it's failing on some tests that aren't even in this branch.

geekingfrog:
Merged since it's just useful, but we need to have some deeper look into this cursed CI

L-e-x-o-n:
On the prod server this will now fail to find the bots by their old emails and will create duplicates with the new ones, right?

geekingfrog:
> On the prod server this will now fail to find the bots by their old emails and will create duplicates with the new ones, right?

Just deployed on the dev server and indeed, that's what happens. Do you think that's a problem?

PR #718, opened 2025-08-06
simple matchmaking admin dashboard

### LINKED ISSUE #684: [Feature]: Create ultra basic admin dashboard for matchmaking queues
### Describe the Feature

Once we have https://github.com/beyond-all-reason/teiserver/issues/683 we can display this on a webpage in the admin panel.

### Expected Behavior

* Add a new live view under `lib/teiserver_web/live/admin/matchmaking{.ex,.html.heex}`
* Add this view to a route behind admin permissions `/admin/dashboard/matchmaking`
* get the queues and then their stats
* display all that.

### Screenshots

_No response_

### Additional context

_No response_

wrath625:
Adding a simple admin dashboard for viewing matchmaking queues at `/admin/matchmaking`

image

Closes #684

### DIFF
--- a/lib/teiserver/matchmaking/queue_server.ex
+++ b/lib/teiserver/matchmaking/queue_server.ex
@@ -22,14 +22,22 @@ defmodule Teiserver.Matchmaking.QueueServer do
 
   @type id :: String.t()
 
+  # Internal queue statistics (without player_count)
+  @typep internal_stats :: %{
+           total_joined: non_neg_integer(),
+           total_left: non_neg_integer(),
+           total_wait_time_s: non_neg_integer(),
+           total_matched: non_neg_integer()
+         }
   @typedoc """
-  Queue statistics for monitoring
+  Queue statistics for monitoring (includes player_count)
   """
   @type stats :: %{
           total_joined: non_neg_integer(),
           total_left: non_neg_integer(),
           total_wait_time_s: non_neg_integer(),
-          total_matched: non_neg_integer()
+          total_matched: non_neg_integer(),
+          player_count: non_neg_integer()
         }
 
   @typedoc """
@@ -79,7 +87,7 @@ defmodule Teiserver.Matchmaking.QueueServer do
           },
 
           # queue statistics for monitoring
-          stats: stats()
+          stats: internal_stats()
         }
 
   @type cancelled_reason ::
@@ -258,13 +266,25 @@ defmodule Teiserver.Matchmaking.QueueServer do
           |> add_member_to_queue(new_member)
           |> update_in([:stats, :total_joined], &(&1 + 1))
 
+        broadcast_update(new_state)
         {:reply, {:ok, self()}, new_state}
     end
   end
 
   @impl true
   def handle_call(:get_stats, _from, state) do
-    {:reply, {:ok, state.stats}, state}
+    player_count = calculate_player_count(state)
+    private_stats = Map.put(state.stats, :player_count, player_count)
+
+    public_stats = %{
+      total_joined: private_stats.total_joined,
+      total_left: private_stats.total_left,
+      total_wait_time_s: private_stats.total_wait_time_s,
+      total_matched: private_stats.total_matched,
+      player_count: private_stats.player_count
+    }
+
+    {:reply, {:ok, public_stats}, state}
   end
 
   @impl true
@@ -290,6 +310,7 @@ defmodule Teiserver.Matchmaking.QueueServer do
 
           Logger.info("Party #{party_id} with members #{inspect(member.player_ids)} joined queue")
 
+          broadcast_update(new_state)
           {:reply, {:ok, self()}, new_state}
 
         {[_], rest} ->
@@ -321,6 +342,7 @@ defmodule Teiserver.Matchmaking.QueueServer do
 
       {:ok, state} ->
         new_state = update_in(state, [:stats, :total_left], &(&1 + 1))
+        broadcast_update(new_state)
         {:reply, :ok, new_state}
     end
   end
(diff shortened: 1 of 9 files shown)


lib/teiserver_web/live/admin/matchmaking/index.ex
```diff
@@ -0,0 +1,90 @@
...
+  end
+
+  @impl true
+  def handle_params(_params, _url, socket) do
+    {:noreply, get_queues(socket)}
+  end
+
+  @impl true
+  def handle_info(:refresh_queues, socket) do
+    :timer.send_interval(5_000, :refresh_queues)
```
geekingfrog: `:send_interval` sends a message every `n` ms. Calling that every time mean every 5s, you are doubling the amount of message sent to the process.
Either use one `send_interval` once in init, or `send_later` in init and the handler.

lib/teiserver_web/live/admin/matchmaking/index.ex
```diff
@@ -0,0 +1,90 @@
...
+    case Matchmaking.get_stats(queue_id) do
+      {:ok, stats} -> stats
+      _ -> %{total_joined: 0, total_left: 0, total_matched: 0, total_wait_time_s: 0}
+    end
+  end
+
+  defp get_queue_player_count(queue_id) do
+    queue_id
+    |> Matchmaking.QueueRegistry.lookup()
+    |> get_queue_state()
```
geekingfrog: This should be avoided. As soon as you depend on the state of the queue, that means it's no longer internal and it becomes part of its public API, which is quite bad.
If you need to know the number of player in queue, expose a function to get this piece of data and expose it as part of the matchmaking context.
But you can also make a case that this should belong inside the stats object.

lib/teiserver_web/live/admin/matchmaking/index.ex
```diff
@@ -0,0 +1,90 @@
...
+  end
+
+  defp get_queue_state(nil), do: nil
+
+  defp get_queue_state(pid) do
+    case GenServer.call(pid, :get_state, 1000) do
+      {:ok, state} -> state
+      _ -> nil
+    end
+  end
```
geekingfrog: If anything, this should live in the queue_server file, not here.
The only return value is `{:ok, state}` so the `case` is redundant as well.

geekingfrog (commented):
For such work, it's helpful to add some screenshot of what it looks like.

Also, I think this isn't very idiomatic in phoenix/liveview land.
Instead of polling, you should use pubsub to get updates. When the stats update, the queue server should broadcast the new value. Each liveview subscribe to the channel and updates based on these broadcasted events.
You should also put the number of player in the stats object and broadcast that as well.

The general pattern for pubsub is

```elixir
    PubSub.broadcast(
      Teiserver.PubSub,
      "global_moderation",  # channel name as string
      %{
        channel: "global_moderation",  # the same thing
        event: :new_response,  # this key is usually sent as a tag for the payload
        # anything after is free form.
        response: response
      }
    )
```

You should put the subscribe method in the public api of matchmaking, though the abstraction is a bit leaky because you need to pattern match on the message received, so you can't fully abstract away the channel name. But at least you can put some doc on the `subscribe` function.

lib/teiserver/matchmaking/queue_server.ex
```diff
@@ -604,7 +621,36 @@ defmodule Teiserver.Matchmaking.QueueServer do
...
+  @spec get_player_count(pid()) :: non_neg_integer()
+  def get_player_count(pid) when is_pid(pid) do
+    GenServer.call(pid, :get_player_count, 1000)
+  end
+
+  @spec get_player_count(String.t()) :: non_neg_integer()
+  def get_player_count(queue_id) when is_binary(queue_id) do
+    via_tuple = via_tuple(queue_id)
+    GenServer.call(via_tuple, :get_player_count, 1000)
+  end
```
geekingfrog: I think this is now unused and can be removed.

lib/teiserver/matchmaking/queue_server.ex
```diff
@@ -258,13 +260,22 @@ defmodule Teiserver.Matchmaking.QueueServer do
...
-    {:reply, {:ok, state.stats}, state}
+    player_count = calculate_player_count(state)
+    stats = Map.put(state.stats, :player_count, player_count)
+    {:reply, {:ok, stats}, state}
+  end
+
+  @impl true
+  def handle_call(:get_player_count, _from, state) do
+    player_count = calculate_player_count(state)
+    {:reply, player_count, state}
```
geekingfrog: `:get_player_count` is now unused I believe.

lib/teiserver/matchmaking/queue_server.ex
```diff
@@ -604,7 +621,36 @@ defmodule Teiserver.Matchmaking.QueueServer do
...
+    via_tuple = via_tuple(queue_id)
+    GenServer.call(via_tuple, :get_player_count, 1000)
+  end
+
+  @spec calculate_player_count(state()) :: non_neg_integer()
+  defp calculate_player_count(state) do
+    state.members
+    |> Enum.reduce(0, fn member, acc ->
+      acc + length(member.player_ids)
+    end)
```
geekingfrog: You can also use `Enum.sum_by/2`

lib/teiserver/matchmaking/queue_server.ex
```diff
@@ -604,7 +621,36 @@ defmodule Teiserver.Matchmaking.QueueServer do
   """
   @spec get_stats(id()) :: {:ok, stats()} | {:error, :not_found}
   def get_stats(queue_id) do
-    via_tuple = QueueRegistry.via_tuple(queue_id)
+    via_tuple = via_tuple(queue_id)
```
geekingfrog, line 633: why the short timeout btw? If you deviate from the default, you should at the very least put a comment explaining why.

geekingfrog (commented):
You need to fix the tests as well.

There is one thing where I'm not too sure about. When you broadcast a stats update you get the player count there. The nice thing with this approach is that you don't need to keep a separate count in the stats object that must be kept in sync.
I think I like this approach. However, you then need to have a separate type for the internal `stats` map of the queue, that does *not* have this `:player_count` key, and the stats object you are broadcasting, which does have it.

lib/teiserver/matchmaking/queue_server.ex
```diff
@@ -23,7 +23,17 @@ defmodule Teiserver.Matchmaking.QueueServer do
   @type id :: String.t()
 
   @typedoc """
-  Queue statistics for monitoring
+  Internal queue statistics (without player_count)
+  """
+  @type internal_stats :: %{
```
geekingfrog: Since this is only used internally, you can use `@typep`. Not sure if `@typedoc` works with that, `@doc` doesn't with `defp` for example. If you get a warning or error, change it to a comment.

wrath625:
modifying the type to private caused me to need to return a public map structure of stats in `handle_call(:get_stats, ..)`

not sure if this is the best approach but it appeased the test failures I was getting

lib/teiserver/matchmaking/queue_server.ex
```diff
@@ -23,15 +23,27 @@ defmodule Teiserver.Matchmaking.QueueServer do
...
+  @typep stats :: %{
+           total_joined: non_neg_integer(),
+           total_left: non_neg_integer(),
+           total_wait_time_s: non_neg_integer(),
+           total_matched: non_neg_integer(),
+           player_count: non_neg_integer()
+         }
+
+  # Public type alias for external modules to reference
+  @type queue_stats :: stats()
```
geekingfrog: There is no need for this alias since it's the same type, might as well declare `stats` as public already.

wrath625: yeah that was a big mixup, should be good now

geekingfrog:
Going to merge that. The failing test is a flake (☠️)

PR #719, opened 2025-08-06
Fix s.battle.teams crashes

wrath625:
Fix for crashes related to #705 

Discord convo: https://discord.com/channels/549281623154229250/564591092360675328/1402703973617635481

## Problem
Error logs showed KeyError crashes when accessing `:host_teamsize` from nil values:

```
(KeyError) key :host_teamsize not found in: nil
(teiserver 0.1.0) lib/teiserver/lobby/libs/lobby_lib.ex:554: Teiserver.Lobby.LobbyLib.get_team_config/1
```

Also found login timeouts due to blocking `get_team_config(:all)` calls.

### 1. Added nil checks in `get_team_config/1`
- Added `case` blocks to handle nil returns from `Teiserver.Coordinator.get_team_config/1`
- Prevents KeyError when consul servers don't exist

### 2. Made battle teams login bulk broadcast asynchronous
- Moved `get_team_config(:all)` call to async task in `spring_out.ex`
- Prevents login timeouts

### 3. Added nil checks in downstream functions
- Added nil check in `spring_tcp_server.ex` before calling `SpringOut.reply`
- Added nil guard in `spring_battle_out.ex` to prevent Jason.encode! crashes

### DIFF
--- a/lib/teiserver/lobby/libs/lobby_lib.ex
+++ b/lib/teiserver/lobby/libs/lobby_lib.ex
@@ -532,16 +532,31 @@ defmodule Teiserver.Lobby.LobbyLib do
           tasks =
             Enum.map(lobby_ids, fn lobby_id ->
               Task.async(fn ->
-                team_config = Teiserver.Coordinator.get_team_config(lobby_id)
+                try do
+                  case Teiserver.Coordinator.get_team_config(lobby_id, 4_000) do
+                    %{host_teamsize: team_size, host_teamcount: team_count} ->
+                      {lobby_id, %{teamSize: team_size, nbTeams: team_count}}
 
-                {lobby_id,
-                 %{teamSize: team_config.host_teamsize, nbTeams: team_config.host_teamcount}}
+                    _ ->
+                      nil
+                  end
+                catch
+                  _, _ -> nil
+                end
               end)
             end)
 
-          data = Task.await_many(tasks, 2_000) |> Map.new()
-          Teiserver.cache_put(:application_temp_cache, :battle_teams, {data, now})
-          data
+          data =
+            Task.await_many(tasks)
+            |> Enum.reject(&is_nil/1)
+            |> Map.new()
+
+          if map_size(data) == 0 do
+            nil
+          else
+            Teiserver.cache_put(:application_temp_cache, :battle_teams, {data, now})
+            data
+          end
       end
     else
       nil
@@ -550,8 +565,17 @@ defmodule Teiserver.Lobby.LobbyLib do
 
   def get_team_config(lobby_id) when is_integer(lobby_id) do
     if Teiserver.Config.get_site_config_cache("lobby.Broadcast Battle Teams Information") do
-      team_config = Teiserver.Coordinator.get_team_config(lobby_id)
-      %{lobby_id => %{teamSize: team_config.host_teamsize, nbTeams: team_config.host_teamcount}}
+      try do
+        case Teiserver.Coordinator.get_team_config(lobby_id) do
+          %{host_teamsize: team_size, host_teamcount: team_count} ->
+            %{lobby_id => %{teamSize: team_size, nbTeams: team_count}}
+
+          _ ->
+            nil
+        end
+      catch
+        _, _ -> nil
+      end
     else
       nil
     end
--- a/lib/teiserver/protocols/spring/spring_out.ex
+++ b/lib/teiserver/protocols/spring/spring_out.ex
@@ -738,11 +738,16 @@ defmodule Teiserver.Protocols.SpringOut do
     end)
 
     # Send BATTLETEAMS message
-    teams_data = Teiserver.Battle.get_team_config(:all)
+    # Make this async to avoid blocking login
+    current_process = self()
 
-    if teams_data do
-      send(self(), {:battle_teams, teams_data})
-    end
+    Task.Supervisor.start_child(Teiserver.TaskSupervisor, fn ->
+      teams_data = Teiserver.Battle.get_team_config(:all)
+
+      if teams_data do
+        send(current_process, {:battle_teams, teams_data})
+      end
+    end)
 
     # CLIENTSTATUS entries
     clients
--- a/lib/teiserver/application.ex
+++ b/lib/teiserver/application.ex
@@ -27,6 +27,7 @@ defmodule Teiserver.Application do
 
         # Start phoenix pubsub
         {Phoenix.PubSub, name: Teiserver.PubSub},
+        {Task.Supervisor, name: Teiserver.TaskSupervisor},
         TeiserverWeb.Telemetry,
 
         # Start the Ecto repository
--- a/lib/teiserver/coordinator.ex
+++ b/lib/teiserver/coordinator.ex
@@ -119,15 +119,15 @@ defmodule Teiserver.Coordinator do
     end
   end
 
-  @spec call_consul(pid() | T.lobby_id(), any) :: any
-  def call_consul(lobby_id, msg) when is_integer(lobby_id) do
+  @spec call_consul(pid() | T.lobby_id(), any, timeout()) :: any
+  def call_consul(lobby_id, msg, timeout \\ 5_000) when is_integer(lobby_id) do
     case get_consul_pid(lobby_id) do
       nil ->
         nil
 
       pid ->
         try do
-          GenServer.call(pid, msg)
+          GenServer.call(pid, msg, timeout)
 
           # If the process has somehow died, we just return nil
         catch
@@ -291,9 +291,9 @@ defmodule Teiserver.Coordinator do
     CacheUser.send_direct_message(get_coordinator_userid(), userid, msg)
   end
 
-  @spec get_team_config(integer()) :: map()
-  def get_team_config(lobby_id) do
-    call_consul(lobby_id, :get_team_config)
+  @spec get_team_config(integer(), timeout()) :: map()
+  def get_team_config(lobby_id, timeout \\ 5_000) do
+    call_consul(lobby_id, :get_team_config, timeout)
   end
 
   # Debug stuff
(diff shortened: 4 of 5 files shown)


lib/teiserver/lobby/libs/lobby_lib.ex
```diff
@@ -532,14 +532,30 @@ defmodule Teiserver.Lobby.LobbyLib do
...
             end)
 
-          data = Task.await_many(tasks, 2_000) |> Map.new()
+          data =
+            Task.await_many(tasks, 5_000)
+            |> Enum.filter(fn {_lobby_id, config} ->
+              case config do
+                nil -> false
+                _ -> true
+              end
```
geekingfrog: `config != nil`
but at that point, you should just return nil in the task above and use `Enum.reject(&is_nil/1)`

lib/teiserver/tcp/spring/spring_tcp_server.ex
```diff
@@ -414,7 +414,12 @@ defmodule Teiserver.SpringTcpServer do
       if Config.get_site_config_cache("lobby.Broadcast Battle Teams Information") and
            (Map.has_key?(new_values, :host_teamsize) or Map.has_key?(new_values, :host_teamcount)) do
         teams_data = Teiserver.Battle.get_team_config(lobby_id)
-        SpringOut.reply(:battle, :battle_teams, teams_data, nil, new_state)
+
+        if teams_data do
```
geekingfrog: What is that for? In other parts you wrapped the call with a try/rescue but here you are testing if you get nil? What is the correct way?

lib/teiserver/lobby/libs/lobby_lib.ex
```diff
@@ -532,14 +532,30 @@ defmodule Teiserver.Lobby.LobbyLib do
...
+                try do
+                  case Teiserver.Coordinator.get_team_config(lobby_id) do
+                    %{host_teamsize: team_size, host_teamcount: team_count} ->
+                      {lobby_id, %{teamSize: team_size, nbTeams: team_count}}
+
+                    _ ->
+                      {lobby_id, nil}
+                  end
+                rescue
+                  _ -> {lobby_id, nil}
```
geekingfrog: How did you test the rescue bit? I thought for timeout it was
```
catch
  :exit, {:timeout, _} -> ...
```
you should also catch `:noproc` I think in case the genserver isn't there.

lib/teiserver/lobby/libs/lobby_lib.ex
```diff
@@ -532,16 +532,35 @@ defmodule Teiserver.Lobby.LobbyLib do
...
+                      {lobby_id, %{teamSize: team_size, nbTeams: team_count}}
+
+                    _ ->
+                      {lobby_id, nil}
+                  end
+                catch
+                  :exit, {:timeout, _} -> {lobby_id, nil}
+                  :exit, :noproc -> {lobby_id, nil}
+                  :exit, _ -> {lobby_id, nil}
+                  _, _ -> {lobby_id, nil}
```
geekingfrog: you don't need to match like that if you're going to return the same value either way.
`catch _, _ -> nil` is enough.

lib/teiserver/lobby/libs/lobby_lib.ex
```diff
@@ -532,16 +532,35 @@ defmodule Teiserver.Lobby.LobbyLib do
...
+                end
               end)
             end)
 
-          data = Task.await_many(tasks, 2_000) |> Map.new()
-          Teiserver.cache_put(:application_temp_cache, :battle_teams, {data, now})
-          data
+          data =
+            Task.await_many(tasks, 5_000)
+            |> Enum.reject(fn {_lobby_id, config} -> is_nil(config) end)
```
geekingfrog: And because you are discarding the value anyway, you don't need to have a tuple with the lobby id.
`tasks` can be a `{lobby_id, stats}` or nil, and then you can just reject nil value, no need for a tuple in the nil case.

lib/teiserver/lobby/libs/lobby_lib.ex
```diff
@@ -532,16 +532,35 @@ defmodule Teiserver.Lobby.LobbyLib do
...
+                  _, _ -> {lobby_id, nil}
+                end
               end)
             end)
 
-          data = Task.await_many(tasks, 2_000) |> Map.new()
-          Teiserver.cache_put(:application_temp_cache, :battle_teams, {data, now})
-          data
+          data =
+            Task.await_many(tasks, 5_000)
```
geekingfrog: `5_000` is the default timeout, you don't need to specify it. However, it's also the default timeout for a `GenServer.call`. So when you call `get_team_config`, you have a 5s timeout. There, you are really careful to trap a timeout, but if a task is going to timeout, the entire `await_many` is also going to timeout (not always, but almost).

I would suggest you allow `get_team_config` to also take an optional timeout value (default to 5s), and when calling it there, lower it to 4s for example.
Unless the system is under tremendous load, that mean a timeout to `get_team_config` would not blow up `await_many`

lib/teiserver/lobby/libs/lobby_lib.ex
```diff
@@ -550,8 +569,20 @@ defmodule Teiserver.Lobby.LobbyLib do
...
+            %{lobby_id => %{teamSize: team_size, nbTeams: team_count}}
+
+          _ ->
+            nil
+        end
+      catch
+        :exit, {:timeout, _} -> nil
+        :exit, :noproc -> nil
+        :exit, _ -> nil
+        _, _ -> nil
```
geekingfrog, line 577: Same comment there regarding useless pattern match

lib/teiserver/protocols/spring/spring_out.ex
```diff
@@ -738,11 +738,11 @@ defmodule Teiserver.Protocols.SpringOut do
     end)
 
     # Send BATTLETEAMS message
-    teams_data = Teiserver.Battle.get_team_config(:all)
-
-    if teams_data do
+    # Make this async to avoid blocking login
+    Task.start(fn ->
```
geekingfrog: This isn't ideal. Because it doesn't spawn a task belonging to a supervision tree, when the application is shutting down, theses tasks won't be terminated as part of the application.

I would suggest following the recommendation of the doc about [dynamically supervised tasks](https://hexdocs.pm/elixir/Task.html#module-dynamically-supervised-tasks), add a `Task.Supervisor` to the main tree, and then use this supervisor to spawn this task.

lib/teiserver/protocols/spring/spring_out.ex
```diff
@@ -738,11 +738,11 @@ defmodule Teiserver.Protocols.SpringOut do
...
+      teams_data = Teiserver.Battle.get_team_config(:all)
       send(self(), {:battle_teams, teams_data})
-    end
+    end)
```
geekingfrog, line 750: This won't work anymore. Because `self()` represent the current process, now you are targeting the task itself, and not the process associated with the tcp connection. You need to store `self()` from outside the task and reference this binding from the task.

geekingfrog:
Please don't merge the master branch back into yours, it leads to very confusing history and harder to review code. If you want to incorporate changes that were made on the master branch, you should instead rebase your branch. Please remove the last merge commit (and optionally rebase).

Also, the tests are failing. Running `mix test --exclude needs_attention test/teiserver/protocols` for example leads to many failures like

```
** (FunctionClauseError) no function clause matching in Teiserver.SpringTcpServer.handle_info/2
    (teiserver 0.1.0) lib/teiserver/tcp/spring/spring_tcp_server.ex:199: Teiserver.SpringTcpServer.handle_info({#Reference<0.0.404355.3091102346.482934793.99784>, {:battle_teams, nil}},
<snip>
```

wrath625:
not sure why my rebase merged master into the branch, sorry about that - it should be sorted out now

lib/teiserver/lobby/libs/lobby_lib.ex
```diff
@@ -532,16 +532,31 @@ defmodule Teiserver.Lobby.LobbyLib do
...
+                  _, _ -> nil
+                end
               end)
             end)
 
-          data = Task.await_many(tasks, 2_000) |> Map.new()
-          Teiserver.cache_put(:application_temp_cache, :battle_teams, {data, now})
-          data
+          data =
+            Task.await_many(tasks, 4_000)
```
geekingfrog: This should not have a timeout of 4s, let's keep the default of 5s. Because if one of the task timeouts at 4s, then the `await_many` call is very very very likely to also timeout, so all the guarding you're doing to catch the timeout is going to be useless.

wrath625: reverted to 5

lib/teiserver/lobby/libs/lobby_lib.ex
```diff
@@ -532,16 +532,31 @@ defmodule Teiserver.Lobby.LobbyLib do
           tasks =
             Enum.map(lobby_ids, fn lobby_id ->
               Task.async(fn ->
-                team_config = Teiserver.Coordinator.get_team_config(lobby_id)
-
-                {lobby_id,
-                 %{teamSize: team_config.host_teamsize, nbTeams: team_config.host_teamcount}}
+                try do
+                  case Teiserver.Coordinator.get_team_config(lobby_id) do
```
geekingfrog: You are still missing the point with regards to timeout. As long as the timeout of `get_team_config` is the same as `await_many` the following can happen:

* start task at t0
* start waiting on task at t0 + ε (you can ignore the ε)
* task times out -> caught with a nil  ✓
* await many times out -> crash process 💀 Because the reporting of the task completion (nil) is async.

A relatively safe setting is setting a timeout of 4s for `get_team_config`, and 5s (the default) for `await_many`. If the interleaving lasts more than 1s, that mean the system is under such tremendous load that we'll have other , bigger problems.

wrath625: ahh okay its making sense now. got it.  Thanks for the detailed explanation.

PR #727, opened 2025-08-20
Add user/self event broadcast when roles are changed

### LINKED ISSUE #723: [Feature]: Send `user/self` event when user roles are changed
### Describe the Feature

Follow up on https://github.com/beyond-all-reason/teiserver/issues/675
When roles are changed through the admin web UI, if the user is connected they should receive an updated event.

### Expected Behavior

_No response_

### Screenshots

_No response_

### Additional context

_No response_

wrath625:
- Added role change detection to user_controller.ex
- Added `send_admin_role_update/1` function to session.ex to communicate with connected users
- Added `admin_role_update` handler and extracted `build_user_self_event/2` from `init/1` for reusability

Closes #723

### DIFF
--- a/lib/teiserver/player/session.ex
+++ b/lib/teiserver/player/session.ex
@@ -328,6 +328,14 @@ defmodule Teiserver.Player.Session do
       %{status: :offline}
   end
 
+  @doc """
+  Sends a user/self event to a user when their roles are updated
+  """
+  @spec update_user_roles(T.userid(), [String.t()]) :: :ok
+  def update_user_roles(user_id, roles) do
+    GenServer.cast(via_tuple(user_id), {:user, {:role_updated, roles}})
+  end
+
   @spec create_party(T.userid()) ::
           {:ok, Party.id()} | {:error, :already_in_party} | {:error, reason :: term()}
   def create_party(user_id) do
@@ -964,6 +972,12 @@ defmodule Teiserver.Player.Session do
     {:noreply, state}
   end
 
+  def handle_cast({:user, {:role_updated, roles}}, state) do
+    send_to_player!({:user, {:role_updated, roles}}, state)
+    new_state = %{state | user: %{state.user | roles: roles}}
+    {:noreply, new_state}
+  end
+
   def handle_cast({:party, {:updated, party_state}}, state) do
     if (state.party.current_party == party_state.id && state.party.version <= party_state.version) ||
          Enum.any?(state.party.invited_to, fn {v, id} ->
@@ -1306,7 +1320,7 @@ defmodule Teiserver.Player.Session do
       # but in the meantime, just defensively get the country
       country: Map.get(user, :country, "??"),
       status: status,
-      roles: Teiserver.Player.TachyonHandler.convert_teiserver_roles_to_tachyon(user.roles)
+      roles: user.roles
     }
 
     PubSub.broadcast!(
--- a/lib/teiserver/player/tachyon_handler.ex
+++ b/lib/teiserver/player/tachyon_handler.ex
@@ -52,25 +52,7 @@ defmodule Teiserver.Player.TachyonHandler do
     user = initial_state.user
     Logger.metadata(user_id: user.id)
 
-    {friends, incoming, outgoing} = get_user_friends(state.user.id)
-
-    event = %{
-      user: %{
-        userId: to_string(user.id),
-        username: user.name,
-        displayName: user.name,
-        clanId: user.clan_id,
-        countryCode: user.country,
-        status: :menu,
-        party: party_state_to_tachyon(sess_state.party),
-        invitedToParties: Enum.map(sess_state.invited_to_parties, &party_state_to_tachyon/1),
-        friendIds: Enum.map(friends, fn %{userId: uid} -> uid end),
-        outgoingFriendRequest: outgoing,
-        incomingFriendRequest: incoming,
-        ignoreIds: [],
-        roles: convert_teiserver_roles_to_tachyon(user.roles)
-      }
-    }
+    event = build_user_self_event(user, sess_state)
 
     {:event, "user/self", event, state}
   end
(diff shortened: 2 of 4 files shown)


lib/teiserver/player/session.ex
```diff
@@ -1340,4 +1340,15 @@ defmodule Teiserver.Player.Session do
     [current | invited_to] = Task.await_many(tasks)
     {current, invited_to}
   end
+
+  @doc """
+  Sends a user/self event to a user when their roles are updated
+  """
+  @spec send_admin_role_update(T.user()) :: :ok
+  def send_admin_role_update(user) do
```
geekingfrog: Good to have this function, however it should sit more toward the top of the file. You can see [a big comment block](https://github.com/geekingfrog/teiserver/blob/3165b334f832a50d738ef982a1801f62818d572a/lib/teiserver/player/session.ex#L119-L123) for the public api of the module. This should be close to other user related functions, like `get_user_info`.

The message should also be somewhat namespaced like `{:user, :role_updated}` or something like that.
I am not a fan of the name, we shouldn't really care about how the roles were updated, just that they were.

`def update_user_roles(user_id, roles), do:` is probably better, and put the roles in the message directly.

lib/teiserver/player/session.ex
```diff
@@ -1340,4 +1340,15 @@ defmodule Teiserver.Player.Session do
...
+    case Player.SessionRegistry.lookup(user.id) do
+      nil -> :ok
+      session_pid -> GenServer.cast(session_pid, {:admin_role_update, user})
+    end
```
geekingfrog: You don't need the lookup+cast, `GenServer.cast(via_tuple(user.id), ...)` does the same thing, it's how every other function in this module operate.

lib/teiserver/player/tachyon_handler.ex
```diff
@@ -168,6 +150,12 @@ defmodule Teiserver.Player.TachyonHandler do
     {:event, "user/updated", event, state}
   end
 
+  def handle_info({:admin_role_update, updated_user}, state) do
+    sess_state = %{party: state.party, invited_to_parties: state.invited_to_parties}
```
geekingfrog: why are you taking a subset of the state and calling it `sess_state`? I find that more confusing than anything, and you can't really type that without adding more type alias. Just pass the entire state.

lib/teiserver_web/controllers/admin/user_controller.ex
```diff
@@ -307,6 +309,10 @@ defmodule TeiserverWeb.Admin.UserController do
           {:ok, user} ->
             Account.decache_user(user.id)
 
+            if roles_changed do
+              Teiserver.Player.Session.send_admin_role_update(user)
```
geekingfrog: This function to notify of a change of role should live in the top level context: `Teiserver.Player`. Modules that are not under this module should not directly access internals of the context.
This should read like `Teiserver.Player.update_roles(user.id, user.roles)` or something similar, and then a `defdelegate` to the correct internal module if required.

lib/teiserver_web/controllers/admin/user_controller.ex
```diff
@@ -265,6 +265,8 @@ defmodule TeiserverWeb.Admin.UserController do
       |> Enum.reject(&(&1 == nil))
       |> Enum.uniq()
 
+    roles_changed = new_roles != user.roles
```
geekingfrog: kind of a nitpick, but this should use a set comparison because the order is irrelevant.

lib/teiserver/player/session.ex
```diff
@@ -328,6 +328,14 @@ defmodule Teiserver.Player.Session do
       %{status: :offline}
   end
 
+  @doc """
+  Sends a user/self event to a user when their roles are updated
+  """
+  @spec update_user_roles(T.userid(), [String.t()]) :: :ok
+  def update_user_roles(user_id, roles) do
+    GenServer.cast(via_tuple(user_id), {:user, :role_updated, roles})
```
geekingfrog: prefer `{:user, {:role_updated, roles}` to keep consistent with other message structure in the module.
The idea is that maaaaybe at one point we could split this module into different handlers, and we can dispatch on the "namespace" of the message. I'm not too sure about that, but in the meantime, let's keep the same message structure anyway.

lib/teiserver/player/session.ex
```diff
@@ -1305,8 +1318,7 @@ defmodule Teiserver.Player.Session do
       # we get the same struct every time, possibly with nil values for some keys
       # but in the meantime, just defensively get the country
       country: Map.get(user, :country, "??"),
-      status: status,
-      roles: Teiserver.Player.TachyonHandler.convert_teiserver_roles_to_tachyon(user.roles)
+      status: status
```
geekingfrog: I would keep the user roles in the state. We are going to need them when joining lobbies.

wrath625: this removal was what was causing the test failures

lib/teiserver/player/session.ex
```diff
@@ -964,6 +972,11 @@ defmodule Teiserver.Player.Session do
     {:noreply, state}
   end
 
+  def handle_cast({:user, {:role_updated, roles}}, state) do
+    send_to_player!({:user, {:role_updated, roles}}, state)
```
geekingfrog, line 976: actually, this should also update the roles in the session state. It holds the user map, with a roles key in it, that should be updated. I missed that earlier

lib/teiserver/player/session.ex
```diff
@@ -964,6 +972,12 @@ defmodule Teiserver.Player.Session do
...
+    new_state = %{state | user: %{state.user | roles: roles}}
```
geekingfrog, line 977: This is fine, but there's an easier way on doing that: `put_in(state.user.roles, roles)` see [put_in](https://hexdocs.pm/elixir/Kernel.html#put_in/2)

geekingfrog (approved):

PR #728, opened 2025-08-21
Fix last_login db field being null

### LINKED ISSUE #543: [Bug]: last_login is null in account_users table
### Describe the Bug

last_login column is not being populated correctly in account_users table

### Reproduce the bug

It exists on integration server. In development environment if you login to chobby and then check the database you will see the last_login column is null, but the last_login_timex column is not empty.

### Screenshots

_No response_

### Additional context

The code for populated the last_login column looks like this:
```
    user = %{
      user
      | last_login: round(System.system_time(:second) / 60),
        last_login_timex: Timex.now(),
```

I'm unsure why we have both these columns. Perhaps last_login should just use Timex.now() ?

wrath625:
Fixes #543

Fixing `last_login` field always being null in DB,  also unravelling some duplication of data.

Originally we had:

DB (ie. `user.last_login`):

`last_login` :datetime  - always null
`last_login_timex` :datetime - set correctly

Cache (ie. `user.data['last_login']`:

`last_login` - unix timestamp
`last_login_timex` - datetime
`last_login_mins` - unix timestamp (same value as `last_login`)

In the new state, `last_login` populates correctly, and the data duplication is reduced

DB:
`last_login` :datettime - correctly populates
~~last_login_timex~~ - removed from all code references and replaced with `last_login`, and there is a database migration to purge it

Cache:
`last_login` - datettime
~~last_login_timex~~ removed, references updated to `last_login`
`last_login_mins` - untouched

This could be further reduced by blowing away the `last_login_mins` field and quickly computing that where needed, but I didn't go that far.

I also found a weird column mismatch in smurf_search I fixed while I was fixing the templates that used this data.

Note: as mentioned this includes a database migration to remove `last_login_timex` but if you don't run it everything should be fine as no code touches it anymore as far as I was able to grep

I'll squash commits once this is reviewed

### DIFF
--- a/lib/teiserver/account/caches/user_cache_lib.ex
+++ b/lib/teiserver/account/caches/user_cache_lib.ex
@@ -239,20 +239,6 @@ defmodule Teiserver.Account.UserCacheLib do
     data =
       CacheUser.data_keys()
       |> Map.new(fn k -> {to_string(k), Map.get(user, k, Account.default_data()[k])} end)
-      |> Map.put(
-        "last_login",
-        case Map.get(user, :last_login) do
-          dt when is_map(dt) ->
-            try do
-              div(Timex.to_unix(dt), 60)
-            rescue
-              _ -> Map.get(user, :last_login)
-            end
-
-          _ ->
-            Map.get(user, :last_login)
-        end
-      )
 
     obj_attrs =
       CacheUser.keys()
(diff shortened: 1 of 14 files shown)


wrath625:
Good call on the migration - that totally makes sense all of the existing data will null and blow everything up.

I'm absolutely unsure of the column mismatch, I figured there was a distinct possibility my local test data was missing something important.  I'll revert that and just assume prod data makes it work!

PR #730, opened 2025-08-23
Add pagination to phoenix UI pages

wrath625:
Added pagination to various pages listed below using a reusable pagination component defined in the new file `lib/teiserver_web/components/pagination_components.ex`

Additionally I needed to convert/normalize a number of the search forms from using POST to GET so that their results could be properly paginated.  This seemed inconsistent across the application anyway - GET was already being used in places.

Admin:
- admin/user list
- admin/user/matches/{id} list
- admin/user/ratings/{id} list
- admin/matches/search list
  - Converted route to `admin/matches` (since `admin/matches/search?search=true` is redundant) and updated links to this page
  - Converted search form from POST to GET to enable pagination on search results
  - Replaced (from what I could tell broken) user picker with a standard input field that searches on partial username match

Matches
- matches (/battle)
- ratings (/battle/ratings)

Moderation:
- Overwatch
  - Pagination
  - All filtering will re-query data to give a full data set based on the current filters and limit set, resetting to page 1
- Reports 
  - Pagination
  - Converted search form from POST to GET to enable pagination on search results
  - Normalized the search routing to use `search=true` like is done for actions and various other places
  - Added the section menu showing List and Search to maintain consistency with actions and bans
  - Added `type` to search clause as it was not filtering the search on this before
- Actions 
  - Pagination
  - Converted search form from POST to GET to enable pagination on search results
- Bans
  - Pagination
  - Added Search to the section menu
  - Converted search form from POST to GET to enable pagination on search results
  - Added new _search/3 function to make the existing (but unseen) search form work

Telemetry
- Infologs
  - Pagination
  - Converted search form from POST to GET to enable pagination on search results

Added in a validator which is called via a plug for non-liveview pages, or called directly in liveviews to clamp limit and page parameters to a max limit of 500, and a max page of 10,000.

### DIFF
--- a/lib/teiserver/account.ex
+++ b/lib/teiserver/account.ex
@@ -22,6 +22,12 @@ defmodule Teiserver.Account do
   @spec list_users(list) :: [User]
   defdelegate list_users(args), to: UserLib
 
+  @spec count_users() :: integer
+  defdelegate count_users(), to: UserLib
+
+  @spec count_users(list) :: integer
+  defdelegate count_users(args), to: UserLib
+
   @spec get_user!(non_neg_integer()) :: User.t()
   defdelegate get_user!(user_id), to: UserLib
 
--- a/lib/teiserver/account/libs/user_lib.ex
+++ b/lib/teiserver/account/libs/user_lib.ex
@@ -64,6 +64,18 @@ defmodule Teiserver.Account.UserLib do
     |> Repo.all()
   end
 
+  @spec count_users() :: integer
+  def count_users() do
+    count_users([])
+  end
+
+  @spec count_users(list) :: integer
+  def count_users(args) do
+    args
+    |> UserQueries.count_users()
+    |> Repo.aggregate(:count, :id)
+  end
+
   @doc """
   Gets a single user.
 
--- a/lib/teiserver/account/queries/user_queries.ex
+++ b/lib/teiserver/account/queries/user_queries.ex
@@ -16,6 +16,22 @@ defmodule Teiserver.Account.UserQueries do
     |> do_order_by(args[:order_by])
     |> query_select(args[:select])
     |> limit_query(args[:limit] || 50)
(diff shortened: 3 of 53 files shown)


wrath625:
> The issue with the reports under the Overwatch section is, that it loads 50 reports, and afterwards filter for chat/action reports. So, if you moslty have open chat reports, but want to see open action reports, there will barely any be loaded

Gotcha.  It will re-query now and return new data if you change a filter so you always get a data set the size of the result limit you set with the new pagination component.

L-e-x-o-n:
First of all, this is a fantastic feature and will help a lot. Thanks!

Deployed the PR to integration server if you want to test it there.

Generally it seems to work well but I noticed a few issues:
- The "Go to page input" doesn't work on any page
- The reports group (`moderation/overwatch`) seems to not have it's total count correct, it shows for example "Showing 1-500 of 50", I think this is the reason why the pagination component also doesn't show the page selector, only the limit dropdown.
- There should be an internal limit to how high the limit parameter will be accepted. Right now I can edit the URL and increase the limit to ten thousand and the server will try to fulfil this request.

It would be great if you could also add pagination to the admin matches page (`teiserver/admin/matches`) as well please.

wrath625:
Glad it helps!

- Fixed go to page input correctly setting a URL with the page parameter
- Fixed (I think? my local dataset for Overwatch is absolutely terrible) the 1-500 of 50 display error (I found the liveviews I added pagination to were loading data twice)
- Clamped limit to 500 and page to 10,000 using a plug and validator (validator called directly in liveviews)
- Added pagination to `admin/matches` 
  -  removed the `/search` part of the route as it was redundant when also using `?search=true`

wrath625:
I may as well remove the page limit entirely then.

L-e-x-o-n:
Everything appears to be working now, great!

PR #736, opened 2025-08-26
Generate users that can log in

geekingfrog:
Roles are soooo confusing


L-e-x-o-n:
Don't remove the roles list, add it in additions to roles in data, I know it's confusing, verification roles are especially chaotic, might also have to add "Verified" permission to be safe.

PR #748, opened 2025-08-30
Exposes prometheus metrics

geekingfrog:
Yet another way of exposing metrics (see https://github.com/beyond-all-reason/teiserver/pull/703)

We get `/metrics` served on a different port (4001 by default). After some exchanges with Marek, that's the preferred solution (vs putting the endpoint behind auth).
PromEx ships with some additional plugins that may be useful, but I haven't looked into them, like ecto and oban.

PromEx also has some nice, but not really truly important feature: export grafana dashboards. For example the one about the beam metrics:
beam-dashboard

Although, we can then export the dashboard as json and plug it elsewhere.

The custom tachyon metrics look like:
```
# HELP tachyon_player_count how many players are connected?
# TYPE tachyon_player_count gauge
tachyon_player_count 0
```

and we also get the existing spring/teiserver metrics.

### DIFF
--- a/lib/teiserver/prom_ex.ex
+++ b/lib/teiserver/prom_ex.ex
@@ -70,7 +70,8 @@ defmodule Teiserver.PromEx do
       # Plugins.Broadway,
 
       # Add your own PromEx metrics plugins
-      Teiserver.Monitoring.Tachyon
+      Teiserver.Monitoring.Tachyon,
+      Teiserver.Monitoring.Spring
     ]
   end
 
--- a/lib/teiserver/monitoring/spring.ex
+++ b/lib/teiserver/monitoring/spring.ex
@@ -1,0 +1,11 @@
+defmodule Teiserver.Monitoring.Spring do
+  use PromEx.Plugin
+
+  @impl true
+  def event_metrics(_opts) do
+    Event.build(
+      :spring_metrics,
+      Teiserver.Telemetry.metrics()
+    )
+  end
+end


NortySpock (approved):
Nice find, well done! 

I hadn't really thought in terms of changing the port number and having docker manage the networking side, that does clean things up.  Good to see that it goes though a router that you wrote rather than how #703 somehow "magically works in ways I don't understand".

This approach seems better than #703 and I'm a-ok with that. I had serious concerns about the draft I had :) 

----

Only thing I can think of, just in terms of "thoughts for the future", is that this is pull-based / scrape-based. Possibly there will be a point where y'all want Teiserver to fire telemetry information, as it happens, at a webhook (metrics, logs, traces).  It seems to me that Teiserver could handle the authentication, load-balancing, etc of all the client-facing stuff, and (as a series of "side" Tasks, fire all the relevant telemetry (possibly from lots of clients) over to some configured webhooks (e.g. VictoriaMetrics, VictoriaLogs, VictoriaTraces, and any other observability servers you set up). If designed to handle more-than-one observability server targets (webhooks), it would in theory be easy to gradually transition from one observability system to another if one falls out of favor.)

I don't think that changes the config here on this PR, but just a thought on future config-file design.

lib/teiserver/prom_ex.ex
```diff
@@ -0,0 +1,103 @@
+defmodule Teiserver.PromEx do
```
geekingfrog, line 1: Btw, this file has been automatically generated with `mix prom_ex.gen.config --datasource VictoriaMetrics`

lib/teiserver/prom_ex.ex
```diff
@@ -0,0 +1,103 @@
...
+      # {Plugins.Phoenix, router: TeiserverWeb.Router, endpoint: TeiserverWeb.Endpoint},
+      # Plugins.Ecto,
+      # Plugins.Oban,
+      # Plugins.PhoenixLiveView,
+      # Plugins.Absinthe,
+      # Plugins.Broadway,
+
+      # Add your own PromEx metrics plugins
+      Teiserver.Monitoring.Tachyon,
+      Teiserver.Monitoring.Spring
```
geekingfrog, line 74: only these 2 lines have been added

PR #750, opened 2025-09-02
Added a discord command to find reports by action discord message id

ChocoboCoding:
I've added a simple command for the Teiserver bot that returns all reports connected to an action, and the notes of that action, as an ephemeral response, when it is run

`/findreports message_id:[id]`

### DIFF
--- a/lib/teiserver/moderation/libs/report_lib.ex
+++ b/lib/teiserver/moderation/libs/report_lib.ex
@@ -187,7 +187,6 @@ defmodule Teiserver.Moderation.ReportLib do
       where: reports.closed or not is_nil(reports.result_id)
   end
 
-
   def _search(query, :has_discord_message_id, true) do
     from reports in query,
       where: not is_nil(reports.discord_message_id)
@@ -211,7 +210,6 @@ defmodule Teiserver.Moderation.ReportLib do
   def _search(query, :type, "actions") do
     from reports in query,
       where: reports.type == "actions"
-
   end
 
   @spec order_by(Ecto.Query.t(), String.t() | nil) :: Ecto.Query.t()


lib/teiserver/bridge/commands/findreports_command.ex
```diff
@@ -0,0 +1,113 @@
...
+              )
+              |> Repo.one()
+
+            if action == nil do
+              "Please provide a valid message id"
+            else
+              reports =
+                from(r in Moderation.Report,
+                  where: r.result_id == ^action.id
+                )
```
L-e-x-o-n: Use `Moderation.list_reports` instead e.g.
```ex
Moderation.list_reports(
    search: [result_id: action.id],
    order_by: "Newest first"
)
```

lib/teiserver/bridge/commands/findreports_command.ex
```diff
@@ -0,0 +1,113 @@
...
+      if message_id_str == nil do
+        "Please provide a message id"
+      else
+        case Integer.parse(message_id_str) do
+          {message_id, ""} ->
+            action =
+              from(a in Moderation.Action,
+                where: a.discord_message_id == ^message_id
+              )
+              |> Repo.one()
```
L-e-x-o-n: Use `Moderation.getAction/1` instead, something like `Moderation.getAction(search: [discord_message_id: message_id])`.

lib/teiserver/bridge/commands/findreports_command.ex
```diff
@@ -0,0 +1,113 @@
...
+
+            if action == nil do
+              "Please provide a valid message id"
+            else
```
L-e-x-o-n: Not necessarily something that needs fixing, but it might be more informative to say that the bot was unable to find the action, which could happen if for example the action was deleted even if the message ID is valid.

lib/teiserver/bridge/commands/findreports_command.ex
```diff
@@ -0,0 +1,113 @@
...
+              reports =
+                from(r in Moderation.Report,
+                  where: r.result_id == ^action.id
+                )
+                |> Repo.all()
+
+              report_links =
+                reports
+                |> Enum.filter(&(&1.discord_message_id != nil))
```
L-e-x-o-n: Not a big issue here, but the query could be edited to only return reports with `discord_message_id``.

lib/teiserver/bridge/commands/findreports_command.ex
```diff
@@ -0,0 +1,113 @@
...
+                Enum.join(content, "\n")
+              end
+            end
+
+          :error ->
+            "Please provide a valid message id"
+        end
+      end
+
+    Communication.new_interaction_response(content, 64)
```
L-e-x-o-n: What is this magic 64 here? Use a constant instead.

lib/teiserver/bridge/discord_bridge_bot.ex
```diff
@@ -313,7 +335,8 @@ defmodule Teiserver.Bridge.DiscordBridgeBot do
           []
         else
           from(r in Moderation.Report,
-            where: r.match_id == ^report.match_id and r.type == ^report.type
+            where: r.match_id == ^report.match_id and r.type == ^report.type,
+            order_by: [asc: r.inserted_at]
```
L-e-x-o-n: I missed this the first time, but this too should be `Moderation.list_reports`.

ChocoboCoding: Done

PR #751, opened 2025-09-03
Update tachyon_setup to include maps for local environment

### LINKED ISSUE #729: [Feature]: setup map metadata locally for tachyon matchmaking
### Describe the Feature

ffs template

### Expected Behavior

When setting up a local teiserver, in order to play matchmaking, there needs to be a list of map metadata. This is currently missing from the task `teiserver.tachyon_setup`.

The module `Teiserver.Tachyon.Tasks.SetupAssets` should export a function `ensure_maps` that make sure the local DB is seeded with some valid maps.

Running `mix teiserver.setup_tachyon` should result in some map metadata saved into the local DB.

For reference, the json blob pushed to the admin API to set the list is at The json blob pushed to the admin API can be found at https://maps-metadata.beyondallreason.dev/latest/teiserver_maps.validated.json
The task should not read from this URL though. there should be a local copy and insert them directly to the DB without going through the API.

### Screenshots

_No response_

### Additional context

_No response_

wrath625:
Closes #729 

Adds `maps.json` with 7 maps for different queue sizes (I manually added in the matchmakingQueues list values from the data on https://www.beyondallreason.info/maps since there is no data for queues besides 1v1 in the the [teiserver maps json](https://maps-metadata.beyondallreason.dev/latest/teiserver_maps.validated.json))

Adds `ensure_maps` to `tachyon_setup` to add these maps into the database during setup


lib/teiserver/tachyon/tasks/setup_assets.ex
```diff
@@ -71,4 +71,65 @@ defmodule Teiserver.Tachyon.Tasks.SetupAssets do
...
+    %{
+      "spring_name" => map_json["springName"],
+      "display_name" => map_json["displayName"],
+      "matchmaking_queues" => map_json["matchmakingQueues"] || [],
+      "thumbnail_url" => map_json["thumbnail"],
+      "startboxes_set" => map_json["startboxesSet"] || [],
+      "modoptions" => map_json["modoptions"] || %{}
+    }
+  rescue
+    _ -> nil
```
geekingfrog: not a fan of swallowing the error whole. It's a cli tool, let it crash loudly and we can then fix it quickly, instead of wasting time wondering why such and such map isn't loaded.

lib/teiserver/tachyon/tasks/setup_assets.ex
```diff
@@ -71,4 +71,65 @@ defmodule Teiserver.Tachyon.Tasks.SetupAssets do
...
+      {:error, {reason, changeset}} -> {:error, {:update, reason, changeset}}
+      {:error, reason} -> {:error, {:update, reason, nil}}
+    end
+  end
+
+  defp load_maps_from_json do
+    json_path = Path.join([Application.app_dir(:teiserver), "priv", "data", "maps.json"])
+
+    with {:ok, content} <- File.read(json_path),
+         {:ok, %{"maps" => maps}} <- Jason.decode(content) do
```
geekingfrog: I would use the `!` version for these 2 functions and just assume things work.
There is no real point in putting nice error recovery and message for a cli tool.

lib/teiserver/tachyon/tasks/setup_assets.ex
```diff
@@ -71,4 +71,65 @@ defmodule Teiserver.Tachyon.Tasks.SetupAssets do
         {:ok, {:noop, game}}
     end
   end
+
+  def ensure_maps do
+    case Asset.get_all_maps() do
+      [] -> create_maps()
+      _existing_maps -> update_maps()
```
geekingfrog: I don't think you should treat these differently. The `update_maps` function already does the heavy lifting, just use that function. For the first run it's going to show "0 deleted, n created", which is fine and make the code simpler.

lib/teiserver/mix_tasks/tachyon_setup.ex
```diff
@@ -39,5 +39,21 @@ defmodule Mix.Tasks.Teiserver.TachyonSetup do
...
+      {:ok, {:updated, %{deleted_count: deleted, created_count: created}}} ->
+        Logger.info("Updated maps: deleted #{deleted}, created #{created}")
+
+      {:error, {:create, reason, changeset}} ->
+        Logger.error("Failed to create maps: #{reason}")
+        Logger.debug("Changeset: #{inspect(changeset)}")
+
+      {:error, {:update, reason, changeset}} ->
+        Logger.error("Failed to update maps: #{reason}")
+        Logger.debug("Changeset: #{inspect(changeset)}")
```
geekingfrog: I know the file is already using Logger, but I think I made a mistake there. Because you are in a CLI task there, I would instead use [Mix.Shell.info/error](https://hexdocs.pm/mix/Mix.Shell.html) functions.
Especially since I'm starting to change the logging severity to `:error` on some mix task because it's too noisy and often irrelevant.

geekingfrog (approved):
Don't bother for this one, but for future PR, it would be better to split commits that are purely about refactoring from commits that add features.

Here you could have done a commit to add the feature, and then a commit to switch from logger to shell output (or the other way around).

PR #752, opened 2025-09-03
Add public maps page

### LINKED ISSUE #589: [Feature]: Add a public map page
### Describe the Feature

Add a new section on the landing page (after signing up probably since there's nothing truly public) to display the list of map known to teiserver. We have that since https://github.com/beyond-all-reason/teiserver/pull/569

At the top, keep a link to the much nicer page https://www.beyondallreason.info/maps
This should be a simple list with thumbnails of all the maps and their matchmaking queues. Add an option to filter by matchmaking queue perhaps

This is mostly a quality of life improvement, it's easier to check map data than opening a sql shell to the db.

Or maybe, rework the matchmaking page, removing the existing code and putting the maps for each queue there?

### Expected Behavior

_No response_

### Screenshots

_No response_

### Additional context

_No response_

wrath625:
Closes #589

image

Adds a `/maps` page that links to the beyondallreason.info maps page at that top, and as an onclick for each card.

Couple big caveats:

This does NOT have pagination yet.  Waiting on #730 - I'll add it once that is merged.

Each map card will actually show up like this:

image

 This is from the CSS `overflow: scroll` on the `card` component which is fixed in #754

### DIFF
--- a/lib/teiserver_web/live/general/home/index.html.heex
+++ b/lib/teiserver_web/live/general/home/index.html.heex
@@ -59,6 +59,10 @@
     Matches
   </.menu_card>
 
+  <.menu_card icon={Teiserver.Asset.MapLib.icon()} icon_class="fa-solid" url={~p"/maps"}>
+    Maps
+  </.menu_card>
+
   <.menu_card icon={Teiserver.Microblog.icon()} icon_class="fa-solid" url={~p"/microblog"}>
     Microblog
   </.menu_card>
--- a/lib/teiserver_web/live/maps/index.ex
+++ b/lib/teiserver_web/live/maps/index.ex
@@ -1,0 +1,77 @@
+defmodule TeiserverWeb.MapsLive.Index do
+  use TeiserverWeb, :live_view
+  alias Teiserver.Asset
+
+  @impl true
+  def mount(_params, _session, socket) do
+    socket =
+      socket
+      |> assign(:site_menu_active, "maps")
+      |> assign(:view_colour, :default)
+      |> assign(:maps, [])
+      |> assign(:filtered_maps, [])
+      |> assign(:selected_queue, "all")
+      |> assign(:available_queues, [])
+      |> assign(:current_user, socket.assigns[:current_user])
+      |> add_breadcrumb(name: "Maps", url: "/maps")
+      |> load_maps()
+
+    {:ok, socket}
+  end
+
+  @impl true
+  def handle_params(params, _url, socket) do
+    queue = params["queue"] || "all"
+
+    socket =
+      socket
+      |> assign(:selected_queue, queue)
+      |> apply_filter(queue)
+
+    {:noreply, socket}
+  end
+
+  @impl true
+  def handle_event("filter-by-queue", %{"queue" => queue}, socket) do
+    {:noreply, push_patch(socket, to: ~p"/maps?queue=#{queue}")}
+  end
+
+  defp load_maps(socket) do
+    maps = Asset.get_all_maps()
+    available_queues = extract_unique_queues(maps)
+
+    socket
+    |> assign(:maps, maps)
+    |> assign(:available_queues, available_queues)
+    |> apply_filter(socket.assigns.selected_queue)
+  end
+
+  defp apply_filter(socket, queue) do
+    filtered_maps = filter_maps_by_queue(socket.assigns.maps, queue)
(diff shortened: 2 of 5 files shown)


lib/teiserver_web/live/maps/index.ex
```diff
@@ -0,0 +1,74 @@
...
+  end
+
+  defp extract_unique_queues(maps) do
+    maps
+    |> Enum.flat_map(fn map -> map.matchmaking_queues || [] end)
+    |> Enum.uniq()
+    |> Enum.sort()
+  end
+
+  defp map_display_name_to_url(display_name) do
```
geekingfrog, line 72: This function needs a comment to explain what url you are constructing. It's so that it links to the main bar map site right?

lib/teiserver_web/live/maps/index.html.heex
```diff
@@ -0,0 +1,83 @@
...
+      <div class="alert alert-info">
+        <i class="fa-solid fa-info-circle"></i> No maps found for the selected queue.
+      </div>
+    <% else %>
+      <div class="row">
+        <%= for map <- @filtered_maps do %>
+          <div class="col-sm-6 col-md-4 col-lg-3 mb-4">
+            <div
+              class="card h-100"
+              style="cursor: pointer;"
```
geekingfrog: Not a fan of using divs for links and mucking around with the pointer. It wrecks accessibility, and breaks things like "hovering shows the target url at the bottom of the screen" for example. Is there a way to make this element a proper hyperlink `<a href=...>`?

lib/teiserver_web/live/maps/index.html.heex
```diff
@@ -0,0 +1,83 @@
...
+              onclick={"window.open('https://www.beyondallreason.info/map/#{map_display_name_to_url(map.display_name)}', '_blank')"}
+            >
+              <%= if map.thumbnail_url do %>
+                <img
+                  src={map.thumbnail_url}
+                  class="card-img-top"
+                  alt={map.display_name}
+                  style="height: 150px; object-fit: cover;"
+                  onerror="this.src='/images/placeholder-map.png'"
```
geekingfrog: that url doesn't seem to exist. I don't think you should bother with the `onerror` bit, at least for now. The metadata we get from upstream are already validated I reckon.

lib/teiserver_web/live/maps/index.html.heex
```diff
@@ -0,0 +1,83 @@
+<div class="row section-menu">
+  <div class="col-md-12">
+    <div class="mb-3">
```
geekingfrog, line 9: I'd like a fairly prominent warning there to mention this is a view of the map pools for matchmaking, and it's currently under development. For the full list of maps, go to xxx.
Or something along these lines. Since this page is publicly accessible, we're bound to have some people stumbling upon it and wondering what's that about.

lib/teiserver_web/router.ex
```diff
@@ -268,6 +268,17 @@ defmodule TeiserverWeb.Router do
     end
   end
 
+  scope "/maps", TeiserverWeb.MapsLive, as: :maps do
+    pipe_through([:live_browser, :app_layout])
+
+    live_session :maps_view,
+      on_mount: [
+        {Teiserver.Account.AuthPlug, :mount_current_user}
+      ] do
```
geekingfrog: do you need this plug? What's that for?

wrath625:
I removed it from the top level nav component since we're considering it 'in development' e so it will still be route-able through the base index page but not visible everywhere.

lib/teiserver_web/live/general/home/index.html.heex
```diff
@@ -59,6 +59,10 @@
     Matches
   </.menu_card>
 
+  <.menu_card icon={Teiserver.Asset.MapLib.icon()} icon_class="fa-solid" url={~p"/maps"}>
+    Maps
+  </.menu_card>
```
geekingfrog, line 64: the link is still there

wrath625: @geekingfrog That was intended, should I remove that as well?

geekingfrog: > I removed it from the top level nav component since we're considering it 'in development'. It will still be route-able through the base index page but not visible everywhere.

I thought you wanted to remove it.

PR #754, opened 2025-09-04
Improve/fix card component CSS and enable table horizontal scrolling/responsiveness

wrath625:
The bootstrap CSS component `card` had `overflow: scroll` enabled in #257 so tables could be viewed on smaller screens, however there is a better built in way of doing this in bootstrap that saves us from the horror and agony of viewing all `card` components with these hideous scrollbars:

image

Wrapping tables with `<div class="table-responsive">` will allow tables to respond responsively to browser window size changes / mobile device screens and at a certain point enable their own ability to scroll horizontally:

image

I've added this repsonsive wrapper to both the phoenix <.table> component, as well as pretty much every other table I could find.

I also normalized the layout on a few admin pages so they share a similar look to the rest:
- OAuth applications
- Bots
- Engine & game versions
- Matchmaking

I needed to modify the public ratings page CSS slightly to allow it to be responsive.

Warning: The diff for this is a nightmare because all of the indentation changes in the template files from wrapping each table in a new `<div>`


wrath625:
Okay I rebased this to fix conflicts with pagination

PR #758, opened 2025-09-05
Add configurable limit for friends, ignores, avoids, and blocks

### LINKED ISSUE #573: [Feature]: Limit relation lists size

wrath625:
This adds a configurable limit to relationships (friend, ignore, avoid, block) that can be changed in the admin site config User Permissions tab.  It defaults to 300.

Currently the limit for each type is calculated separately, so you can have 300 friends AND 300 ignores AND 300 avoids.

Easy to change if I misunderstood the intention there.

Closes #573

### DIFF
--- a/lib/teiserver/account/libs/relationship_lib.ex
+++ b/lib/teiserver/account/libs/relationship_lib.ex
@@ -63,16 +63,19 @@ defmodule Teiserver.Account.RelationshipLib do
     })
   end
 
-  @spec ignore_user(T.userid(), T.userid()) :: {:ok, Account.Relationship.t()}
+  @spec ignore_user(T.userid(), T.userid()) ::
+          {:ok, Account.Relationship.t()} | {:error, String.t()}
   def ignore_user(from_user_id, to_user_id)
       when is_integer(from_user_id) and is_integer(to_user_id) do
-    decache_relationships(from_user_id)
+    with :ok <- check_relationship_limit(from_user_id, :ignore) do
+      decache_relationships(from_user_id)
 
-    Account.upsert_relationship(%{
-      from_user_id: from_user_id,
-      to_user_id: to_user_id,
-      ignore: true
-    })
+      Account.upsert_relationship(%{
+        from_user_id: from_user_id,
+        to_user_id: to_user_id,
+        ignore: true
+      })
+    end
   end
 
   @spec unignore_user(T.userid(), T.userid()) :: {:ok, Account.Relationship.t()}
@@ -87,28 +90,34 @@ defmodule Teiserver.Account.RelationshipLib do
     })
   end
 
-  @spec avoid_user(T.userid(), T.userid()) :: {:ok, Account.Relationship.t()}
+  @spec avoid_user(T.userid(), T.userid()) ::
+          {:ok, Account.Relationship.t()} | {:error, String.t()}
   def avoid_user(from_user_id, to_user_id)
       when is_integer(from_user_id) and is_integer(to_user_id) do
-    decache_relationships(from_user_id)
+    with :ok <- check_relationship_limit(from_user_id, :avoid) do
+      decache_relationships(from_user_id)
 
-    Account.upsert_relationship(%{
-      from_user_id: from_user_id,
-      to_user_id: to_user_id,
-      state: "avoid"
-    })
+      Account.upsert_relationship(%{
+        from_user_id: from_user_id,
+        to_user_id: to_user_id,
+        state: "avoid"
+      })
+    end
   end
 
-  @spec block_user(T.userid(), T.userid()) :: {:ok, Account.Relationship.t()}
+  @spec block_user(T.userid(), T.userid()) ::
+          {:ok, Account.Relationship.t()} | {:error, String.t()}
   def block_user(from_user_id, to_user_id)
       when is_integer(from_user_id) and is_integer(to_user_id) do
-    decache_relationships(from_user_id)
+    with :ok <- check_relationship_limit(from_user_id, :block) do
+      decache_relationships(from_user_id)
 
-    Account.upsert_relationship(%{
-      from_user_id: from_user_id,
-      to_user_id: to_user_id,
-      state: "block"
-    })
+      Account.upsert_relationship(%{
+        from_user_id: from_user_id,
+        to_user_id: to_user_id,
+        state: "block"
+      })
+    end
   end
 
   @spec reset_relationship_state(T.userid(), T.userid()) :: {:ok, Account.Relationship.t()}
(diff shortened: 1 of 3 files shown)


lib/teiserver/account/libs/relationship_lib.ex
```diff
@@ -154,6 +163,34 @@ defmodule Teiserver.Account.RelationshipLib do
...
+    max_relationships =
+      Config.get_site_config_cache("relationships.Maximum relationships per user")
+
+    current_count =
+      case relationship_type do
+        :friend -> Account.list_friend_ids_of_user(user_id) |> length()
+        :ignore -> list_userids_ignored_by_userid(user_id) |> length()
+        :avoid -> list_userids_avoided_by_userid(user_id) |> length()
+        :block -> list_userids_blocked_by_userid(user_id) |> length()
+      end
```
L-e-x-o-n, line 201: This should include invites as well, otherwise it's possible to send, for example, 500 invites and then accept them all at once to go over the limit.

wrath625: Ah good catch!

lib/teiserver/account/libs/relationship_lib.ex
```diff
@@ -154,6 +163,44 @@ defmodule Teiserver.Account.RelationshipLib do
...
+        :avoid ->
+          list_userids_avoided_by_userid(user_id) |> length()
+
+        :block ->
+          list_userids_blocked_by_userid(user_id) |> length()
+      end
+
+    if current_count >= max_relationships do
+      type_name = relationship_type |> to_string() |> Kernel.<>("s")
+      {:error, "Maximum #{max_relationships} #{type_name} reached"}
```
L-e-x-o-n: The `s` could just be added in the error reason after `#{type_name}`, right? Not a problem but seems simpler.

wrath625: That does seem much simpler haha.  I forget what I was doing with it before its final form.

lib/teiserver/account/libs/relationship_lib.ex
```diff
@@ -154,6 +163,43 @@ defmodule Teiserver.Account.RelationshipLib do
...
+  Checks if a user has reached the maximum number of relationships of a given type.
+
+  Returns `:ok` if the user is under the limit, or `{:error, reason}` if they've reached it.
+  The limit is configurable via the admin interface.
+  """
+  @spec check_relationship_limit(T.userid(), :friend | :ignore | :avoid | :block) ::
+          :ok | {:error, String.t()}
+  def check_relationship_limit(user_id, relationship_type) do
+    max_relationships =
+      Config.get_site_config_cache("relationships.Maximum relationships per user")
```
geekingfrog: To ease the testing bits, I suggest the `max_relationships` be a parameter with this default value

wrath625: Okay I think I've achieved allowing a test to pass a `max_relationships` parameter, or otherwise the function will still use the site config normally.

lib/teiserver/account/libs/relationship_lib.ex
```diff
@@ -154,6 +163,48 @@ defmodule Teiserver.Account.RelationshipLib do
...
+  - `user_id` - The user to check
+  - `relationship_type` - The type of relationship (:friend, :ignore, :avoid, :block)
+  - `max_relationships` - Optional limit override (defaults to site config value). Useful for testing or special cases.
+  """
+  @spec check_relationship_limit(T.userid(), :friend | :ignore | :avoid | :block, integer() | nil) ::
+          :ok | {:error, String.t()}
+  def check_relationship_limit(user_id, relationship_type, max_relationships \\ nil) do
+    max_relationships =
+      max_relationships ||
+        Config.get_site_config_cache("relationships.Maximum relationships per user")
```
geekingfrog: You can inline all of this into the argument declaration: `max_relationships \\ Config.get_site_config_cache("relationships.Maximum relationships per user")` it's evaluated anew every function call (not like python with the stupid shared mutable default args)

wrath625: oh nice!

L-e-x-o-n:
`ignore_user`/`avoid_user`/`block_user`  can now return an  `:error` and it's not being handled, for example `spring_user_in.ex` and the relationship pages use these functions and will currently return successful message even if ignoring/avoiding/blocking fails.

geekingfrog:
I still want some tests to make sure this actually works (and stays that way). At the very least, all the function you modified in RelationshipLib should have tests asserting error when you go over the limit.

wrath625:
Roger that, I'll add in some tests

test/teiserver/account/relationship_lib_test.exs
```diff
@@ -55,4 +56,79 @@ defmodule Teiserver.Account.RelationshipLibTest do
...
+    # Verify we're at the limit using check_relationship_limit
+    assert {:error, _} = RelationshipLib.check_relationship_limit(user1.id, :block, limit)
+
+    # Try to add another block - should fail due to limit
+    assert {:error, _} = RelationshipLib.block_user(user1.id, user4.id)
+  end
+
+  test "check_relationship_limit function works correctly" do
+    limit = 1
+    Config.update_site_config("relationships.Maximum relationships per user", limit)
```
geekingfrog, line 138: This is likely going to cause flaky tests somewhere because you are effectively modifying a global shared state without cleanup :/

I don't think there's a painless way of avoiding this problem in the general case, at least without some serious refactoring.
I suggest you start the test by getting the value and at then end with `ExUnit.Callbacks.on_exit(fn -> <restore the value there> end)` you can set this initial value back.

wrath625: I ran into a bunch of issues with ExUnit running in a different process.  Does this approach I used suffice or do you want something cleaner?

geekingfrog: What errors? Something about not finding sandbox ownership process?
If yes, then I'd suggest only putting a `Teiserver.cache_delete(:config_site_cache, key)` in the cleanup callback. I wish there was a nicer way but oh well.
If that's another issue, I need more info than just "a bunch of issues".

Your approach works but is rather unusual and doesn't compose well, I'd rather have the cleanup working.

wrath625: specifically `  ** (DBConnection.OwnershipError) cannot find ownership process for #PID<0.2229.0>.`

so yeah, process ownership

wrath625: using `cache_delete` works without error though

lib/teiserver/protocols/spring/spring_user_in.ex
```diff
@@ -47,14 +47,24 @@ defmodule Teiserver.Protocols.Spring.UserIn do
...
 
   def do_handle("decline_friend_request", from_id_str, msg_id, state) do
     from_id = int_parse(from_id_str)
-    result = Account.decline_friend_request(from_id, state.userid)
+
+    result =
+      case Account.decline_friend_request(from_id, state.userid) do
+        :ok -> :ok
+        {:error, reason} -> {:error, reason}
+      end
```
L-e-x-o-n: The case here is unnecessary.

lib/teiserver/protocols/spring/spring_user_in.ex
```diff
@@ -47,14 +47,24 @@ defmodule Teiserver.Protocols.Spring.UserIn do
 
   def do_handle("accept_friend_request", from_id_str, msg_id, state) do
     from_id = int_parse(from_id_str)
-    result = Account.accept_friend_request(from_id, state.userid)
+
+    result =
+      case Account.accept_friend_request(from_id, state.userid) do
+        :ok -> :ok
+        {:error, reason} -> {:error, reason}
+      end
```
L-e-x-o-n: The case here is unnecessary.

PR #764, opened 2025-09-10
Only use plug TeiserverWeb.Plugs.ValidatePaginationParams on controllers with pagination

L-e-x-o-n:


lib/teiserver_web/live/battles/match/ratings.ex
```diff
@@ -41,6 +41,9 @@ defmodule TeiserverWeb.Battle.MatchLive.Ratings do
   def handle_params(params, _url, socket) do
     validated = TeiserverWeb.Validators.PaginationParams.validate_params(params)
 
+    page = (params["page"] || "1") |> String.to_integer() |> max(1) |> then(&(&1 - 1))
+    limit = (params["limit"] || "100") |> String.to_integer() |> max(1)
```
geekingfrog: Is this required? It's going to crash if `page=""` for example, and I believe the `validate_params` above already does this parsing anyway.

PR #768, opened 2025-09-12
Production dockerfile

geekingfrog:
Create a dockerfile for production release.
It is copied and adapted from [the ansible-teiserver](https://github.com/beyond-all-reason/ansible-teiserver/blob/main/roles/teiserver/templates/teiserver.Dockerfile.j2) repo. Now that we have moved over from separate configuration files to env vars, the jinja templating done via ansible is no longer required so we can have a "simple" dockerfile in the repo itself and the deploy playbook can use it directly.

The other benefit is that one can run release build locally without having to configure much.
This dockerfile is not suitable for development though, we'll need more work for that.

I've tested this with a local run of the ansible playbook on a vm.

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -172,8 +172,7 @@ if use_tls? do
     force_ssl: [hsts: true],
     root: ".",
     cache_static_manifest: "priv/static/cache_manifest.json",
-    server: true,
-    version: Mix.Project.config()[:version]
+    server: true
 else
   config :teiserver, Teiserver,
     ports: [
--- a/config/config.exs
+++ b/config/config.exs
@@ -35,7 +35,8 @@ config :teiserver, TeiserverWeb.Endpoint,
   pubsub_server: Teiserver.PubSub,
   debug_errors: Config.config_env() == :dev,
   code_reloader: Config.config_env() == :dev,
-  check_origin: Config.config_env() == :prod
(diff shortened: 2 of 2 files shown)


config/runtime.exs
```diff
@@ -27,113 +27,182 @@ end
...
+  node_name: Teiserver.ConfigHelpers.get_env("TEI_NODE_NAME", "local"),
+  enable_managed_lobbies: true,
+  user_agreement:
+    "A verification code has been sent to your email address. Please read our terms of service at https://#{domain_name}/privacy_policy and the code of conduct at https://www.beyondallreason.info/code-of-conduct. Then enter your six digit code below if you agree to the terms."
+
+config :teiserver, Teiserver.Repo,
+  hostname: Teiserver.ConfigHelpers.get_env("TEI_DB_HOSTNAME", "localhost"),
+  username: Teiserver.ConfigHelpers.get_env("TEI_DB_USERNAME", "teiserver_dev"),
+  password: Teiserver.ConfigHelpers.get_env("TEI_DB_PASSWORD", "123456789"),
+  database: Teiserver.ConfigHelpers.get_env("TEI_DB_NAME", "teiserver_dev"),
```
geekingfrog: all the `TEI_DB_*` vars are set in all environment, these are the default for local dev.

PR #769, opened 2025-09-12
Disallow reports with empty extra_text field

ChocoboCoding:
Disables the `SUBMIT REPORT` button when for reports, when the `extra_text` field is empty.
Rewording of the first sentence on the page.

Still a draft until approval of Moderation team of the changes


ChocoboCoding:
Got approval

PR #770, opened 2025-09-13
Vendor font awesome

geekingfrog:
Since we are only using the free versions, we can vendor the dependencies directly, it simplifies the setup.


lib/teiserver_web/components/nav_components.ex
```diff
@@ -290,8 +290,7 @@ defmodule TeiserverWeb.NavComponents do
 
     ~H"""
     <.link navigate={@url} class={"btn btn-outline-#{@bsname} #{@active_class}"}>
-      <Fontawesome.icon :if={@icon} icon={@icon} style={if @active, do: "solid", else: "regular"} />
-      &nbsp; {render_slot(@inner_block)}
+      <Fontawesome.icon :if={@icon} icon={@icon} style="solid" /> &nbsp; {render_slot(@inner_block)}
```
L-e-x-o-n, line 293: Why remove different style on `if @active` ?

geekingfrog: Because the free version doesn't have the "regular" styling for most of the icons.
to give you an idea:
2025-09-16-420x423-scrot

I think it's better to have free icons, even if we lose this nicety.

PR #777, opened 2025-09-16
Address mix credo issues

L-e-x-o-n:


config/config.exs
```diff
@@ -122,6 +122,8 @@ config :logger, :info_log,
   metadata: metadata,
   level: :info
 
+config :logger, :console, metadata: metadata
```
geekingfrog: Is that the same as `LoggerBackends.Console`? I think I'd rather have an additional bit in `.credo.exs` to [configure this check](https://hexdocs.pm/credo/Credo.Check.Warning.MissedMetadataKeyInLoggerConfig.html#module-check-specific-parameters) so that it finds the correct metadata.

geekingfrog (approved):

PR #783, opened 2025-09-18
Exclude mix tasks from compilation in prod env

geekingfrog:
Now that we have fatal compiler warnings, this causes the production
build to fail when some dependencies are not available in the prod env.
For example, HTTPoison used in one task.
Since mix isn't available on the prod servers, we can disable the
compilation of theses tasks.


L-e-x-o-n:
`mix` is not present but the tasks can still be ran through remote IEx shell.

PR #784, opened 2025-09-18
Added findreport mode functionality

TheFutureKnight:
- Findreport slash command can now search by Report ID
- Using the Mode argument, switches logic
- Exported some mutual lines into separate helper functions

### DIFF
--- a/lib/teiserver/bridge/commands/findreports_command.ex
+++ b/lib/teiserver/bridge/commands/findreports_command.ex
@@ -96,7 +96,7 @@ defmodule Teiserver.Bridge.Commands.FindreportsCommand do
         end
 
       if content == [] do
-        "No Report Link or Action Link/Notes available"
+        "No Report Link available"
       else
         Enum.join(content, "\n")
       end
--- a/lib/teiserver/bridge/commands/findreports_command.ex
+++ b/lib/teiserver/bridge/commands/findreports_command.ex
@@ -22,9 +22,21 @@ defmodule Teiserver.Bridge.Commands.FindreportsCommand do
         %{
           # type3 = String
           type: 3,
-          name: "message_id",
+          name: "id",
           description: "Message ID of the actions discord message",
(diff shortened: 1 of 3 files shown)


lib/teiserver/bridge/commands/findreports_command.ex
```diff
@@ -34,73 +46,113 @@ defmodule Teiserver.Bridge.Commands.FindreportsCommand do
...
+      end
+
+    content = report_link ++ action_link ++ notes
+    if content == [] do
+      "No Report Link or Action Link/Notes available"
+    else
+      Enum.join(content, "\n")
+    end
+  else
+    :error -> "Please provide a valid report id"
```
geekingfrog: `Integer.parse/2` can also return `{parsed_int, remainder}` like in `Integer.parse("123 coucou")` returns `{123, " coucou"}`
you should handle that, or decide you don't care about any trailing string and change the pattern match on line 88 with `{report_id, _rest} <- ...`

lib/teiserver/bridge/commands/findreports_command.ex
```diff
@@ -34,73 +46,113 @@ defmodule Teiserver.Bridge.Commands.FindreportsCommand do
...
+        ["**Attached Reports:**"] ++ report_links
+      end ++ notes
+
+    if content == [] do
+      "No attached Reports with links, or Notes"
+    else
+      Enum.join(content, "\n")
+    end
+  else
+    :error -> "Please provide a valid message id"
```
geekingfrog: ditto wrt `Integer/parse`

geekingfrog (commented):
Small comment but otherwise seems good.
I am not sure how to test that on my end though :/

When merging (if I don't do it), the commits should be squashed as their history isn't particularly useful. Though the one about refactoring could be extracted out, doing so is likely going to be more effort than it's worth.

PR #790, opened 2025-09-21
Live lobby kill switch

geekingfrog:


NortySpock:
Seems like a reasonable temporary solution! I appreciate the video explaining how to use the feature.

PR #800, opened 2025-09-25
Fix live chat internal server error

### LINKED ISSUE #778: [Bug]: Opening Live Chat `battle/lobbies/chat/{lobby_id}` fails
### Describe the Bug

Opening any "Live chat" for running lobby throws internal error.

### Reproduce the bug

Open any lobby chat. Looking at teiserver logs is not as useful because the stacktrace is truncated because it prints a very large object (with full user data!!) into logs.

### Screenshots

_No response_

### Additional context

_No response_

wrath625:
Closes #778

Fixed PlayerListComponent.update_many/1 to properly handle {assigns, socket} tuple format


L-e-x-o-n:
Thanks!

PR #801, opened 2025-09-25
Fix MatchMonitorServer crash on nil lobby state and add to supervision

### LINKED ISSUE #774: [Bug]: MatchMonitorServer terminated because Battle.get_combined_lobby_state(lobby_id) returned nil
### Describe the Bug

**Short version, per Lexon:**   Battle.get_combined_lobby_state(lobby_id) is the right hand side value and is nil. This terminates the MatchMonitorServer which causes other problems because it's responsible for handling match starting, ending, messages etc.

**Background:**

Somewhat before 2025-09-13 03:35 UTC (imprecise estimate from Discord chat logs, better information may be available), all lobbies listed in Chobby, in prod, became inactive (no players listed, and no lobby could be joined) Many other users reported issues in the chat and in discord. (recovered around 2025-09-13 05:40 UTC due to Teiserver reboot, according to the discord notification)

Per Discord conversation with Lexon, the following stack trace points towards the source of the issue: 

```
2025-09-13 02:38:39.500 [error] request_id=MatchMonitorServer pid=<0.6211.0>  GenServer #PID<0.6211.0> terminating
** (MatchError) no match of right hand side value: nil
    (teiserver 0.1.0) lib/teiserver/battle/libs/match_lib.ex:74: Teiserver.Battle.MatchLib.match_from_lobby/1
    (teiserver 0.1.0) lib/teiserver/battle.ex:304: Teiserver.Battle.start_match/1
    (teiserver 0.1.0) lib/teiserver/battle/servers/match_monitor_server.ex:74: Teiserver.Battle.MatchMonitorServer.handle_info/2
    (stdlib 5.2.3.5) gen_server.erl:1095: :gen_server.try_handle_info/3
    (stdlib 5.2.3.5) gen_server.erl:1183: :gen_server.handle_msg/6
    (stdlib 5.2.3.5) proc_lib.erl:241: :proc_lib.init_p_do_apply/3
Last message: {:new_message_ex, 247952, "autohosts", "* Launching game..."}
```

_Lexon_: 
> This seems to be the start of the issues, Battle.get_combined_lobby_state(lobby_id) is the right hand side value and is nil. This terminates the MatchMonitorServer which causes other problems because it's responsible for handling match starting, ending, messages etc.

BlackMelon added: 

> Hum, tbh this is a deeper problem. There is already a restart directive if the entire server die, that is, the beam keel over.
But this is only a (beam) process. Erlang has tools to restart the process or escalate the failure, but this system wasn't really designed with these in mind, and if a part dies, it impact many other parts without a clear path for recovery. 
We could improve that, but it's a non trivial amount of effort tbh. That can be something interesting for someone who wants to learn more about the beam though.

Lexon noted: "MatchMonitorServer should be able to restart if it dies already"

**Expected behavior:**
(a) (required) Some sort of automated test that mimics the issue and validates the recovery
(b) (required) MatchMonitorServer exists in supervision tree that can recover from this sort of issue and does not cause downstream issues due to being unavailable or in an invalid state. 
(c) (optional) better logging in case this happens again to troubleshoot what is causing  `Battle.get_combined_lobby_state(lobby_id)` to return nil

### Reproduce the bug

It seems like a test that spins up MatchMonitorServer and then injects a `nil` output from Battle.get_combined_lobby_state(lobby_id)  should reproduce the issue. 

### Screenshots

_No response_

### Additional context

_No response_

wrath625:
- Handle nil returns from Battle.get_combined_lobby_state in MatchLib.match_from_lobby
- Add error handling for nil client lookups in MatchMonitorServer
- Move MatchMonitorServer to main supervision tree for automatic restart
- Remove manual startup in favor of supervised startup
- Add test coverage for unavailable lobby scenarios

Closes #774 

Ready for the comments on this :)


lib/teiserver/battle/servers/match_monitor_server.ex
```diff
@@ -50,16 +50,25 @@ defmodule Teiserver.Battle.MatchMonitorServer do
...
-           true do
-        pid = self()
-
-        spawn(fn ->
-          :timer.sleep(1000)
-          send(pid, :begin)
-        end)
+      if Application.get_env(:teiserver, Teiserver)[:test_mode] == true do
+        # Skip initialization in test mode
+        %{}
```
geekingfrog: what? This is **deeply** fishy.
To be fair, the entire thing about waiting for a global flag to indicate the startup is "complete" is flawed and completely bypass the idea behind supervision trees. But changing this behaviour like that just for test is also not right.

test/teiserver/battle/match_monitor_test.exs
```diff
@@ -56,9 +51,31 @@ defmodule Teiserver.Battle.MatchMonitorTest do
     result = _recv_until(socket)
 
     assert result =~
-             "SAYPRIVATE AutohostMonitor endGameData eyJrZXkiOiJUaGlzIGlzIGEgdmVyeSBsb25nIHN0cmluZyIsInN1YiI6eyJrZXkiOiJ"
+             "SAYPRIVATE AutohostMonitor endGameData eyJzdWIiOnsic3ViIjp7InN1YiI6eyJzdWIiOnsic3ViIjp7InN1YiI6eyJzdWIiOns"
```
geekingfrog: why do you need to change these strings?

test/teiserver/battle/match_monitor_test.exs
```diff
@@ -56,9 +51,31 @@ defmodule Teiserver.Battle.MatchMonitorTest do
...
+    # Get the MatchMonitorServer PID
+    monitor_pid = Teiserver.Battle.MatchMonitorServer.get_match_monitor_pid()
+    assert monitor_pid != nil
+
+    # Send a launch message for a non-existent lobby
+    # This simulates the exact scenario from the bug report
+    send(monitor_pid, {:new_message, 99999, "autohosts", "* Launching game..."})
+
+    # Give the server a moment to process the message
+    :timer.sleep(100)
```
geekingfrog: Don't do that, there are polling function under `test/support/polling` instead.

lib/teiserver/application.ex
```diff
@@ -156,6 +156,10 @@ defmodule Teiserver.Application do
         # System throttle
         {Teiserver.Account.LoginThrottleServer, name: Teiserver.Account.LoginThrottleServer},
 
+        # Match Monitor
+        {Teiserver.Battle.MatchMonitorServer,
+         name: Teiserver.Battle.MatchMonitorServer, restart: :permanent},
```
geekingfrog: This is useless, this server is already started, [under a supervisor](https://github.com/geekingfrog/teiserver/blob/1177af13b219db36accfd170f39c3c4a05d92097/lib/teiserver/battle/servers/match_monitor_server.ex#L15-L21).
I don't know why the existing process requires be started after the fact, but changing it like that seems a bit rash.

wrath625:
> The bit about handling a nil return value is ok, but everything around the supervisor is wrong. The existing process is already under a supervisor. The issue is that, upon restart, it also reset its state, and so it becomes desync with the actual lobbies.
> 
> Fixing this issue is probably quite hard, and may not be worth it.

Ok sounds like I should just handle the nil return so it doesn't crash and otherwise leave everything else the way it was since its probably well beyond my understanding to change properly?

test/teiserver/battle/match_monitor_test.exs
wrath625: It was failing locally.  I reverted it and it still fails locally, but passes when pushed to github?  shrug

geekingfrog: See the tag: `@tag :needs_attention` ? These tests are disabled on CI. `mix test --exclude needs_attention`
I disabled a ton of them when I setup the CI because too much bitrot. If it's for old tests regarding spring protocol I don't care too much since they're slated for deletion.
For new features related to tachyon or the website, we definitely should not have flaky tests like that though.

wrath625: That explains that, awesome!

PR #803, opened 2025-09-26
Responsive Navbar & BAR logo

wrath625:
FIxing the navbar to be properly responsive for small display sizes

- Added responsiveness to navbar, when collapsed the menu is accessible through the hamburger icon
- Added the BAR logo.svg (stolen from bar-lobby) to replace the robot and the BAR text
- Renamed `My account` and `Account` to `Profile` to reflect the actual url route and avoid 'my account' collapsing weirdly to two lines

Uncollapsed:
image

Collapsed:

image

Note:  I made a css change so a `mix assets.deploy` will need to be run to recompile this so the dropdown menus properly expand in the collapsed nav.

### DIFF
--- a/assets/scss/mdb/_navbar.scss
+++ b/assets/scss/mdb/_navbar.scss
@@ -30,6 +30,26 @@
   position: absolute;
 }
 
+// Only apply absolute positioning when navbar is expanded (not collapsed)
+@media (min-width: 992px) {
+  .navbar-nav .dropdown-menu {
+    position: absolute;
+  }
+}
+
+// When navbar is collapsed, let dropdowns flow naturally
+@media (max-width: 991px) {
+  .navbar-nav .dropdown-menu {
+    position: static;
+    float: none;
+    width: auto;
+    margin-top: 0;
+    background-color: transparent;
+    border: 0;
+    box-shadow: none;
+  }
+}
+
 .navbar-light {
   .navbar-toggler-icon {
     background-image: none;
--- a/lib/teiserver_web/components/nav_components.ex
+++ b/lib/teiserver_web/components/nav_components.ex
@@ -5,6 +5,7 @@ defmodule TeiserverWeb.NavComponents do
   # import TeiserverWeb.Gettext
 
   import Teiserver.Account.AuthLib, only: [allow?: 2, allow_any?: 2]
+  import Phoenix.HTML, only: [raw: 1]
 
   use Phoenix.VerifiedRoutes,
     endpoint: TeiserverWeb.Endpoint,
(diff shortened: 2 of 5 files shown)


mix.exs
```diff
@@ -38,7 +38,7 @@ defmodule Teiserver.MixProject do
     [
       mod: {Teiserver.Application, []},
       included_applications: [:nostrum],
-      extra_applications: [:logger, :runtime_tools, :os_mon, :iex] ++ nostrum_extras
+      extra_applications: [:logger, :runtime_tools, :tools, :os_mon, :iex] ++ nostrum_extras
```
geekingfrog: What is this thing? It looks like an artifact from another unrelated commit.

assets/scss/mdb/_navbar.scss
```diff
@@ -30,6 +30,26 @@
...
 
+// Only apply absolute positioning when navbar is expanded (not collapsed)
+@media (min-width: 992px) {
+  .navbar-nav .dropdown-menu {
+    position: absolute;
+  }
+}
+
+// When navbar is collapsed, let dropdowns flow naturally
+@media (max-width: 991.98px) {
```
geekingfrog: you can use `width < 992px` instead of this arbitrary looking limit (which doesn't work if the width is 991.99)

lib/teiserver_web/components/nav_components.ex
```diff
@@ -44,25 +44,32 @@ defmodule TeiserverWeb.NavComponents do
     <nav class="navbar navbar-expand-lg m-0 p-0" id="top-nav">
       <!-- Container wrapper -->
       <div class="container-fluid">
-        <!-- Collapsible wrapper -->
+        <!-- Navbar brand -->
+        <a class="navbar-brand" href="/">
+          <img src={~p"/images/logo/logo.svg"} alt="BAR Logo" style="height: 24px; width: auto;" />
+        </a>
+        
+    <!-- Navbar toggler button -->
```
geekingfrog: useless comment, the class right after says the same thing

lib/teiserver_web/components/nav_components.ex
```diff
@@ -44,25 +44,32 @@ defmodule TeiserverWeb.NavComponents do
...
+          data-bs-toggle="collapse"
+          data-bs-target="#navbarSupportedContent"
+          aria-controls="navbarSupportedContent"
+          aria-expanded="false"
+          aria-label="Toggle navigation"
+        >
+          <span class="navbar-toggler-icon"></span>
+        </button>
+        
+    <!-- Collapsible wrapper -->
```
geekingfrog: likewise

assets/scss/mdb/_navbar.scss
wrath625: apparently the recommended way for this is to just set it to 991 to stay consistent with bootstrap

wrath625:
thanks for the mix.exs catch, not sure how that got pulled in :\

lib/teiserver_web/components/nav_components.ex
```diff
@@ -44,25 +44,30 @@ defmodule TeiserverWeb.NavComponents do
...
-          </a>
-          <!-- Left links -->
-          <ul class="navbar-nav me-auto mb-2 mb-lg-0">
+          <!-- Left navigation -->
+          <ul class="navbar-nav me-auto">
             <.top_nav_item text="Home" active={@active == "central_home"} route={~p"/"} />
 
             <.top_nav_item
-              text="My account"
+              text="Profile"
```
geekingfrog: meh, not a fan of that, though it's not a big deal. If you don't want a line break between the two, replace the space with a non breakable space `&nbsp`

wrath625: yeah of course

PR #804, opened 2025-09-27
TOTP based 2FA

### LINKED ISSUE #103: Implement 2FA option for accounts

ChocoboCoding:
This adds TOTP based 2FA

Closes #103

# 2FA setup

Under `/teisersver/account` you have a new button to get to the 2FA settings
image

On that page, it shows the 2FA status and different buttons, depending if the 2FA status is active or inactive
image

If it's inactive, you can enable 2FA, by adding it to the authenticator application of your choice and entering a valid OTP to verify that you've actually added it
image

When you successfully add it, you get an info popup telling you that you've updated TOTP (text to be updated).
If 2FA is enabled, you get 2 different buttons, one to disable 2FA, and one to reset the secret (Text to be changed to `Reset 2FA secret`). 
image

# OTP check on login

If you have 2FA enabled for your account, you get asked to input a valid OTP, puting in valid login credentials (email + password)
image

### DIFF
--- a/lib/teiserver/account.ex
+++ b/lib/teiserver/account.ex
@@ -106,9 +106,6 @@ defmodule Teiserver.Account do
   @spec get_user_secret(integer) :: binary | nil
   defdelegate get_user_secret(user_id), to: TOTPLib
 
-  @spec get_last_used(integer) :: :inactive | String.t()
-  defdelegate get_last_used(user_id), to: TOTPLib
-
   @spec set_secret(integer, binary) :: {:ok, TOTP.t()} | {:error, Ecto.Changeset.t()}
   defdelegate set_secret(user_id, secret), to: TOTPLib
 
--- a/test/teiserver/account/user_totp_lib_test.ex
+++ b/test/teiserver/account/user_totp_lib_test.ex
@@ -14,10 +14,6 @@ defmodule Teiserver.Account.TOTPLibTest do
     %{user_without_totp: user_without_totp, user_with_totp: user_with_totp, secret: secret}
   end
 
-  #  defp datetimes(_context) do
-  #    %{datetime_valid: ~U[1970-01-01 00:00:30Z], datetime_invalid: ~U[1970-01-01 00:01:00Z]}
-  #  end
-
   # ----------------------------------------
   # Database write functions
   # ----------------------------------------


lib/teiserver/account/plugs/auth_plug.ex
```diff
@@ -29,15 +29,24 @@ defmodule Teiserver.Account.AuthPlug do
       Logger.metadata([user_id: user.id] ++ Logger.metadata())
     end
 
+    #    IO.inspect(user)
```
geekingfrog: comment to remove

lib/teiserver/account/schemas/user_totp.ex
```diff
@@ -0,0 +1,24 @@
+defmodule Teiserver.Account.TOTP do
+  use TeiserverWeb, :schema
+
+  @primary_key {:user_id, :id, autogenerate: false}
+  @foreign_key_type :id
+
+  @type t :: term()
```
geekingfrog: That's not a great type, you should create the correct typing info based on the schema [like there for example](https://github.com/beyond-all-reason/teiserver/blob/master/lib/teiserver/asset/schemas/map.ex#L4-L11)

ChocoboCoding: Changed to a proper type

lib/teiserver/account/plugs/auth_plug.ex
ChocoboCoding: Removed

lib/teiserver/account/schemas/user_totp.ex
```diff
@@ -0,0 +1,30 @@
...
+          updated_at: NaiveDateTime.t() | nil
+        }
+
+  @primary_key {:user_id, :id, autogenerate: false}
+  @foreign_key_type :id
+
+  schema "teiserver_account_user_totps" do
+    belongs_to :user, Teiserver.Account.User, define_field: false, type: :id
+    field :secret, :binary
+    field :last_used, :string, default: nil
```
geekingfrog: This is supposed to be a date, store it as such, not as a string. You can't use any date related function on a string, and you can also store strings that aren't date with this data type.

lib/teiserver/account/libs/user_totp_lib.ex
```diff
@@ -0,0 +1,146 @@
...
+        else
+          case validate_totp(totp.secret, otp) do
+            {:ok, info} ->
+              set_last_used(user, otp)
+              {:ok, info}
+
+            {:error, :invalid} ->
+              {:error, :invalid}
+          end
+        end
```
geekingfrog: Instead of having increasingly nested blocks, you should use the [with](statement)

lib/teiserver/account/libs/user_totp_lib.ex
```diff
@@ -0,0 +1,146 @@
...
+
+  def validate_totp(secret, otp) do
+    now = System.os_time(:second)
+
+    cond do
+      NimbleTOTP.valid?(secret, otp, time: now) ->
+        {:ok, :valid}
+
+      NimbleTOTP.valid?(secret, otp, time: now - 5) ->
+        {:ok, :grace}
```
geekingfrog: Is this really necessary? I would keep the code simple and just check once against `now - 5`, perhaps with a comment to indicate it's a grace period.

lib/teiserver/account/libs/user_totp_lib.ex
```diff
@@ -0,0 +1,146 @@
...
+
+      true ->
+        {:error, :invalid}
+    end
+  end
+
+  @spec validate_last_used(String.t(), String.t()) :: boolean
+  def validate_last_used(last_used, otp) do
+    last_used == otp
+  end
```
geekingfrog, line 202: I don't think this function brings anything. It's one more indirection to understand what's going on, and I don't find the name very good. `already_used?` would be better, but considering it's a simple comparison, with exactly one call site, I would inline the whole thing.

lib/teiserver/account/libs/user_totp_lib.ex
```diff
@@ -0,0 +1,146 @@
...
+    |> Repo.update()
+  end
+
+  @spec set_secret(User.t(), binary) :: {:ok, TOTP.t()} | {:error, Ecto.Changeset.t()}
+  def set_secret(%User{} = user, secret) do
+    set_totp(user, %{user_id: user.id, secret: secret})
+  end
+
+  @spec set_last_used(User.t(), String.t()) :: {:ok, TOTP.t()} | {:error, Ecto.Changeset.t()}
+  def set_last_used(%User{} = user, last_used) do
```
geekingfrog: See comment about the data type on the schema, this should be a datetime object, not a string.

lib/teiserver_web/controllers/account/security_controller.ex
```diff
@@ -2,7 +2,7 @@ defmodule TeiserverWeb.Account.SecurityController do
   use TeiserverWeb, :controller
 
   alias Teiserver.Account
-
+  alias Teiserver.Account.{TOTP, TOTPLib}
```
geekingfrog: A controller shouldn't directly import things like that. It should only have knowledge about the context, in this case `Account`, and not reach for the internals of the context.
You need to re-export (with defdelegate) the relevant functions from `account.ex` (there are plenty of examples).
Phoenix [has a guide](https://hexdocs.pm/phoenix/contexts.html) about this topic if you want to read more.

lib/teiserver_web/controllers/account/security_controller.ex
```diff
@@ -26,6 +26,80 @@ defmodule TeiserverWeb.Account.SecurityController do
...
+    |> render("totp.html")
+  end
+
+  @spec edit_totp(Plug.Conn.t(), map()) :: Plug.Conn.t()
+  def edit_totp(conn, _params) do
+    user = Account.get_user!(conn.assigns.current_user.id)
+    {_status, secret} = TOTPLib.get_or_generate_secret(user)
+    encoded_secret = Base.encode32(secret, padding: false)
+    changeset = TOTP.changeset(%TOTP{user_id: user.id, secret: encoded_secret})
+    otpauth_uri = NimbleTOTP.otpauth_uri("BAR:#{user.name}", secret, issuer: "Beyond All Reason")
```
geekingfrog: The url generation logic should be a function in the context, and not in a controller.

lib/teiserver_web/controllers/account/security_controller.ex
```diff
@@ -26,6 +26,80 @@ defmodule TeiserverWeb.Account.SecurityController do
...
+
+      {:error, _reason} ->
+        changeset = TOTP.changeset(%TOTP{user_id: user.id, secret: totp_params["secret"]})
+
+        qr_svg =
+          totp_params["otpauth_uri"]
+          |> EQRCode.encode()
+          |> EQRCode.svg(width: 250)
+
+        conn
```
geekingfrog, line 90: You should display somehow that something went wrong there. Either a flash or a message inside the template.

lib/teiserver_web/controllers/account/session_controller.ex
```diff
@@ -1,6 +1,7 @@
 defmodule TeiserverWeb.Account.SessionController do
   use TeiserverWeb, :controller
   alias Teiserver.Account
+  alias Teiserver.Account.TOTPLib
```
geekingfrog: Same comment regarding imports (even though there are already violation of this rule later in this file, I'd rather not introduce more)

lib/teiserver_web/controllers/account/session_controller.ex
```diff
@@ -24,6 +25,7 @@ defmodule TeiserverWeb.Account.SessionController do
       conn
       |> Guardian.Plug.sign_out(clear_remember_me: true)
       |> assign(:changeset, changeset)
+      |> assign(:user, nil)
```
geekingfrog: Why is that required?

lib/teiserver_web/controllers/account/session_controller.ex
```diff
@@ -34,9 +36,57 @@ defmodule TeiserverWeb.Account.SessionController do
   def login(conn, %{"user" => %{"email" => email, "password" => password}}) do
     email = String.trim(email)
 
-    conn
-    |> UserLib.authenticate_user(email, password)
-    |> login_reply(conn)
+    case UserLib.authenticate_user(conn, email, password) do
```
geekingfrog, line 37: I think I'd rather have the check about OTP inside this function, and amend its return value to indicate the user needs the additional verification when it's relevant, like `{:ok, user} | {:requires_2fa, user} | {:error, reason}`

Although this function is only used in one place, if you split the check into 2, it's too easy to, later on, forget to check for 2fa.

test/teiserver/account/user_totp_lib_test.ex
```diff
@@ -0,0 +1,192 @@
...
+      attrs
+      |> Enum.into(%{
+        email: "test#{System.unique_integer()}@example.com",
+        password: "password123",
+        name: "TestUser#{System.unique_integer()}"
+      })
+      |> Account.create_user()
+
+    user
+  end
```
geekingfrog, line 15: There is already a function for that: `Central.Helpers.GeneralTestLib.make_user` you can see [some](https://github.com/beyond-all-reason/teiserver/blob/master/test/support/tachyon.ex#L19) [examples](https://github.com/beyond-all-reason/teiserver/blob/master/test/teiserver_web/controllers/admin/o_auth_application_controller_test.exs#L8)

lib/teiserver/account/libs/user_totp_lib.ex
```diff
@@ -0,0 +1,146 @@
+defmodule Teiserver.Account.TOTPLib do
+  use TeiserverWeb, :library
+  alias Teiserver.Account.{User, TOTP}
+
+  @spec get_user_totp(User.t()) :: {:active, TOTP.t()} | {:inactive, nil}
+  defp get_user_totp(%User{id: nil}) do
+    {:inactive, nil}
+  end
```
geekingfrog: How can this happen? And also, shouldn't this be covered by the next clause, where `get_by(nil)` would return nil?

lib/teiserver/account/libs/user_totp_lib.ex
```diff
@@ -0,0 +1,146 @@
...
+
+            {:error, :invalid} ->
+              {:error, :invalid}
+          end
+        end
+    end
+  end
+
+  def validate_totp(secret, otp) do
+    now = System.os_time(:second)
```
geekingfrog: You should pass the current time as an argument, with this default value, it makes testing possible.

test/teiserver/account/user_totp_lib_test.ex
```diff
@@ -0,0 +1,192 @@
...
+      otp = NimbleTOTP.verification_code(secret)
+      assert {:ok, :valid} = TOTPLib.validate_totp(user, otp)
+      assert {:error, :used} = TOTPLib.validate_totp(user, otp)
+    end
+
+    test "returns {:error, :invalid} for OTP from 30 seconds ago", %{
+      valid_user: user,
+      old_otp: old_otp
+    } do
+      # Assuming your system allows OTPs within a 30-second window
```
geekingfrog: Useless comment, you are testing said system and asserting such things, it's even in the name of the test.

test/teiserver/account/user_totp_lib_test.ex
```diff
@@ -0,0 +1,192 @@
...
+    old_otp = NimbleTOTP.verification_code(valid_secret, time: System.os_time(:second) - 30)
+
+    {:ok,
+     valid_user: valid_user,
+     invalid_user: invalid_user,
+     valid_secret: valid_secret,
+     invalid_secret: invalid_secret,
+     totp: totp,
+     old_otp: old_otp}
+  end
```
geekingfrog, line 15: Not a fan of such big setup function. It makes things harder to follow in the tests, not knowing which bits are relevant. It's also not clear when reading the tests what are these values, especially as you go further down the file (what was that value again? why is it invalid?)

You can have multiple and smaller setup functions, scoped to `describe` blocks, there are plenty of [examples](https://github.com/beyond-all-reason/teiserver/blob/master/test/teiserver_web/controllers/admin/o_auth_application_controller_test.exs) around.

test/teiserver/account/user_totp_lib_test.ex
```diff
@@ -0,0 +1,192 @@
...
+  describe "disable_totp/1" do
+    test "removes TOTP for active user", %{valid_user: user, valid_secret: secret} do
+      {:ok, deleted_totp} = TOTPLib.disable_totp(user)
+      assert deleted_totp.secret == secret
+
+      db_totp = Repo.get_by(TOTP, user_id: user.id)
+      assert db_totp == nil
+    end
+
+    test "returns {:ok, nil} for user with no TOTP", %{valid_user: user} do
```
geekingfrog: Not a fan of the naming of these tests. It's obvious what the return value should be 2 lines below. "works for user without totp" is a better name, it describe what you are testing, not how.

test/teiserver/account/user_totp_lib_test.ex
```diff
@@ -0,0 +1,192 @@
...
+  end
+
+  describe "get_user_secret/1" do
+    test "returns secret for active user", %{valid_user: user, valid_secret: secret} do
+      assert TOTPLib.get_user_secret(user) == secret
+    end
+
+    test "returns :inactive for user with no TOTP", %{invalid_user: invalid_user} do
+      assert TOTPLib.get_user_secret(invalid_user) == :inactive
+    end
```
geekingfrog, line 129: you should also test that you get `:inactive` for valid users without the setting

geekingfrog (commented):
When entering invalid confirmation code (on the setup page), there is no feedback to tell the user so. You should put a flash like you did for the authentication flow.


If you click on "reset totp", it immediately generates a new code, and then, even if you don't confirm the code, it is still considered active and the user cannot log in anymore with the old code. I think it's a problem. It also doesn't match the behaviour of the activation path, where nothing happen until the user actually confirmed the code.

I would also like to have at least one test testing the web flow: check that a user with totp active indeed gets redirected to the correct page after submitting email&password, and another one testing that they can indeed log in after submitting the correct code.

lib/teiserver/account/schemas/user_totp.ex
ChocoboCoding: `last_used` stores the last used OTP, as a string. This makes the validate function more lightweight, as we can just compare the entered OTP against the content of last_used.
Additionally, since we allow a grace period of 5 seconds (using an OTP that is up to 5 seconds out of date, for ease of use for the user) saving the naive datetime could lead to it saving a datetime of an OTP that has not been used yet.
It's also not a security concern, as the OTP can easily be calculated via a single function from the NimbleTOTP library, so there is no difference in terms of security between those

lib/teiserver/account/libs/user_totp_lib.ex
ChocoboCoding: Yes, will change that

lib/teiserver/account/libs/user_totp_lib.ex
ChocoboCoding: If we just check against `now - 5` it can happen that it does not accept a valid secret if you enter the OTP that just became valid, as it compares it against the OTP that was valid 5 seconds ago

lib/teiserver/account/libs/user_totp_lib.ex
ChocoboCoding: I thought I would use this more often, but since it's just used once, I agree. Will inline it

lib/teiserver/account/libs/user_totp_lib.ex
ChocoboCoding: Answered why it is a string under the other comment

lib/teiserver_web/controllers/account/security_controller.ex
ChocoboCoding: I was not aware of that it shouldn't have direct imports. Will change that

lib/teiserver_web/controllers/account/security_controller.ex
ChocoboCoding: Are you talking about when they add an invalid OTP during setup?

lib/teiserver_web/controllers/account/session_controller.ex
ChocoboCoding: Can be removed, I've added it in an older version, and forgot to remove it

lib/teiserver_web/controllers/account/session_controller.ex
ChocoboCoding: I agree, will change

lib/teiserver/account/libs/user_totp_lib.ex
ChocoboCoding: It was rather to prevent that if it ever happens thatsomeone somehow tries to parse the function with id: nil, that it's already prevented

lib/teiserver_web/controllers/account/security_controller.ex
ChocoboCoding: Added the function delegations to Teiserver.Account. It still imports Teiserver.Account.TOTP

lib/teiserver_web/controllers/account/security_controller.ex
badosu: I just tested, it doesn't display anything on the screen. I got confused by it (as it seemed it worked?).

lib/teiserver_web/templates/account/security/edit_totp.html.heex
```diff
@@ -0,0 +1,50 @@
...
+          <div class="bg-gray-800 p-6 mb-4">
+            {raw(@qr_svg)}
+          </div>
+
+          <div class="mb-4">
+            <p class="text-sm text-gray-600 break-all">{@otpauth_uri}</p>
+          </div>
+
+          <div class="mb-4">
+            {label(f, :otp, "One Time Password", class: "form-label text-white")}
```
badosu, line 36: Should show some helper text, explaining what is expected here, e.g.: "Open your authenticator app and insert the one time password for Teiserver you just set up"

ChocoboCoding: I agree, I'll add that

lib/teiserver_web/controllers/account/security_controller.ex
ChocoboCoding: Added
image

lib/teiserver_web/templates/account/security/edit_totp.html.heex
ChocoboCoding: Added
image

lib/teiserver/account/libs/user_totp_lib.ex
ChocoboCoding: Added

test/teiserver/account/user_totp_lib_test.ex
ChocoboCoding: I'll rename them tomorrow

test/teiserver/account/user_totp_lib_test.ex
ChocoboCoding: I'll simplify the setup and work more with `describe`

lib/teiserver/account/libs/user_totp_lib.ex
```diff
@@ -0,0 +1,144 @@
...
+        {:error, :invalid}
+    end
+  end
+
+  @spec generate_otpauth_uri(User.t()) :: {:new | :existing, String.t()}
+  def generate_otpauth_uri(user) do
+    {status, secret} = get_or_generate_secret(user)
+    otpauth_uri = generate_otpauth_uri(user.name, secret)
+    {status, otpauth_uri}
+  end
```
L-e-x-o-n: This seems unused.

test/teiserver/account/user_totp_lib_test.ex
ChocoboCoding: Renamed the tests to be a bit more descriptive of what they're testing

test/teiserver/account/user_totp_lib_test.ex
ChocoboCoding: Setup massively simplified to only create 2 users, set TOTP for one, and using `describe` blocks now

lib/teiserver_web/templates/account/security/totp.html.heex
```diff
@@ -0,0 +1,57 @@
...
+            <%= if @totp_status == :inactive do %>
+              {link("Enable 2FA",
+                to: ~p"/teiserver/account/security/totp/edit",
+                class: "btn btn-primary"
+              )}
+            <% else %>
+              {link("Reset TOTP",
+                to: ~p"/teiserver/account/security/totp/edit",
+                class: "btn btn-primary"
+              )}
```
L-e-x-o-n, line 47: Maybe could be reduced to:
```ex
<%= 
  link(if @totp_status == :inactive, do: "Enable 2FA", else: "Reset TOTP",
    to: ~p"/teiserver/account/security/totp/edit",
    class: "btn btn-primary"
  )
%>
```

lib/teiserver/account/libs/user_totp_lib.ex
```diff
@@ -0,0 +1,144 @@
...
+    end
+  end
+
+  @spec validate_totp(binary, String.t(), integer) :: {:ok, :valid | :grace} | {:error, :invalid}
+  def validate_totp(secret, otp, time \\ System.os_time(:second)) do
+    cond do
+      NimbleTOTP.valid?(secret, otp, time: time) ->
+        {:ok, :valid}
+
+      NimbleTOTP.valid?(secret, otp, time: time - 5) ->
```
L-e-x-o-n: [NimbleTOTP.valid?](https://hexdocs.pm/nimble_totp/NimbleTOTP.html#valid?/3) accepts `:since` option - the time when the last secret was used.

While your approach works, using `:since` would move the check to the library, additionally saving the information of when the last code was used is more useful than what the code was if, for example,  we wanted to show this information to the user or we wanted to check it for debugging.

lib/teiserver_web/templates/account/security/totp.html.heex
ChocoboCoding: Can not, as it is accessing a different path now `~p/teiserver/account/security/totp/reset`, to actually reset it

lib/teiserver/account/libs/user_totp_lib.ex
ChocoboCoding: You're right, but I figured it might be used. I'll comment it for now

lib/teiserver/account/libs/user_totp_lib.ex
ChocoboCoding: I'll need to check this out

ChocoboCoding: Yes, I'll change it

ChocoboCoding: Actually, as I said above
> Additionally, since we allow a grace period of 5 seconds (using an OTP that is up to 5 seconds out of date, for ease of use for the user) saving the naive datetime could lead to it saving a datetime of an OTP that has not been used yet.

lib/teiserver/account/schemas/user_totp.ex
ChocoboCoding: Changed, as all of my points are no longer valid

lib/teiserver/account/libs/user_totp_lib.ex
ChocoboCoding: Changed

lib/teiserver/account/libs/user_totp_lib.ex
ChocoboCoding: Changed

lib/teiserver/account/libs/user_totp_lib.ex
geekingfrog: hum, good point. I'm still not convinced we need to surface the `grace` result. I think it's just simpler to treat it as `:valid` unless something higher up the stack needs to know the difference (which I don't think so)

test/teiserver_web/controllers/account/session_controller_test.exs
```diff
@@ -5,6 +5,42 @@ defmodule TeiserverWeb.Account.SessionControllerTest do
...
+    setup do
+      GeneralTestLib.conn_setup(Teiserver.TeiserverTestLib.player_permissions())
+      |> Teiserver.TeiserverTestLib.conn_setup()
+    end
+
+    test "no OTP check when no secret", %{conn: conn, user: user} do
+      conn = GeneralTestLib.login(conn, user.email)
+
+      assert redirected_to(conn) == ~p"/"
+      assert ConnTest.get_flash(conn, :info) == "Welcome back!"
```
geekingfrog: I wouldn't test that, it brings nothing for this test and is going to get annoying when we make just a copy change. Here you are only concerned about log in logic, not messaging or presentation.

test/teiserver_web/controllers/account/session_controller_test.exs
```diff
@@ -5,6 +5,42 @@ defmodule TeiserverWeb.Account.SessionControllerTest do
...
+      assert Guardian.Plug.current_resource(conn).id == user.id
+    end
+
+    test "OTP check when secret set", %{conn: conn, user: user} do
+      {_, secret} = Account.get_or_generate_secret(user)
+      Account.set_secret(user, secret)
+      conn = GeneralTestLib.login(conn, user.email)
+
+      assert redirected_to(conn) == ~p"/otp"
+      assert get_session(conn, :pending_2fa_user_id) == user.id
```
geekingfrog, line 27: Add an assert that the user is not yet logged in at that stage.

test/teiserver_web/controllers/account/session_controller_test.exs
```diff
@@ -5,6 +5,42 @@ defmodule TeiserverWeb.Account.SessionControllerTest do
...
+    end
+
+    test "OTP check for user", %{conn: conn, user: user} do
+      {_, secret} = Account.get_or_generate_secret(user)
+      Account.set_secret(user, secret)
+      otp = NimbleTOTP.verification_code(secret)
+      conn = GeneralTestLib.login_opt(conn, user, otp)
+
+      assert redirected_to(conn) == ~p"/"
+      assert ConnTest.get_flash(conn, :info) == "Welcome back!"
```
geekingfrog: same comment regarding the flash message

lib/teiserver/account/schemas/user_totp.ex
```diff
@@ -0,0 +1,32 @@
...
+          wrong_otp: integer()
+        }
+
+  @primary_key {:user_id, :id, autogenerate: false}
+  @foreign_key_type :id
+
+  schema "teiserver_account_user_totps" do
+    belongs_to :user, Teiserver.Account.User, define_field: false, type: :id
+    field :secret, :binary
+    field :last_used, :naive_datetime, default: nil
```
geekingfrog: Why `naive_datetime`? There is `utc_datetime`, and the corresponding elixir type `DateTime`. Unless you have a good reason to strip timezone info, utc should be preferred.

geekingfrog (commented):
In the future, please don't sneakily add more feature to an already massive feature. The locking mechanism can be done later on another branch.

lib/teiserver/account/libs/user_lib.ex
```diff
@@ -223,6 +223,7 @@ defmodule Teiserver.Account.UserLib do
 
   def password_reset_update_user(%User{} = user, attrs) do
     Account.recache_user(user.id)
+    TOTPLib.disable_totp(user)
```
L-e-x-o-n: Will this disable TOTP even if password reset was unsuccessful?  Is that intentional?
`password_reset` changeset might not update the password if password confirmation doesn't match.

lib/teiserver/account/libs/user_totp_lib.ex
```diff
@@ -0,0 +1,215 @@
...
+        {:error, :invalid}
+    end
+  end
+
+  @spec validate_totp(binary, String.t(), NaiveDateTime.t()) ::
+          {:ok, :valid | :grace} | {:error, :invalid | :used}
+  def validate_totp(secret, otp, time, since: last_used) do
+    if is_nil(last_used) do
+      validate_totp(secret, otp, time)
+    else
```
L-e-x-o-n: Add another `validate_totp` function that matches `since: nil` instead.

lib/teiserver/account/libs/user_totp_lib.ex
```diff
@@ -0,0 +1,215 @@
...
+        NimbleTOTP.valid?(secret, otp, time: time, since: last_used) ->
+          {:ok, :valid}
+
+        NimbleTOTP.valid?(secret, otp, time: time - @grace, since: last_used) ->
+          {:ok, :grace}
+
+        true ->
+          case validate_totp(secret, otp, time) do
+            {:error, :invalid} ->
+              {:error, :invalid}
```
L-e-x-o-n: A comment would be nice to explain why this second check is done.

lib/teiserver/account.ex
```diff
@@ -94,6 +94,34 @@ defmodule Teiserver.Account do
...
+
+  @spec disable_totp(User.t()) :: {:ok, TOTP.t() | nil} | {:error, Ecto.Changeset.t()}
+  defdelegate disable_totp(user), to: TOTPLib
+
+  @spec validate_totp(User.t() | binary, String.t()) ::
+          {:ok, :valid | :grace} | {:error, :inactive | :invalid | :used}
+  defdelegate validate_totp(user_or_secret, otp), to: TOTPLib
+
+  #  @spec generate_otpauth_uri(User.t()) :: {:new | :existing, String.t()}
+  #  defdelegate generate_otpauth_uri(user), to: TOTPLib
```
L-e-x-o-n: Why is this commented out? It doesn't seem to exist, please delete it.

lib/teiserver/account/libs/user_totp_lib.ex
```diff
@@ -0,0 +1,215 @@
...
+  @spec disable_totp(User.t()) :: {:ok, TOTP.t() | nil} | {:error, Ecto.Changeset.t()}
+  def disable_totp(%User{id: _user_id} = user) do
+    case get_user_totp(user) do
+      {:active, totp} ->
+        Repo.delete(totp)
+
+      {:inactive, _} ->
+        {:ok, nil}
+    end
+  end
```
L-e-x-o-n: It seems unnecessary that all of these functions are taking the whole user struct as parameter when only the user id is actually needed.
This means that the callers which already have user id now have to unnecessarily get the whole user to use these functions.

While it doesn't appear to be the problem in your current code, because you are using the whole `user` in other places, IMO it's still an unnecessary requirement which could lead to unnecessary queries in the future.

lib/teiserver/account/libs/user_totp_lib.ex
```diff
@@ -0,0 +1,215 @@
...
+    {_status, totp} = get_user_totp(user)
+
+    cond do
+      is_nil(totp) ->
+        false
+
+      totp.wrong_otp < 5 ->
+        false
+
+      totp.wrong_otp >= 5 ->
```
L-e-x-o-n: Extract the 5 to a constant.

lib/teiserver/account/libs/user_totp_lib.ex
ChocoboCoding: The login is treating `:valid` and `:grace` exactly the same, as it just checks for `{:ok, _}`

lib/teiserver/account/libs/user_totp_lib.ex
ChocoboCoding: I did. I'm not happy with how cluttered it is now, so if there are any suggestions on how to make it cleaner, please tell me

lib/teiserver/account/libs/user_lib.ex
ChocoboCoding: The function is only called if the password should be set, but it should be in the session controller instead, as we do not want to remove 2FA, if there is an error while updating

lib/teiserver/account/schemas/user_totp.ex
ChocoboCoding: I don't think we need a the timezone info, as the server should always have the same timezone. As it is wanted by Lexon for consistency sake, I will add it without any arguments though

test/teiserver_web/controllers/account/session_controller_test.exs
ChocoboCoding: Changed, it's asserting that now

test/teiserver_web/controllers/account/session_controller_test.exs
ChocoboCoding: Changed

lib/teiserver/account/libs/user_totp_lib.ex
```diff
@@ -0,0 +1,224 @@
...
+  end
+
+  @spec disable_totp(integer) :: {:ok, TOTP.t() | nil} | {:error, Ecto.Changeset.t()}
+  def disable_totp(user_id) do
+    case get_user_totp(user_id) do
+      {:active, totp} ->
+        Repo.delete(totp)
+
+      {:inactive, _} ->
+        {:ok, nil}
```
L-e-x-o-n: Why return `nil`? Seems unnecessary.

lib/teiserver/account/libs/user_totp_lib.ex
```diff
@@ -0,0 +1,224 @@
...
+  end
+
+  @spec set_last_used(User.t(), integer) ::
+          {:ok, TOTP.t()} | {:error, Ecto.Changeset.t()} | {:inactive, User.t()}
+  def set_last_used(%User{id: user_id} = user, last_used) do
+    if get_user_totp_status(user_id) == :active do
+      {:ok, utc_datetime} = DateTime.from_unix(last_used, :second)
+      set_totp(user_id, %{user_id: user_id, last_used: utc_datetime})
+    else
+      {:inactive, user}
```
L-e-x-o-n: Why is this returning the user? Seems to be used in tests only, should just return `:inactive` here and the new `last_used` can be grabbed in tests if needed. This function could then accept just `user_id`.

lib/teiserver/account/libs/user_totp_lib.ex
```diff
@@ -0,0 +1,237 @@
+defmodule Teiserver.Account.TOTPLib do
+  use TeiserverWeb, :library
+  alias Teiserver.Account.{User, TOTP}
+
+  @grace 5
```
geekingfrog: this attribute is used exactly once, and also the name isn't great. Just inline the value where it's used, it'll make the code less confusing.

lib/teiserver/account/libs/user_lib.ex
```diff
@@ -414,6 +414,13 @@ defmodule Teiserver.Account.UserLib do
         {:error,
          "Alt account detected. Please log in using your original account instead. If you're not sure what that account is or have trouble accessing it, please contact the moderation team at https://discord.gg/beyond-all-reason -> #open-ticket"}
 
+      TOTPLib.get_account_locked(user.id) ->
+        {:error,
+         "The 2FA one time password has been entered wrong too many times. Please reset your password to remove 2FA from your account."}
+
+      Account.get_user_totp_status(user.id) == :active ->
+        {:requires_2fa, user}
```
geekingfrog, line 422: Looking at this usecase, there is a small issue (not really a problem, just general code smell). The API you provided **requires** you to do 2 db call, but really, only 1 is necessary.
Instead, you should use `get_user_totp` to get the TOTP object, and then use this object as the argument for `get_account_locked` and `totp_status`, making these function pure (no side effect).

In general, looking at the `user_totp_lib`, almost all functions could be changed in a similar way. And to make the call sites simpler, you can also change `get_user_totp :: TOTP.t() | nil`, and make the other function handle nil (it's called nil punning in some languages like clojure).
So then you can write things like

```elixir
locked? = get_user_totp(user_id) |> totp_locked?()
```

lib/teiserver/account/libs/user_totp_lib.ex
```diff
@@ -0,0 +1,237 @@
...
+    case get_user_totp(user_id) do
+      {:inactive, nil} ->
+        {:new, NimbleTOTP.secret()}
+
+      {:active, totp} ->
+        {:existing, totp.secret}
+    end
+  end
+
+  @spec get_user_secret(integer) :: binary | nil
```
geekingfrog: invalid spec, you can also return `:inactive`

lib/teiserver/account/libs/user_totp_lib.ex
```diff
@@ -0,0 +1,237 @@
...
+
+  @spec set_totp(nil, map()) ::
+          {:ok, TOTP.t()} | {:error, Ecto.Changeset.t()}
+  defp set_totp(nil, _attrs) do
+    changeset = TOTP.changeset(%TOTP{}, %{})
+    {:error, %{changeset | errors: [user: {"user must be persisted", []}]}}
+  end
+
+  @spec set_totp(TOTP.t(), map()) ::
+          {:ok, TOTP.t()} | {:error, Ecto.Changeset.t()}
```
geekingfrog: The way to type multiple function clause is to only have a single `@spec` before the first clause. Here that would be `@spec set_totp(TOTP.t() | nil, map()) :: ...`

lib/teiserver/account/libs/user_totp_lib.ex
geekingfrog, line 1: For this entire file, you should not use `integer()` for the user_id typing. There is `Teiserver.Data.Types.userid()` for that. It's generally used with an alias `alias Teiserver.Data.Types, as: T`

the types should also be function. So instead of `binary`, it should be `binary()`. See the [reference types](https://hexdocs.pm/elixir/typespecs.html).

lib/teiserver/account/libs/user_totp_lib.ex
```diff
@@ -0,0 +1,237 @@
...
+    from(t in TOTP, where: t.user_id == ^totp.user_id)
+    |> Repo.update_all(inc: [wrong_otp: 1])
+  end
+
+  def reset_wrong_otp_counter(totp) do
+    from(t in TOTP, where: t.user_id == ^totp.user_id)
+    |> Repo.update_all(set: [wrong_otp: 0])
+  end
+
+  def validate_totp(user_or_secret, otp, time \\ System.os_time(:second))
```
geekingfrog: This isn't a great datatype to work with. You should instead use a DateTime and not a raw timestamp.

First, it help figuring out what's what and make swapping parameters more obvious. `validate(1234, 789321)` isn't great, which is which? `validate(1234, %DateTime{})` is much clearer, and if you swap them, it'll be real obvious.

Also, you are using this timestamp for checking duration and whatnot. Time is [an incredibly complex and nuance topic](https://gist.github.com/timvisee/fcda9bbdff88d45cc9061606b4b923ca) and you should use standard libraries as much as possible.
In your specific use case it's not too bad or hard, but it's generally a better habit to carry DateTime object around instead of raw timestamp.

And lastly, with datetime objects you eliminate questions like "what's the unit of this timestamp? second? ms?"

lib/teiserver/account/libs/user_totp_lib.ex
```diff
@@ -0,0 +1,237 @@
...
+  end
+
+  defp increase_wrong_otp_counter(%TOTP{} = totp) do
+    from(t in TOTP, where: t.user_id == ^totp.user_id)
+    |> Repo.update_all(inc: [wrong_otp: 1])
+  end
+
+  def reset_wrong_otp_counter(totp) do
+    from(t in TOTP, where: t.user_id == ^totp.user_id)
+    |> Repo.update_all(set: [wrong_otp: 0])
```
geekingfrog, line 143: same as above

lib/teiserver/account/libs/user_totp_lib.ex
```diff
@@ -0,0 +1,237 @@
...
+        end
+
+      {:inactive, _} ->
+        :ok
+    end
+  end
+
+  defp increase_wrong_otp_counter(%TOTP{} = totp) do
+    from(t in TOTP, where: t.user_id == ^totp.user_id)
+    |> Repo.update_all(inc: [wrong_otp: 1])
```
geekingfrog, line 138: Isn't `user_id` the primary key for this object? You should not have to use `update_all`. Looking at that code the immediate question is "wait, there are multiple tokens? why are we updating them all?"

lib/teiserver/account/libs/user_totp_lib.ex
geekingfrog: Yes it does, but should it have to do that? Since it's discarding the result, shouldn't this function just return `:ok`? I consider the grace period an implementation details of the otp logic. But making it appear in the API you are breaking this interface boundary and now all callers have to be aware of this details.

I'd rather have a simple `:ok`

lib/teiserver/account/libs/user_totp_lib.ex
```diff
@@ -0,0 +1,237 @@
...
+        # Second test is needed. If :since is provided and NimbleTOTP.valid? returns false, it could either be that the OTP is wrong, or that it got used.
+        # To figure out which one it is, we need to test a second time without since, as a false this time indicates that the OTP is invalid, and a True that it got used
+        case validate_totp(secret, otp, time) do
+          {:error, :invalid} ->
+            {:error, :invalid}
+
+          _ ->
+            {:error, :used}
+        end
+    end
```
geekingfrog, line 196: This function is a duplicate of `validate_totp/3`. You should instead the first function set a default for the `since` argument and then delegate to this one. Something like

```elixir
defp validate_totp(secret, otp, time), do: validate_totp(secret, otp, time, since: nil)`
```
and adjust that function to handle a `since: nil`
This way you don't have the duplicate logic.

lib/teiserver/account/libs/user_totp_lib.ex
```diff
@@ -0,0 +1,237 @@
...
+
+      true ->
+        # Second test is needed. If :since is provided and NimbleTOTP.valid? returns false, it could either be that the OTP is wrong, or that it got used.
+        # To figure out which one it is, we need to test a second time without since, as a false this time indicates that the OTP is invalid, and a True that it got used
+        case validate_totp(secret, otp, time) do
+          {:error, :invalid} ->
+            {:error, :invalid}
+
+          _ ->
+            {:error, :used}
```
geekingfrog, line 191: I am not convinced by this whole logic. For a similar reason to the surfacing of `:grace` on the API. Do callers really need to know that a used token is invalid? The fact that token can't be reused is also a detail of how totp work, as a caller, you merely want to know if the code you've been given is valid.
So I think returning a simple `:ok | :error` should be enough and it'll simplify both this logic and the callers.

lib/teiserver_web/controllers/account/session_controller.ex
```diff
@@ -34,9 +34,62 @@ defmodule TeiserverWeb.Account.SessionController do
...
+        login_reply({:ok, user}, conn)
+
+      {:error, reason} ->
+        flash_message =
+          case reason do
+            :used ->
+              "OTP has already been used."
+
+            :invalid ->
+              "Invalid OTP."
```
geekingfrog: continuing on the previous comments, I think this is useless. A simple "Invalid code" is enough.
Also, this is going to be shown to user, the vast majority have zero clue what `OTP` means. For them, it's just a 6 digit code that refreshes often.

lib/teiserver/account/libs/user_totp_lib.ex
ChocoboCoding: I can strip it from `validate_totp(user, otp, time)`, but not from `validate_totp(secret, otp, time)`, as that information is important for `validate_totp(user, otp, time)`, to properly save the last used timestamp

lib/teiserver/account/libs/user_totp_lib.ex
ChocoboCoding: The attribute is used 3 times, and @L-e-x-o-n will bonk me, if I remove it and inline the value. I can rename it though

lib/teiserver/account/libs/user_totp_lib.ex
ChocoboCoding: the `time` parameter purely exists for consistent testing, as it allows to just input an integer, instead of having to generate a DateTime struct, so I don't see much point in changing it

lib/teiserver/account/libs/user_totp_lib.ex
ChocoboCoding: I am not aware of another option to increase the field with only a single database query. If you are, please tell me, and I'll change

lib/teiserver/account/libs/user_totp_lib.ex
ChocoboCoding: Yes, the caller really needs to know, as we could not give the user an accurate flash message of why they could not log in. If we would just tell them that it is invalid, they would wonder "But I've just put this in, and it is valid for 20 more seconds". The only way how we can do this though, is by returning if the code has is invalid, or if it has been used

lib/teiserver/account/libs/user_totp_lib.ex
ChocoboCoding: It is not a duplicate, but needed for the distinction if an OTP has been used or is just invalid. If we call `NimbleTOTP.valid?(secret, otp, time: time, since: last_used)` we have no distinction at all, if the entered OTP is invalid, or if it has been used yet, as it will return `false` in both cases, So we need to do a second test, as explained in the 2 line comment a 6 lines above the call

lib/teiserver/account/libs/user_totp_lib.ex
ChocoboCoding: I've removed the grace period from the code, as it is more bother than it is worth it in the end

lib/teiserver/account/libs/user_lib.ex
ChocoboCoding: If you really like me to change that, I will, otherwise I'd like to leave it as is, and change it later, in a seperate, small PR

lib/teiserver/account/libs/user_totp_lib.ex
geekingfrog: And I *really* don't want to start having integer to track raw timestamp everywhere. Next time someone looks at the code to see how they should treat time, I don't want them to pick this file and spread the pattern around.
`DateTime.utc_now/0` isn't even longer to type than `System.os_time(:second)` and you don't have to worry about the unit.
Especially since the lib you're using also support DateTime object directly, you won't even need to convert it to a timestamp.

lib/teiserver/account/libs/user_totp_lib.ex
geekingfrog: ah indeed you're right, from the doc:
> Updates are used to update the filtered entries. In order for updates to be applied, [Ecto.Repo.update_all/3](https://hexdocs.pm/ecto/Ecto.Repo.html#c:update_all/3) must be invoked.

lib/teiserver/account/libs/user_totp_lib.ex
geekingfrog: I see, is there a usecase for the first version of the function, which is really just a wrapper around `NimbleTOTP.valid?`
I'd rather have exactly one way of checking if a code is valid to avoid misuse of the library. I think you could remove the first bit and replace its call site by inlining `NimbleTOTP.valid?`.

ChocoboCoding: I guess not, as `since: nil` should just have the effect of leaving it out. Will test later

lib/teiserver/account/libs/user_totp_lib.ex
ChocoboCoding: I see, that's a good and valid point. I'll change it later

lib/teiserver/account/libs/user_totp_lib.ex
ChocoboCoding: Removed it

lib/teiserver/account.ex
```diff
@@ -106,8 +106,8 @@ defmodule Teiserver.Account do
   @spec get_user_secret(integer) :: binary | nil
   defdelegate get_user_secret(user_id), to: TOTPLib
 
-  @spec get_last_used_otp(integer) :: :inactive | String.t()
-  defdelegate get_last_used_otp(user_id), to: TOTPLib
+  @spec get_last_used(integer) :: :inactive | String.t()
+  defdelegate get_last_used(user_id), to: TOTPLib
```
geekingfrog: Why the rename there? In the context of accounts, `get_last_used` isn't very useful. `get_last_used_otp` is much better.
If the context of the TOTPLib, yes, the rename makes sense. You can bridge this gap with `:as`:

`defdelegate get_last_used_totp(user_id), to: TOTPLib, as: :get_last_used`

test/teiserver/account/user_totp_lib_test.ex
```diff
@@ -14,8 +14,8 @@ defmodule Teiserver.Account.TOTPLibTest do
     %{user_without_totp: user_without_totp, user_with_totp: user_with_totp, secret: secret}
   end
 
-  defp last_used(_context) do
-    %{last_used_integer: 30, last_used_utc: ~U[1970-01-01 00:00:30Z]}
+  defp datetimes(_context) do
+    %{datetime_valid: ~U[1970-01-01 00:00:30Z], datetime_invalid: ~U[1970-01-01 00:01:00Z]}
```
geekingfrog: As I'm reading the tests, I don't find this "setup" helpful.
First, it's not really doing any setup, it's just some constants.
Second, it's not immediately clear why one is valid and the other invalid. In fact, there is no such thing as in invalid datetime without context.

I think it would be clearer to drop this setup, and copy/paste the datetime (as the `~U` sigil) where required.
And when you need something "invalid", make it clear with deltas. Something like:

```elixir
now = ~U[1970-01-01 00:00:30Z]
otp = NimbleTOTP.verification_code(secret, time: now)
assert {:error, :invalid} = TOTPLib.validate_totp(user, otp, DateTime.add(now, 30, :seconds)
```

test/teiserver/account/user_totp_lib_test.ex
```diff
@@ -158,34 +158,33 @@ defmodule Teiserver.Account.TOTPLibTest do
...
     end
 
     test "handles no TOTP", %{user_without_totp: user} do
       assert {:error, :inactive} = TOTPLib.validate_totp(user, "000000")
     end
 
-    test "does not work for wrong/outdated OTP", %{user_with_totp: user, secret: secret} do
-      otp = NimbleTOTP.verification_code(secret, time: 30)
-      assert {:error, :invalid} = TOTPLib.validate_totp(user, otp, 90)
+    test "does not work for wrong/outdated OTP", %{
```
geekingfrog: drop the "wrong" there, you are testing the logic around timing there

test/teiserver/account/user_totp_lib_test.ex
```diff
@@ -158,34 +158,33 @@ defmodule Teiserver.Account.TOTPLibTest do
...
+      datetime_valid: datetime_valid,
+      datetime_invalid: datetime_invalid
     } do
-      otp = NimbleTOTP.verification_code(secret, time: 30)
-      assert {:error, :invalid} = TOTPLib.validate_totp(user, otp, 65)
+      otp = NimbleTOTP.verification_code(secret, time: datetime_valid)
+      assert :ok = TOTPLib.validate_totp(user, otp, datetime_valid)
+      assert {:error, :used} = TOTPLib.validate_totp(user, otp, datetime_valid)
+      otp = NimbleTOTP.verification_code(secret, time: datetime_invalid)
+      assert :ok = TOTPLib.validate_totp(user, otp, datetime_invalid)
```
geekingfrog: I don't understand these last 2 lines, aren't they doing the exact same thing as the first 2 in the test?

ChocoboCoding: It's asserting that a new otp will actually work again, if the previous one was used

geekingfrog: hum, could be worth leaving a comment saying this. And also, instead of using `datetime_invalid`, just put a different timestamp, because the name makes it look like you need an "invalid" timestamp?

lib/teiserver/account.ex
ChocoboCoding: Removed it, as it is unused outside of the lib and testing

test/teiserver/account/user_totp_lib_test.ex
ChocoboCoding: Added said comment

test/teiserver/account/user_totp_lib_test.ex
```diff
@@ -158,33 +152,30 @@ defmodule Teiserver.Account.TOTPLibTest do
...
-      assert :ok = TOTPLib.validate_totp(user, otp, datetime_valid)
-      assert {:error, :used} = TOTPLib.validate_totp(user, otp, datetime_valid)
-      otp = NimbleTOTP.verification_code(secret, time: datetime_invalid)
-      assert :ok = TOTPLib.validate_totp(user, otp, datetime_invalid)
+      now = ~U[1970-01-01 00:00:30Z]
+      otp = NimbleTOTP.verification_code(secret, time: now)
+      assert :ok = TOTPLib.validate_totp(user, otp, now)
+      assert {:error, :used} = TOTPLib.validate_totp(user, otp, now)
+
+      # check that next OTP (30 seconds later) will work again
```
geekingfrog, line 162: :+1:

PR #807, opened 2025-09-28
Fixed permission check for the admin button

ChocoboCoding:
The admin button in the nav components now only shows for Server, Admin and Moderator roles


ChocoboCoding:
Never gonna edit code on my phone again -.-

PR #808, opened 2025-09-28
Adds reaction to duplicate reports

ChocoboCoding:
This automatically checks if a report from the same person, against the same target, in the same match has been submitted already, and reacts with the `:thiswatched:` emoji 
image

### DIFF
--- a/lib/teiserver/bridge/discord_bridge_bot.ex
+++ b/lib/teiserver/bridge/discord_bridge_bot.ex
@@ -304,110 +304,102 @@ defmodule Teiserver.Bridge.DiscordBridgeBot do
       end
 
     if channel do
-      report = Moderation.get_report!(report.id, preload: [:reporter, :target])
+      report_group = Moderation.get_report_group!(report.id, preload: [:reporter, :target])
 
-      host = Application.get_env(:teiserver, TeiserverWeb.Endpoint)[:url][:host]
-      url = "https://#{host}/moderation/report?target_id=#{report.target_id}"
+      if length(report_group) > 1 do
+        report_group_first = hd(report_group)
 
-      match_icon =
-        cond do
-          report.match_id == nil -> ""
-          true -> ":crossed_swords:"
-        end
-
-      outstanding_count =
-        Moderation.list_outstanding_reports_against_user(report.target_id)
-        |> Enum.count()
-
-      outstanding_msg =
-        cond do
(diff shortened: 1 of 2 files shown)


lib/teiserver/bridge/discord_bridge_bot.ex
```diff
@@ -374,10 +374,41 @@ defmodule Teiserver.Bridge.DiscordBridgeBot do
 
         if length(reports) > 1,
           do: Communication.create_discord_reaction(channel, message_id, "🔼")
+
+        if check_duplicates(report),
+          do:
+            Communication.create_discord_reaction(
+              channel,
+              message_id,
+              Communication.discord_emoji(":thiswatched:", 1_047_256_452_697_423_894)
```
L-e-x-o-n: What is this number? Where is it from? Why is it necessary here and not for the arrows used before?

lib/teiserver/bridge/discord_bridge_bot.ex
```diff
@@ -374,10 +374,41 @@ defmodule Teiserver.Bridge.DiscordBridgeBot do
...
+
+    IO.puts("check_duplicates")
+
+    if length(duplicates) > 1 do
+      IO.puts(true)
+      true
+    else
+      IO.puts(false)
+      false
+    end
```
L-e-x-o-n: This should be just `length(duplicates) > 1`.

lib/teiserver/bridge/discord_bridge_bot.ex
```diff
@@ -374,10 +374,41 @@ defmodule Teiserver.Bridge.DiscordBridgeBot do
...
+      Moderation.list_reports(
+        search: [
+          match_id: report.match_id,
+          reporter_id: report.reporter.id,
+          target_id: report.target.id
+        ],
+        order_by: "Oldest first"
+      )
+
+    IO.puts("check_duplicates")
```
L-e-x-o-n: Probably left from local testing, should be removed.

ChocoboCoding: Sadge, I forgot to remove debug code. Will remove that later -.-

lib/teiserver/bridge/discord_bridge_bot.ex
ChocoboCoding: Yes

PR #809, opened 2025-09-29
User OAuth application management

### LINKED ISSUE #526: [Feature]: Revoke authorized OAuth applications
### Describe the Feature

Under [account > security](https://server4.beyondallreason.info/teiserver/account/security) there is a list of user token.
We could add a section for the authorized OAuth applications (see https://github.com/beyond-all-reason/teiserver/pull/335) and an option to revoke access.

### Expected Behavior

Revoking access to an application should invalidate/remove all access token and refresh token associated with it for that user.

### LINKED ISSUE #590: [Feature]: Add a page to list OAuth access and revoke them
### Describe the Feature

Each user should have the ability to see which OAuth application they have authorized with their scopes (or a nice description of them) and the option to revoke access to the apps.
Revoking access should delete all corresponding code, access and refresh tokens.

### Expected Behavior

_No response_

### Screenshots

_No response_

### Additional context

_No response_

wrath625:
Adds a section under account/security to allow users to revoke all tokens associated with an OAuth application.

image

While I'm looking at this security page I noticed the whole User tokens section seems not to be doing anything anymore?  Should we remove it?

Closes #590
Closes #526

### DIFF
--- a/lib/teiserver/o_auth/queries/application_query.ex
+++ b/lib/teiserver/o_auth/queries/application_query.ex
@@ -2,6 +2,7 @@ defmodule Teiserver.OAuth.ApplicationQueries do
   use TeiserverWeb, :queries
 
   alias Teiserver.OAuth.{Application, TokenQueries, CodeQueries, CredentialQueries}
+  alias Teiserver.Data.Types, as: T
 
   @doc """
   Returns the application corresponding to the given uid/client id
@@ -87,5 +88,72 @@ defmodule Teiserver.OAuth.ApplicationQueries do
   def where_uid(query, uid) do
     from [app: app] in query,
       where: app.uid == ^uid
+  end
+
+  @doc """
+  Returns all OAuth applications that the user has authorized (has tokens for, including expired ones).
+  """
+  @spec list_authorized_applications(T.userid()) :: [Application.t()]
+  def list_authorized_applications(user_id) do
+    base_query()
+    |> join(:left, [app], token in Teiserver.OAuth.Token,
+      on: token.application_id == app.id and token.owner_id == ^user_id
+    )
+    |> where([app, token], not is_nil(token.id))
+    |> distinct([app, token], true)
+    |> preload(:owner)
+    |> Repo.all()
+  end
+
+  @doc """
+  Returns the count of active tokens for each application that the user has authorized.
+  Counts only the user's tokens, not all tokens for the application.
+  """
+  @spec get_application_token_counts(T.userid()) :: %{
+          Application.id() => non_neg_integer()
+        }
+  def get_application_token_counts(user_id) do
+    from(token in Teiserver.OAuth.Token,
+      where: token.owner_id == ^user_id,
+      where: token.expires_at > ^DateTime.utc_now(),
+      group_by: token.application_id,
+      select: {token.application_id, count(token.id)}
+    )
+    |> Repo.all()
+    |> Enum.into(%{})
+  end
+
+  @doc """
+  Deletes all tokens (access and refresh) for a specific application and user.
+  Returns the count of deleted tokens.
+  """
+  @spec delete_user_application_tokens(T.userid(), Application.id()) ::
+          non_neg_integer()
+  def delete_user_application_tokens(user_id, application_id) do
+    {count, _} =
+      from(token in Teiserver.OAuth.Token,
+        where: token.owner_id == ^user_id,
+        where: token.application_id == ^application_id
+      )
+      |> Repo.delete_all()
+
+    count
+  end
+
+  @doc """
+  Deletes all authorization codes for a specific application and user.
+  Returns the count of deleted codes.
+  """
+  @spec delete_user_application_codes(T.userid(), Application.id()) ::
+          non_neg_integer()
+  def delete_user_application_codes(user_id, application_id) do
+    {count, _} =
+      from(code in Teiserver.OAuth.Code,
+        where: code.owner_id == ^user_id,
+        where: code.application_id == ^application_id
+      )
+      |> Repo.delete_all()
+
+    count
   end
 end
--- a/lib/teiserver_web/controllers/account/security_controller.ex
+++ b/lib/teiserver_web/controllers/account/security_controller.ex
@@ -1,7 +1,9 @@
 defmodule TeiserverWeb.Account.SecurityController do
   use TeiserverWeb, :controller
+  require Logger
 
   alias Teiserver.Account
+  alias Teiserver.OAuth
 
   plug(:add_breadcrumb, name: "Account", url: "/teiserver/account")
   plug(:add_breadcrumb, name: "Security", url: "/teiserver/account/security")
@@ -21,8 +23,13 @@ defmodule TeiserverWeb.Account.SecurityController do
         order_by: "Most recently used"
       )
 
+    oauth_applications = OAuth.list_authorized_applications(conn.assigns.current_user.id)
+    oauth_token_counts = OAuth.get_application_token_counts(conn.assigns.current_user.id)
+
     conn
     |> assign(:user_tokens, user_tokens)
+    |> assign(:oauth_applications, oauth_applications)
+    |> assign(:oauth_token_counts, oauth_token_counts)
     |> render("index.html")
   end
 
@@ -46,7 +53,7 @@ defmodule TeiserverWeb.Account.SecurityController do
       {:ok, _user} ->
         conn
         |> put_flash(:info, "Account password updated successfully.")
-        |> redirect(to: Routes.ts_account_security_path(conn, :index))
+        |> redirect(to: ~p"/teiserver/account/security")
 
       {:error, %Ecto.Changeset{} = changeset} ->
         render(conn, "edit_password.html", user: user, changeset: changeset)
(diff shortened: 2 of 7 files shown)


lib/teiserver_web/controllers/account/security_controller.ex
```diff
@@ -68,4 +74,19 @@ defmodule TeiserverWeb.Account.SecurityController do
...
+    case User.revoke_application_access(conn.assigns.current_user.id, application_id) do
+      :ok ->
+        conn
+        |> put_flash(:info, "OAuth application access revoked successfully.")
+        |> redirect(to: Routes.ts_account_security_path(conn, :index))
+
+      {:error, _reason} ->
+        conn
+        |> put_flash(:error, "Failed to revoke OAuth application access.")
+        |> redirect(to: Routes.ts_account_security_path(conn, :index))
```
geekingfrog: I know there are already instances of `Routes.xxx` but the preferred way of doing that now is to use the [path sigil](https://hexdocs.pm/phoenix/Phoenix.VerifiedRoutes.html).

lib/teiserver/o_auth/user.ex
```diff
@@ -0,0 +1,44 @@
...
+
+  @doc """
+  Revokes all tokens and codes for a specific application for a user.
+  This includes access tokens, refresh tokens, and authorization codes.
+  """
+  @spec revoke_application_access(T.userid(), Teiserver.OAuth.Application.id()) ::
+          :ok | {:error, term()}
+  def revoke_application_access(user_id, application_id) do
+    Repo.transaction(fn ->
+      # Delete all tokens (access and refresh) for this user and application
```
geekingfrog: Useless comments.

lib/teiserver/o_auth/user.ex
```diff
@@ -0,0 +1,44 @@
...
+      # Delete all authorization codes for this user and application
+      {code_count, _} = UserQueries.delete_user_application_codes(user_id, application_id)
+
+      total_deleted = token_count + code_count
+
+      if total_deleted > 0 do
+        :ok
+      else
+        {:error, :no_tokens_or_codes_found}
+      end
```
geekingfrog: I don't think this is what you want. If an application is authorized but there are no token, it's a valid scenario and shouldn't be treated as an error.

lib/teiserver/o_auth/queries/user_query.ex
geekingfrog, line 1: I don't like having a `user_query` under `o_auth`, seems out of place. Especially since you are really only concerned about oauth application and not users. It's fine for some queries to take a user object or a user id. I think all of these function should be under the application queries.

lib/teiserver/o_auth/queries/user_query.ex
```diff
@@ -0,0 +1,78 @@
...
+      TokenQueries.base_query()
+      |> TokenQueries.not_expired()
+      |> where([token], token.owner_id == ^user_id)
+      |> select([token], token.application_id)
+      |> distinct(true)
+      |> Repo.all()
+
+    # Get all applications using existing query, then filter
+    ApplicationQueries.list_applications()
+    |> Enum.filter(&(&1.id in app_ids))
```
geekingfrog: You don't need two queries for this. You can directly get the applications with the matching criteria using a join on the token and code tables.

lib/teiserver/o_auth/queries/application_query.ex
```diff
@@ -88,4 +89,65 @@ defmodule Teiserver.OAuth.ApplicationQueries do
...
+    |> Repo.all()
+    |> Enum.into(%{})
+  end
+
+  @doc """
+  Deletes all tokens (access and refresh) for a specific application and user.
+  Returns the count of deleted tokens.
+  """
+  @spec delete_user_application_tokens(T.userid(), Application.id()) ::
+          {non_neg_integer(), nil}
```
geekingfrog: the return type is a bit wonky no? Why returns a tuple with a nil as the second argument? It should just return the deleted count

lib/teiserver_web/controllers/account/security_controller.ex
```diff
@@ -68,4 +74,19 @@ defmodule TeiserverWeb.Account.SecurityController do
     |> put_flash(:info, "Token deleted successfully.")
     |> redirect(to: Routes.ts_account_security_path(conn, :index))
   end
+
+  @spec revoke_oauth_application(Plug.Conn.t(), map()) :: Plug.Conn.t()
+  def revoke_oauth_application(conn, %{"id" => application_id}) do
+    case User.revoke_application_access(conn.assigns.current_user.id, application_id) do
+      :ok ->
+        conn
```
geekingfrog, line 87: I would also log at `info` level which user revoked which application access there. Since it's a hard delete on the DB, it may be useful if we ever need to investigate.

test/teiserver/o_auth/application_query_test.exs
```diff
@@ -108,4 +108,82 @@ defmodule Teiserver.OAuth.ApplicationQueryTest do
...
+    setup [:setup_app]
+
+    test "list_authorized_applications returns correct apps", %{user: user, app: app} do
+      # Create tokens for the app
+      OAuthFixtures.token_attrs(user.id, app) |> OAuthFixtures.create_token()
+
+      apps = ApplicationQueries.list_authorized_applications(user.id)
+
+      assert length(apps) == 1
+      assert hd(apps).id == app.id
```
geekingfrog: You can also do these two tests using pattern matching: `[%{id: ^app_id}] = apps` or, if you prefer not to have a `app_id` var there only for pattern match, you can get the app like so: `[app] = apps`.

I find that you get nicer error messages when something goes wrong this way. I also find that more intuitive but this is more subjective.

test/teiserver/o_auth/application_query_test.exs
```diff
@@ -108,4 +108,82 @@ defmodule Teiserver.OAuth.ApplicationQueryTest do
...
+      yesterday = DateTime.utc_now() |> Timex.subtract(Timex.Duration.from_days(1))
+
+      OAuthFixtures.token_attrs(user.id, app)
+      |> Map.put(:expires_at, yesterday)
+      |> OAuthFixtures.create_token()
+
+      apps = ApplicationQueries.list_authorized_applications(user.id)
+
+      assert length(apps) == 1
+      assert hd(apps).id == app.id
```
geekingfrog: same comment.

test/teiserver_web/controllers/account/security_controller_test.exs
```diff
@@ -13,4 +15,78 @@ defmodule TeiserverWeb.Account.SecurityControllerTest do
...
   end
+
+  describe "OAuth application revocation" do
+    setup do
+      {:ok, kw} =
+        GeneralTestLib.conn_setup(Teiserver.TeiserverTestLib.player_permissions())
+        |> Teiserver.TeiserverTestLib.conn_setup()
+
+      {:ok, conn} = Keyword.fetch(kw, :conn)
+      {:ok, user} = Keyword.fetch(kw, :user)
```
geekingfrog: You can use `fetch!` to avoid the `{:ok, x}`. And you can also do `kw[:conn]`.

test/teiserver_web/controllers/account/security_controller_test.exs
```diff
@@ -13,4 +15,78 @@ defmodule TeiserverWeb.Account.SecurityControllerTest do
...
+
+      # Create an OAuth application
```
geekingfrog: useless comment

test/teiserver_web/controllers/account/security_controller_test.exs
```diff
@@ -13,4 +15,78 @@ defmodule TeiserverWeb.Account.SecurityControllerTest do
...
+      {:ok, app} =
+        OAuth.create_application(%{
+          name: "Test App",
+          uid: "test_app",
+          owner_id: user.id,
+          scopes: ["tachyon.lobby"],
+          redirect_uris: ["http://localhost/callback"]
+        })
+
+      # Create tokens and codes for the app
```
geekingfrog: same, useless comment. Don't put comment to say **what** the code is doing, this should generally be rather obvious. Instead you should write comments to say **why** the code is doing something (when it's warranted), the context and the problem may not be clear by reading the code alone.
If there's a lot of context for an entire change and not just a block of code or file, that's where detailed commit message are also useful.

test/teiserver_web/controllers/account/security_controller_test.exs
```diff
@@ -13,4 +15,78 @@ defmodule TeiserverWeb.Account.SecurityControllerTest do
...
+
+    test "successfully revokes OAuth application access", %{
+      conn: conn,
+      app: app,
+      token: token,
+      code: code
+    } do
+      # Verify tokens and codes exist
+      assert Teiserver.OAuth.TokenQueries.get_token(token.value)
+      assert Teiserver.OAuth.CodeQueries.get_code(code.value)
```
geekingfrog: This is useless. `create_token` should be tested elsewhere, likewise for the queries. You should not have to test your test setup.

test/teiserver_web/controllers/account/security_controller_test.exs
```diff
@@ -13,4 +15,78 @@ defmodule TeiserverWeb.Account.SecurityControllerTest do
...
+
+      # Revoke access
+      conn = delete(conn, ~p"/teiserver/account/security/revoke_oauth/#{app.id}")
+
+      # Should redirect back to security page
+      assert redirected_to(conn) == ~p"/teiserver/account/security"
+
+      # Should show success message
+      assert Phoenix.Flash.get(conn.assigns.flash, :info) ==
+               "OAuth application access revoked successfully."
```
geekingfrog: I'm not a fan of checking for specific human readable output. A simple copy change would mean you also need to change the test, this brings no value.
You could check that there is a flash message and call it a day.

test/teiserver_web/controllers/account/security_controller_test.exs
geekingfrog, line 1: I think you can remove all the comments in this file, I don't find them useful and they just take space, waiting to be obsolete because the code changes

lib/teiserver/o_auth/user.ex
```diff
@@ -0,0 +1,37 @@
...
+  Database queries are delegated to Teiserver.OAuth.ApplicationQueries.
+  """
+
+  alias Teiserver.Repo
+  alias Teiserver.OAuth.{ApplicationQueries}
+  alias Teiserver.Data.Types, as: T
+
+  # Delegate query functions to ApplicationQueries
+  defdelegate list_authorized_applications(user_id), to: ApplicationQueries
+  defdelegate get_application_token_counts(user_id), to: ApplicationQueries
```
geekingfrog: that's weird to delegate that. You should instead expose these functions from the general `OAuth` context and not call `OAuth.User.list_authorized_applications`, just `OAuth.list_authorized_applications`

PR #813, opened 2025-10-03
Change friend error responses to atoms & make tachyon return the correct error

wrath625:
When implementing the friends list functionality in `bar-lobby` I noticed all the returned errors were `invalid_user`.

This makes tachyon return the correct error atoms

To maintain existing functionality I created an internal atom->string mapping for the web-UI and spring/chobby.

### DIFF
--- a/lib/teiserver/account/libs/friend_request_lib.ex
+++ b/lib/teiserver/account/libs/friend_request_lib.ex
@@ -11,6 +11,29 @@ defmodule Teiserver.Account.FriendRequestLib do
   @spec icon :: String.t()
   def icon(), do: "fa-user-plus"
 
+  @spec error_atom_to_user_friendly_string(atom() | String.t()) :: String.t()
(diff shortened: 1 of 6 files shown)


lib/teiserver/account/libs/friend_request_lib.ex
```diff
@@ -11,44 +11,67 @@ defmodule Teiserver.Account.FriendRequestLib do
...
+      :incoming_capacity_reached ->
+        "This user has reached the maximum number of friends"
+
+      atom when is_atom(atom) ->
+        to_string(atom) |> String.replace("_", " ") |> String.capitalize()
+
+      string when is_binary(string) ->
+        string
+    end
+  end
```
geekingfrog, line 35: I'm late to the party, but that kind of thing shouldn't live in this module.
You are doing translation from internal values into user facing values, that kind of logic should belong to modules related to interfaces, like views, or the tachyon/spring handler.

PR #831, opened 2025-10-15
Lobby join queue

geekingfrog:
Implement `lobby/joinQueue`, where a player is automatically added to an ally team with an empty spot, and if such spot doesn't exist, they are put in a join queue.
When a spot become free, the first player in the join queue takes it.

### DIFF
--- a/lib/teiserver/tachyon_lobby/lobby.ex
+++ b/lib/teiserver/tachyon_lobby/lobby.ex
@@ -399,7 +399,14 @@ defmodule Teiserver.TachyonLobby.Lobby do
 
   def handle_call({:spectate, user_id}, _from, state)
       when is_map_key(state.spectators, user_id) do
-    {:reply, :ok, state}
+    if state.spectators[user_id].join_queue_position == nil do
+      {:reply, :ok, state}
+    else
+      state = put_in(state.spectators[user_id].join_queue_position, nil)
+      update = %{spectators: %{user_id => %{join_queue_position: nil}}}
+      broadcast_update({:update, nil, update}, state)
+      {:reply, :ok, state}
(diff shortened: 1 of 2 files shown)


lib/teiserver/tachyon_lobby/lobby.ex
```diff
@@ -414,12 +428,69 @@ defmodule Teiserver.TachyonLobby.Lobby do
...
     broadcast_update({:update, nil, change_map}, state)
     {:reply, :ok, state}
   end
 
+  def handle_call({:join_queue, user_id}, _from, state)
+      when not is_map_key(state.players, user_id) and not is_map_key(state.spectators, user_id),
+      do: {:reply, {:error, :not_in_lobby}, state}
+
+  def handle_call({:join_queue, user_id}, _from, state) when is_map_key(state.players, user_id),
+    do: {:reply, :ok, state}
```
L-e-x-o-n: Checking if I understood correctly and that this is the intended behaviour, if the user is already a player and tries to join queue nothing happens?
They might want to skip a game and be put at the end of the queue for the next one but then they can spectate and join queue instead of doing it directly.

geekingfrog: You are correct. But I think you're also right that this isn't the ideal behaviour and what you describe makes more sense. I'll amend that soon (and fix the conflicts)

geekingfrog: Alright, I've amended [the commit](688dad50f3bf93532bf41510eb436e342729664d) so that it now behaves like the existing lobbies.

L-e-x-o-n (approved):

PR #832, opened 2025-10-16
Fix profile match list formatting

wrath625:
- I incorrectly calculated Skill (should be skill-uncertainty) on the new profile/:id/matches page, 
- and also was rounding to 1 decimal on skill change not two

### DIFF
--- a/lib/teiserver_web/live/account/profile/matches.ex
+++ b/lib/teiserver_web/live/account/profile/matches.ex
@@ -118,8 +118,8 @@ defmodule TeiserverWeb.Account.ProfileLive.Matches do
 
       match
       |> Map.put(:rating_log, rating_log)
-      |> Map.put(:rating_change, rating_log && rating_log.value["rating_value_change"])
-      |> Map.put(:skill, rating_log && rating_log.value["skill"])
+      |> Map.put(:rating_change, get_in(rating_log, [:value, "rating_value_change"]))
+      |> Map.put(:rating, get_in(rating_log, [:value, "rating_value"]))
     end)
   end
 
--- a/lib/teiserver_web/live/account/profile/matches.html.heex
+++ b/lib/teiserver_web/live/account/profile/matches.html.heex
@@ -28,9 +28,9 @@
                 Loss
               <% end %>
             </:col>
-            <:col :let={match} label="Skill">
-              <%= if @can_view_ratings && match.skill do %>
-                {round(match.skill, 1)}
+            <:col :let={match} label="Rating">
+              <%= if @can_view_ratings && match.rating do %>
+                {round(match.rating, 2)}
               <% end %>
             </:col>
             <:col :let={match} label="Rating Change">
@@ -38,7 +38,7 @@
                 <% {text_class, icon} = rating_change_display_class_and_icon(match.rating_change) %>
                 <span class={text_class}>
                   <Fontawesome.icon icon={icon} style="solid" class="me-1" />
-                  {if match.rating_change > 0, do: "+", else: ""}{round(match.rating_change, 1)}
+                  {if match.rating_change > 0, do: "+", else: ""}{round(match.rating_change, 2)}
                 </span>
               <% end %>
             </:col>


lib/teiserver_web/live/account/profile/matches.ex
```diff
@@ -119,7 +119,11 @@ defmodule TeiserverWeb.Account.ProfileLive.Matches do
       match
       |> Map.put(:rating_log, rating_log)
       |> Map.put(:rating_change, rating_log && rating_log.value["rating_value_change"])
-      |> Map.put(:skill, rating_log && rating_log.value["skill"])
+      |> Map.put(
+        :skill,
+        rating_log && rating_log.value["skill"] && rating_log.value["uncertainty"] &&
+          max(0, rating_log.value["skill"] - rating_log.value["uncertainty"])
+      )
```
geekingfrog: this is getting rather unwieldy. It would be better to extract that outside the assignment. Also, you can use `get_in` to access deeply nested properties, the macro does nil punning as well.

L-e-x-o-n: I missed this the first time, sorry.

> I incorrectly calculated Skill (should be skill-uncertainty) on the new profile/:id/matches page,

You didn't incorrectly calculate skill, you just used skill instead of rating.
The `skill - uncertainty` is for calculating rating (specifically match or balance rating, leaderboard rating is `skill - 3 * uncertainity`).
You also shouldn't be doing this calculation yourself either, the `rating_log` already has a correctly calculated rating under the key `rating_value` in the `value` map.
Please rename mentions of `skill` to `rating` here and on the page.

lib/teiserver_web/live/account/profile/matches.html.heex
```diff
@@ -30,15 +30,15 @@
             </:col>
             <:col :let={match} label="Skill">
               <%= if @can_view_ratings && match.skill do %>
-                {round(match.skill, 1)}
+                {round(match.skill, 2)}
```
L-e-x-o-n: Should be renamed to `rating`.

L-e-x-o-n (approved):

L-e-x-o-n:
This worked fine in my tests on integration but fails in production.

Example error message:
```
** (UndefinedFunctionError) function Teiserver.Game.RatingLog.fetch/2 is undefined (Teiserver.Game.RatingLog does not implement the Access behaviour

You can use the "struct.field" syntax to access struct fields. You can also use Access.key!/1 to access struct fields dynamically inside get_in/put_in/update_in)
    (teiserver 0.1.0) Teiserver.Game.RatingLog.fetch(%Teiserver.Game.RatingLog{__meta__: #Ecto.Schema.Metadata<:loaded, "teiserver_game_rating_logs">, id: 12345, user_id: 12345, user: #Ecto.Association.NotLoaded<association :user is not loaded>, rating_type_id: 5, rating_type: #Ecto.Association.NotLoaded<association :rating_type is not loaded>, match_id: 7357839, match: #Ecto.Association.NotLoaded<association :match is not loaded>, party_id: nil, value: %{"num_matches" => 138, "num_wins" => 69, "rating_value" => 16.003534892848048, "rating_value_change" => -0.4659334714420389, "skill" => 19.95219659478102, "skill_change" => -0.47106068087458297, "uncertainty" => 3.948661701932975, "uncertainty_change" => -0.005127209432541413}, inserted_at: ~U[2025-10-20 21:30:25Z], match_membership: #Ecto.Association.NotLoaded<association :match_membership is not loaded>, season: 2}, :value)
    (elixir 1.18.4) lib/access.ex:322: Access.get/3
    (elixir 1.18.4) lib/kernel.ex:2832: Kernel.get_in/2
    (teiserver 0.1.0) lib/teiserver_web/live/account/profile/matches.ex:121: anonymous fn/2 in TeiserverWeb.Account.ProfileLive.Matches.load_rating_data_for_matches/2
    (elixir 1.18.4) lib/enum.ex:1714: Enum."-map/2-lists^map/1-1-"/2
    (teiserver 0.1.0) lib/teiserver_web/live/account/profile/matches.ex:141: TeiserverWeb.Account.ProfileLive.Matches.load_user_matches/1
    (teiserver 0.1.0) lib/teiserver_web/live/account/profile/matches.ex:41: TeiserverWeb.Account.ProfileLive.Matches.handle_params/3
```

wrath625:
pushing fix in #838

PR #839, opened 2025-10-23
Documentation update

L-e-x-o-n:
I made a pass through teiserver docs and reorganised and updated most of the relevant files. Cleaned up most of the unused or out of date ones as well. This should hopefully make the setup process and getting started with development a bit clearer. 

There is still room for future improvements. Something I started already but would like to eventually expand and cover the rest is a short introduction and overview of the main teiserver components (account creation and management, user ratings, balancing, moderation, Discord bot, microblogs, chat, telemetry and more with the introduction of Tachyon).

### DIFF
--- a/documents/guides/development.md
+++ b/documents/guides/development.md
@@ -1,4 +1,6 @@
 # Development
+After finishing the local setup you are ready to start developing Teiserver.
+This guide should hopefully help you with getting started by showing you some helpful commands and giving you a short introduction to different Teiserver components.
 
 ## Useful tools and commands
 ### Fake data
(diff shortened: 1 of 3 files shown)


documents/guides/development.md
```diff
@@ -0,0 +1,126 @@
+# Development
+
+## Useful tools and commands
+### Fake data
+
```
ChocoboCoding: This should probably mention that it is needed to set up the local setup first. Otherwise there will be someone wondering why `mix teiserver.fakedata` is not working

documents/guides/development.md
```diff
@@ -0,0 +1,126 @@
...
+You will probably want some testing data (users, matches etc.) to make development and testing easier.
+
+Running the following command will generate a large amount of fake data and setup a root account for you:
```
ChocoboCoding, line 10: This should probably mention that it is needed to set up the local setup first. Otherwise there will be someone wondering why `mix teiserver.fakedata` is not working

ChocoboCoding (commented):
I generally like the changes a lot, except for one minor thing

p2004a:
Awesome work Lexon! Thanks a lot for doing this :)

documents/guides/development.md
L-e-x-o-n: Good point, I was going to point to this guide from the local setup as well but forgot about it. Added mentions for both,

PR #849, opened 2025-11-04
Rate matchmaking battles

### LINKED ISSUE #672: [Feature]: Rate battle from matchmaking
### Describe the Feature

When a matchmaking battle is over, rating should be updated for all players involved, in the same manner as existing matches.

Either reuse the same code and do this async, or, preferably, do this synchronously right as the battle ended and the outcome is known.
That would cut down the number of questions in discord about "why os no update uwu?"

The game result should also be persisted to the DB like existing battles so that all the other consumers (like https://gex.honu.pw/ or https://bar-rts.com/) can continue to work with tachyon battle.

If/when we decide to change visibility of matchmaking battle, we can adjust this logic then.

L-e-x-o-n:
Resolves #672 

Records and rates Tachyon matchmaking battles.

### DIFF
--- a/lib/teiserver/battle.ex
+++ b/lib/teiserver/battle.ex
@@ -196,6 +196,25 @@ defmodule Teiserver.Battle do
   end
 
   @doc """
+  Creates a Tachyon match.
+
+  ## Examples
+
+      iex> create_tachyon_match(%{field: value})
+      {:ok, %Match{}}
+
+      iex> create_tachyon_match(%{field: bad_value})
+      {:error, %Ecto.Changeset{}}
+
+  """
+  @spec create_tachyon_match(map()) :: {:ok, Match.t()} | {:error, Ecto.Changeset.t()}
+  def create_tachyon_match(attrs \\ %{}) do
+    %Match{}
+    |> Match.create_tachyon_match(attrs)
+    |> Repo.insert()
+  end
+
+  @doc """
   Creates a match based on starting a lobby.
 
   ## Examples
@@ -231,6 +250,25 @@ defmodule Teiserver.Battle do
   def update_match(%Match{} = match, attrs) do
     match
     |> Match.changeset(attrs)
+    |> Repo.update()
+  end
+
+  @doc """
+  Updates a Tachyon match.
+
+  ## Examples
+
+      iex> update_tachyon_match(match, %{field: new_value})
+      {:ok, %Match{}}
+
+      iex> update_tachyon_match(match, %{field: bad_value})
+      {:error, %Ecto.Changeset{}}
+
+  """
+  @spec update_tachyon_match(Match.t(), map()) :: {:ok, Match.t()} | {:error, Ecto.Changeset.t()}
+  def update_tachyon_match(%Match{} = match, attrs) do
+    match
+    |> Match.update_tachyon_match(attrs)
     |> Repo.update()
   end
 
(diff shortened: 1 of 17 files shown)


lib/teiserver/battle.ex
```diff
@@ -364,6 +402,112 @@ defmodule Teiserver.Battle do
...
 
+  def create_tachyon_matchmaking_match(start_script) do
+    ally_teams = start_script.allyTeams
+
+    team_count = ally_teams |> Enum.count()
+
+    team_size =
+      ally_teams
+      |> Enum.map(fn at -> at.teams |> Enum.count() end)
+      |> Enum.max(fn -> 0 end)
```
geekingfrog: it is guaranteed that ally teams is never empty.
Also, you can use `Enum.max_by`

lib/teiserver/battle.ex
```diff
@@ -364,6 +402,112 @@ defmodule Teiserver.Battle do
...
+
+    update_tachyon_match(match, %{started: time})
+  end
+
+  # Called once per player to report who won
+  def end_tachyon_match(match_id, time, _user_id, winning_ally_teams \\ []) do
+    match = get_match!(match_id)
+
+    already_finished? = match.finished != nil
+    winner = List.first(winning_ally_teams, nil)
```
geekingfrog: `List.first([]) == nil`, the second argument is useless. Also I think `winning_ally_team` is a better name for this variable. `winner` strongly suggests a player.

lib/teiserver/tachyon_battle.ex
```diff
@@ -54,6 +40,24 @@ defmodule Teiserver.TachyonBattle do
     end
   end
 
+  @spec start_battle(Teiserver.Autohost.id(), T.match_id()) ::
+          {:ok, T.id(), pid()} | {:error, term()}
```
geekingfrog: there should only be one spec per function, adding `when` for different clauses is ultimately the same function.

lib/teiserver/tachyon_battle.ex
```diff
@@ -41,11 +24,14 @@ defmodule Teiserver.TachyonBattle do
   """
   @spec start_battle(Bot.id(), T.start_script()) ::
           {:ok, {id(), pid()}, Autohost.start_response()} | {:error, term()}
-  def start_battle(autohost_id, start_script) do
-    with {:ok, battle_id, pid} <- start_battle(autohost_id) do
+  def start_battle(autohost_id, start_script) when is_map(start_script) do
+    with {:ok, match} <- Battle.create_tachyon_matchmaking_match(start_script),
```
geekingfrog: `start_battle` is also called [from lobbies](https://github.com/beyond-all-reason/teiserver/blob/master/lib/teiserver/tachyon_lobby/lobby.ex#L641-L644).
I think it would be better not to have a specialized function `create_tachyon_matchmaking_match` but instead something like `create_match_from_start_script(start_script, matchmaking: true|false)` or something like that.

priv/repo/migrations/20251018181107_tachyon_match.exs
```diff
@@ -0,0 +1,11 @@
+defmodule Teiserver.Repo.Migrations.TachyonMatch do
+  use Ecto.Migration
+
+  def change do
+    alter table(:teiserver_battle_matches) do
+      add :matchmaking, :boolean, default: false, null: false
```
geekingfrog, line 6: Are there a lot of matches in the table (I would guess yes)? Having a non-null default will force a backfilling and can take a long time, blocking other operations on the table.

If we can deal with backfilling time then it's by far the simplest method.

If it takes too long however, here's how to add a non nullable column:

1 add a nullable column
2 backfill the column
3 check there are no nulls
4 create a `CHECK` constraint `is not null`
5 validate the constraint
6 alter the column and make it non nullable
7 drop the constraint

See [this guide](https://django-pg-migration-tools.readthedocs.io/en/latest/usage/operations.html#SaferAlterFieldSetNotNull) for steps 4 to 7. Steps 1 must be done first, then step 2 through a task/job, then step 3 (manually), and steps 4 to 7 can be done at once.

lib/teiserver/battle.ex
```diff
@@ -364,6 +402,112 @@ defmodule Teiserver.Battle do
...
+
+    # TODO Currently trusting the first received event, should be reworked to accept what the majority agrees on
+    cond do
+      not already_finished? and winner != nil ->
+        list_match_memberships(search: [match_id: match.id, team_id: winner])
+        |> Enum.each(fn membership ->
+          update_match_membership(membership, %{win: true})
+        end)
+
+        update_tachyon_match(match, %{finished: time, winning_team: winner, processed: true})
```
geekingfrog: the update of match membership and match should happen in a transaction to avoid data corruption.

lib/teiserver/battle.ex
```diff
@@ -364,6 +402,112 @@ defmodule Teiserver.Battle do
...
+            t.players
+            |> Enum.map(fn p ->
+              %{
+                user_id: String.to_integer(p.userId),
+                name: p.name,
+                team_id: index,
+                party_id: nil,
+                match_id: match.id
+              }
+            end)
```
geekingfrog: Usually, when I start having nested flat maps, I prefer to switch to the for comprehensions, like

```elixir
for {at, index} <- Enum.with_index(ally_team), team <- at.teams, p <- team do
%{}
end
```

lib/teiserver/battle/schemas/match.ex
```diff
@@ -2,22 +2,31 @@ defmodule Teiserver.Battle.Match do
   use TeiserverWeb, :schema
 
   schema "teiserver_battle_matches" do
+    # TODO Remove after Tachyon
```
geekingfrog: hum, what?

lib/teiserver/battle.ex
```diff
@@ -364,6 +402,112 @@ defmodule Teiserver.Battle do
...
+      engine_version: start_script.engineVersion,
+      game_version: start_script.gameName,
+      team_count: team_count,
+      team_size: team_size,
+      game_type: game_type,
+      passworded: false,
+      matchmaking: true,
+      bots: Map.new(),
+      tags: Map.new(),
+      started: Timex.now()
```
geekingfrog: Prefer `DateTime.utc_now()`. I am looking at our usage of Timex, and it doesn't look like it brings anything more than what the stdlib offers nowaday. I'd rather use stdlib functions over 3rd party when possible.

lib/teiserver/battle.ex
```diff
@@ -364,6 +402,112 @@ defmodule Teiserver.Battle do
...
+          update_match_membership(membership, %{win: true})
+        end)
+
+        update_tachyon_match(match, %{finished: time, winning_team: winner, processed: true})
+
+      not already_finished? ->
+        update_tachyon_match(match, %{finished: time, processed: true})
+
+      winner != match.winning_team ->
+        Logger.warning("Match #{match_id} winning team conflict!")
```
geekingfrog: I think the logging should be outside the big `cond`, because it doesn't matter when it happen it's always a problem.

lib/teiserver/tachyon_battle.ex
```diff
@@ -54,6 +40,24 @@ defmodule Teiserver.TachyonBattle do
     end
   end
 
+  @spec start_battle(Teiserver.Autohost.id(), T.match_id()) ::
+          {:ok, T.id(), pid()} | {:error, term()}
+  def start_battle(autohost_id, match_id) do
```
geekingfrog: I believe this function should be private to this module, I don't think it's used elsewhere and it's rather confusing to have to function with identical name. Perhaps also rename into `start_battle_process` or something like that.

priv/repo/migrations/20251018181107_tachyon_match.exs
L-e-x-o-n: I think it's optimised in some way when adding a new column with a default value, I tested it on integration and running
```sql
ALTER TABLE teiserver_battle_matches ADD matchmaking boolean NOT NULL DEFAULT false;
```
takes just a few seconds (3-5), the table has 3 065 974 rows.

lib/teiserver/battle/schemas/match.ex
L-e-x-o-n: That was meant for `server_uuid` and `uuid` which won't be necessary in the future, I will remove the comment.

lib/teiserver/battle.ex
L-e-x-o-n: That's nicer, changed.

priv/repo/migrations/20251018181107_tachyon_match.exs
geekingfrog: ok, if it's only a few seconds that may be fine. It's less about the time it takes and more about requiring an exclusive lock over the table. But a few seconds should be alright

lib/teiserver/battle.ex
```diff
@@ -364,6 +402,110 @@ defmodule Teiserver.Battle do
     :ok
   end
 
+  @spec create_match_from_start_script(Teiserver.TachyonBattle.start_script(), boolean()) ::
+          {:ok, Match.t()} | {:error, Ecto.Changeset.t()}
+  def create_match_from_start_script(start_script, matchmaking) do
```
geekingfrog: Looking at this function now, and the new call sites, I think adding a new boolean like that isn't the greatest. It's known as boolean blindness, it doesn't bring a lot.
I think it would be better to either

1. do that through options/keword lists: `is_matchmaking: true|false`
2. use a more descriptive name like `source: :matchmaking | :lobby` and transform that into the boolean required for sql.

I think using a keyword list is important because I'm pretty sure we are going to add new things to this function later.

L-e-x-o-n: I think it could be renamed to `is_matchmaking` or similar to make its purpose clearer. Not sure about using options/keyword lists here, I see those as a way to pass optional arguments which already have a reasonable default value or are less impactful. In this case it is quite important whether the match is from matchmaking or not because that will affect things like which rating is updated. I think having it explicitly passed through a boolean parameter like here makes this expected behaviour clearer and avoids mistakes due to forgetting to pass options or incorrectly assuming what the default option is.
That's just my reasoning but I trust your experience so if you disagree I can change it to options list.

geekingfrog: good point about the fact it's required. Let's leave it as a param then. But I think `is_matchmaking` is better. You can also add that to the spec: `, is_matchmaking :: boolean()`

PR #869, opened 2025-11-27
chore: Streamline initial setup with mise

badosu:
Streamline the documentation to more clearly denote how to start the application and ensure the local setup is working correctly.

Move certain sections from `local_setup.md` to `development.md`.

We still keep the manual method in case we want to have that available.

### DIFF
--- a/documents/guides/local_setup.md
+++ b/documents/guides/local_setup.md
@@ -1,39 +1,53 @@
 # Local setup
+
 This guide should contain all the necessary steps for setting up Teiserver locally.
 
-## Install services
+## Using mise
+
+You can use [mise](https://mise.jdx.dev/dev-tools/) to automatically manage the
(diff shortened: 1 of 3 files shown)


documents/guides/local_setup.md
```diff
@@ -1,39 +1,53 @@
 # Local setup
+
 This guide should contain all the necessary steps for setting up Teiserver locally.
 
-## Install services
+## Using mise
+
+You can use [mise](https://mise.jdx.dev/dev-tools/) to automatically manage the
```
badosu, line 7: This is the more relevant change, in theory we could just drop all the description from `Manual Setup` and just have these two commands.

L-e-x-o-n: Mise seems great and should be the recommended way to setup teiserver but I would still keep manual setup section for now.

PR #872, opened 2025-12-02
refactor: Supervise spring tcp servers

badosu:
This accomplishes the following:

- Allows our application to supervise the spring tcp servers
- Lays the groundwork for manually starting spring servers in test mode for asynchronous tests

Ranch embedded mode: https://ninenines.eu/docs/en/ranch/1.8/guide/embedded/

Additionally, we:

- Decouple configuration of spring tcp servers per listener
- Increase configurability of each listener
- Simplify server initialization process
- Deprecate ranch protocol initialization process with socket argument
- Remove some test cruft

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -35,7 +35,54 @@ config :teiserver, TeiserverWeb.Endpoint,
   debug_errors: Config.config_env() == :dev,
   code_reloader: Config.config_env() == :dev,
   check_origin: Config.config_env() == :prod,
-  version: Mix.Project.config()[:version]
+  version: Mix.Project.config()[:version],
+  https: [
+    versions: [:"tlsv1.2"],
+    port: 8888,
+    otp_app: :teiserver,
+    ciphers: [
+      ~c"ECDHE-ECDSA-AES256-GCM-SHA384",
+      ~c"ECDHE-RSA-AES256-GCM-SHA384",
+      ~c"ECDHE-ECDSA-AES256-SHA384",
+      ~c"ECDHE-RSA-AES256-SHA384",
+      ~c"ECDHE-ECDSA-DES-CBC3-SHA",
+      ~c"ECDH-ECDSA-AES256-GCM-SHA384",
+      ~c"ECDH-RSA-AES256-GCM-SHA384",
+      ~c"ECDH-ECDSA-AES256-SHA384",
+      ~c"ECDH-RSA-AES256-SHA384",
+      ~c"DHE-DSS-AES256-GCM-SHA384",
+      ~c"DHE-DSS-AES256-SHA256",
+      ~c"AES256-GCM-SHA384",
+      ~c"AES256-SHA256",
+      ~c"ECDHE-ECDSA-AES128-GCM-SHA256",
+      ~c"ECDHE-RSA-AES128-GCM-SHA256",
+      ~c"ECDHE-ECDSA-AES128-SHA256",
+      ~c"ECDHE-RSA-AES128-SHA256",
+      ~c"ECDH-ECDSA-AES128-GCM-SHA256",
+      ~c"ECDH-RSA-AES128-GCM-SHA256",
+      ~c"ECDH-ECDSA-AES128-SHA256",
+      ~c"ECDH-RSA-AES128-SHA256",
+      ~c"DHE-DSS-AES128-GCM-SHA256",
+      ~c"DHE-DSS-AES128-SHA256",
+      ~c"AES128-GCM-SHA256",
+      ~c"AES128-SHA256",
+      ~c"ECDHE-ECDSA-AES256-SHA",
+      ~c"ECDHE-RSA-AES256-SHA",
+      ~c"DHE-DSS-AES256-SHA",
+      ~c"ECDH-ECDSA-AES256-SHA",
+      ~c"ECDH-RSA-AES256-SHA",
+      ~c"AES256-SHA",
+      ~c"ECDHE-ECDSA-AES128-SHA",
+      ~c"ECDHE-RSA-AES128-SHA",
+      ~c"DHE-DSS-AES128-SHA",
+      ~c"ECDH-ECDSA-AES128-SHA",
+      ~c"ECDH-RSA-AES128-SHA",
+      ~c"AES128-SHA"
+    ],
+    secure_renegotiate: true,
+    reuse_sessions: true,
+    honor_cipher_order: true
+  ]
 
 config :esbuild,
   version: "0.14.41",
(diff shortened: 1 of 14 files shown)


config/config.exs
```diff
@@ -46,28 +46,46 @@ config :esbuild,
...
-  heartbeat_interval: 30_000,
-  # Heartbeat timeout is seconds
-  heartbeat_timeout: 120,
   test_mode: false,
   server_admin_name: "Server Admin",
   game_name: "Full game name",
   game_name_short: "Game",
   main_website: "https://site.com/",
   discord: nil,
-  default_spring_protocol: Teiserver.Protocols.Spring,
```
badosu, line 70: unused

config/runtime.exs
```diff
@@ -94,90 +94,122 @@ config :teiserver, TeiserverWeb.Endpoint,
...
 
+spring_listeners =
+  Application.get_env(:teiserver, Teiserver.SpringTcpServer)
+  |> Keyword.fetch!(:listeners)
+
+spring_listeners =
+  if Keyword.get(spring_listeners, :tcp) do
+    # If TCP listener is enabled put the port option
+    get_and_update_in(spring_listeners, [:tcp, :socket_opts, :port], fn default ->
+      {default, Teiserver.ConfigHelpers.get_env("TEI_SPRING_TCP_PORT", default, :int)}
```
badosu, line 111: The runtime configuration logic ended up becoming very convoluted, but it was already convoluted to start with, we had to make the adjustments.

lib/teiserver/tcp/spring/spring_tcp_server.ex
```diff
@@ -12,182 +12,33 @@ defmodule Teiserver.SpringTcpServer do
...
-  def get_standard_tcp_opts() do
-    [
-      max_connections: :infinity,
-      nodelay: false,
-      delay_send: true
-    ]
-  end
-
-  # Called at startup
-  def start_link(opts) do
```
badosu, line 41: Unneeded, we start the server as a normal child now, via `:ranch.child_spec`.

lib/teiserver/tcp/spring/spring_tcp_server.ex
```diff
@@ -1295,9 +1144,6 @@ defmodule Teiserver.SpringTcpServer do
       :ring ->
         SpringOut.reply(:ring, {data, state.userid}, nil, state)
 
-      :welcome ->
```
badosu, line 1298: Unneeded signal, we only use this logic at initialization.

lib/teiserver/tcp/spring/spring_tcp_server.ex
```diff
@@ -12,182 +12,33 @@ defmodule Teiserver.SpringTcpServer do
...
-
-    [
-      certfile: certfile,
-      cacertfile: cacertfile,
-      keyfile: keyfile
-    ]
-  end
-
-  def get_standard_tcp_opts() do
-    [
```
badosu, line 33: Moved to config

lib/teiserver/tcp/spring/spring_tcp_server.ex
```diff
@@ -12,182 +12,33 @@ defmodule Teiserver.SpringTcpServer do
...
-      delay_send: true
-    ]
-  end
-
-  # Called at startup
-  def start_link(opts) do
-    use_tls? = Application.get_env(:teiserver, Teiserver)[:use_tls?]
-
-    # start_listener(Ref, Transport, TransOpts0, Protocol, ProtoOpts)
-    case {use_tls?, Keyword.get(opts, :ssl, false)} do
```
badosu, line 45: The child spec is simply not provided instead of checking on initialization.

lib/teiserver/tcp/spring/spring_tcp_server.ex
```diff
@@ -12,182 +12,33 @@ defmodule Teiserver.SpringTcpServer do
...
 
-    redirect_url = Config.get_site_config_cache("system.Redirect url")
-
-    if redirect_url != nil do
-      SpringOut.reply(:redirect, redirect_url, nil, state)
-      send(self(), :terminate)
-      state
-    else
-      send(self(), {:action, {:welcome, nil}})
+    case Teiserver.Config.get_site_config_cache("system.Redirect url") do
```
badosu, line 33: Very unlikely we ever will use this, still I aimed to make it the same logic (and fixed the wrong protocol impl)

lib/teiserver/protocols/spring/spring.ex
```diff
@@ -1,14 +1,6 @@
...
   @moduledoc false
   alias Teiserver.BitParse
-  alias Teiserver.Protocols.SpringIn
-  alias Teiserver.Protocols.SpringOut
-
-  @spec protocol_in :: Teiserver.Protocols.SpringIn
-  def protocol_in(), do: SpringIn
-
-  @spec protocol_out :: Teiserver.Protocols.SpringOut
-  def protocol_out(), do: SpringOut
```
badosu, line 11: This was some testing cruft that was never really needed

config/test.exs
```diff
@@ -46,6 +48,8 @@ config :teiserver, DiscordBridgeBot,
 config :teiserver, TeiserverWeb.Endpoint,
   url: [host: "localhost"],
   http: [port: 4002],
+  # We don't spawn the https endpoint in test
```
badosu, line 51: This was failing tests when the TLS environment variable was set (since tests were expecting only http stuff)

config/runtime.exs
L-e-x-o-n: For the purpose of simplification I think it would be fine to even just remove this env variable and use the `config.exs` defined port, I don't see us changing it at any point as a lot of other components depend on it.

lib/teiserver/tcp/spring/spring_tcp_server.ex
```diff
@@ -1413,4 +1255,166 @@ defmodule Teiserver.SpringTcpServer do
...
+        Config.get_site_config_cache("teiserver.Spring flood rate window size"),
+      last_action_timestamp: nil,
+      server_messages: 0,
+      server_batches: 0,
+      client_messages: 0
+    }
+    |> Map.merge(%{
+      heartbeat_timeout: Keyword.get(spring_config, :heartbeat_timeout),
+      heartbeat_interval: Keyword.get(spring_config, :heartbeat_interval)
+    })
```
L-e-x-o-n, line 1323: Why use `Map.merge` here instead of just making them a part of the initial map?

badosu: Mostly a choice to separate the scope of things that come from the config. We can include on the initial map indeed, just let me know.

config/runtime.exs
badosu: If you could provide a list of all environment variables listed in this file that don't exist in any provisioning scheme (and **absolutely** are not present in the production deployment) I'll be happy to clean this file up.

Of course if you tell me just this environment variable is enough I don't mind either.

badosu: Assuming the provisioning uses the environment variables defined [here](https://github.com/beyond-all-reason/ansible-teiserver/blob/7ca4917c65cc8ebe9a52443d6b77c2477a7a4ae0/roles/teiserver/templates/teiserver.env.j2#L3) and nowhere else I identified the body of work necessary.

If you don't mind I'd rather send it as a separate PR.

L-e-x-o-n: > Assuming the provisioning uses the environment variables defined [here](https://github.com/beyond-all-reason/ansible-teiserver/blob/7ca4917c65cc8ebe9a52443d6b77c2477a7a4ae0/roles/teiserver/templates/teiserver.env.j2#L3) and nowhere else I identified the body of work necessary.
> 
> If you don't mind I'd rather send it as a separate PR.

That's right.
Separate PR is good.

lib/teiserver/tcp/spring/spring_tcp_server.ex
L-e-x-o-n: I would include it in the initial map but it's not an issue if it stays this way.

lib/teiserver/application.ex
```diff
@@ -218,10 +212,6 @@ defmodule Teiserver.Application do
     Teiserver.Startup.startup()
   end
 
-  defp oban_config do
```
NortySpock, line 221: ~Never used, I assume?~

Ah, removed / refactored out.

lib/teiserver/tcp/spring/spring_tcp_server.ex
```diff
@@ -1413,4 +1255,166 @@ defmodule Teiserver.SpringTcpServer do
...
+      # See: https://www.erlang.org/doc/apps/kernel/inet.html#setopts/2
+      # > Received Packet is delivered as defined by Mode.
+      #
+      # See: https://www.erlang.org/doc/apps/kernel/gen_tcp.html#listen/2
+      # > Received `Packet`s are delivered as lists of bytes,
+      mode: :list,
+
+      # FIXME: This looks sketchy!
+      # Why don't we verify the certificate when we do a TLS upgrade?
+      verify: :verify_none
```
geekingfrog, line 1412: No, this is fine. This option is to enable mutual authentication, when the clients are authenticated using the same type of certificates as servers. They have to be provisioned out of band.
We are not using mutual auth so this option makes sense.

PR #873, opened 2025-12-03
Disable rating restrictions and rename unranked lobbies

L-e-x-o-n:
The goal is to improve unranked experience for people who want more casual games without the pressure of ratings. These server changes will be paired with client changes that will hide the ratings in lobbies and in game in unranked lobbies. It should hopefully help with reducing conflicts and toxicity caused by people having different expectations from a game. It's not perfect but it's good enough until we have proper competitive casual split with the new client and matchmaking.

### DIFF
--- a/lib/teiserver/coordinator/consul_server.ex
+++ b/lib/teiserver/coordinator/consul_server.ex
@@ -59,7 +59,7 @@ defmodule Teiserver.Coordinator.ConsulServer do
 
   def handle_call(:get_consul_state, _from, state) do
     result =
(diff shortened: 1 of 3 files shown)


L-e-x-o-n:
> (curious) What problematic behavior does this solve?
> 
> Were there cases where all active lobbies, including unranked, had min or max OS limits, and thus certain players were unable to join any active lobby?

This is the result of internal discussions. The goal is to improve unranked experience for people who want more casual games without the pressure of ratings. These server changes will be paired with client changes that will hide the ratings in lobbies and in game in unranked lobbies. It should hopefully help with reducing conflicts and toxicity caused by people having different expectations from a game. It's not perfect but it's good enough until we have proper competitive casual split with the new client and matchmaking.

L-e-x-o-n:
> Makes sense. Guess you could either full-randomize the rank served when in unranked mode, or somehow retain the rank-drives-player-ordering (so the highest rank is at the top, but nobody knows what the delta is between players)
> 
> I guess full-random might be best for "setting the tone" of an unranked match.

Eventually we will have a hidden unranked rating for the purposes of balancing, but that requires work outside the scope of this PR. In the meantime unranked will continue using "ranked' ratings for balancing but without updating them after the match.

PR #877, opened 2025-12-05
Spring telemetry - command metrics

L-e-x-o-n:


lib/teiserver/telemetry/libs/telemetry_lib.ex
```diff
@@ -71,7 +71,37 @@ defmodule Teiserver.Telemetry.TelemetryLib do
...
+      ),
+      counter(
+        [:teiserver, :spring, :in, :count],
+        event_name: [:spring, :in],
+        measurement: :count,
+        tags: [:command]
+      ),
+      counter(
+        [:teiserver, :spring, :out, :count],
+        event_name: [:spring, :event],
```
p2004a, line 101: @L-e-x-o-n should this be `event_name: [:spring, :out],` ?

Right now `teiserver_spring_out_duration_bucket` is exported, but `teiserver_spring_out_count` is not. The "Spring Outgoing Commands" graph in dashboards shows `in` messages.

L-e-x-o-n: Yes it should. Not sure how I didn't notice that.

L-e-x-o-n: Should be fixed in #894

PR #881, opened 2025-12-06
refactor: only run spring server on tests that require them

badosu:
Also fix duplicate party leave messages when a user disconnects

### DIFF
--- a/lib/teiserver/coordinator.ex
+++ b/lib/teiserver/coordinator.ex
@@ -4,7 +4,8 @@ defmodule Teiserver.Coordinator do
   alias Teiserver.Data.Types, as: T
   require Logger
 
-  @spec start_coordinator() :: {:error, String.t()} | DynamicSupervisor.on_start_child()
+  @spec start_coordinator() ::
+          {:error, {:already_started, pid()}} | DynamicSupervisor.on_start_child()
   def start_coordinator() do
     case get_coordinator_pid() do
       nil ->
@@ -14,8 +15,8 @@ defmodule Teiserver.Coordinator do
           name: Teiserver.Coordinator.CoordinatorServer, data: %{}
         })
 
-      _pid ->
-        {:error, "Already started"}
+      pid ->
+        {:error, {:already_started, pid}}
     end
   end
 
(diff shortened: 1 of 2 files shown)


lib/teiserver/coordinator.ex
```diff
@@ -4,29 +4,18 @@ defmodule Teiserver.Coordinator do
...
+    case get_coordinator_pid() do
+      nil ->
+        # Start the supervisor server
+        DynamicSupervisor.start_child(Teiserver.Coordinator.DynamicSupervisor, {
+          Teiserver.Coordinator.CoordinatorServer,
+          name: Teiserver.Coordinator.CoordinatorServer, data: %{}
+        })
+
+      _pid ->
+        {:error, "Already started"}
```
geekingfrog: might as well use the same format as `start_child`: `{:error, {:already_started, pid}}`.

test/support/teiserver_test_lib.ex
```diff
@@ -557,6 +506,110 @@ defmodule Teiserver.TeiserverTestLib do
...
+    end)
+
+    {:ok, %{spring_server: Map.new(listeners)}}
+  end
+
+  def start_coordinator!() do
+    assert {:ok, pid} = Teiserver.Coordinator.start_coordinator()
+
+    # Wait until the coordinator account exists
+    # Teiserver.Support.Polling.poll_until_some(&Teiserver.Coordinator.get_coordinator_userid/0)
```
geekingfrog: that can be remove it seems

badosu:
> THat looks alright. Most test you modified are disabled in CI (tagged `:needs_attention`). Do this change improve the situation?

Slowly, this is base work for getting everything around up to shape. Specifically, by making the spring listener not be global we can tweak ecto sandbox allowance per process, run things asynchronously etc.

test/support/teiserver_test_lib.ex
badosu: not super important to me, but it is good to "document" what guarantees we might want on our tests. but sure

badosu:
> Seems good, can merge if you resolve geekingfrog's comments.

Done

geekingfrog (approved):

PR #884, opened 2025-12-11
chore dialyzer: switch specs from Atom.t() to atom() ,  Boolean.t() to bool(), List.t() to list() to fix dialyzer unknown_type warnings

NortySpock:
### description
Went looking for trouble with the dialyzer and found some more quick fixes....

### LLM Disclosure

I asked Microsoft Copilot to diagnose the dialyzer issue for one file, and that revealed that some datatypes were being used wrong in the spec, apparently...

I did **not** use it to modify any code. 

### more info
I did a normal old-school find-replace based on the llm recommendation to use the builtin types , and that allowed me to clean up 82 lines in in .dialyzer_ignore.exs

I have manually reviewed every line of the diff and I'm pretty sure this only affects the `@spec` lines but I am only human.

### thoughts for later improvement opportunities
~I am getting one case where `mix dialyzer  . --list-unused-filters` reports~

~But removing that unleashes some cascading invalid_contract issues, so perhaps that is the next hotspot to look at.~ 

I do also suspect (from my dialyzer adventures last week) that the following ignores are hiding some cases where functions in battle.ex can occasionally fail to return, so that might also be another typecheck hotspot to wade into...

```
  {"lib/teiserver/battle.ex", :unknown_type},
  {"lib/teiserver/battle.ex", :call},
  {"lib/teiserver/battle.ex", :no_return},
```

Also (for similar "explosion of dialyzer errors" reasons) I suspect there is also something gnarly being suppressed in client_lib. 
```
{"lib/teiserver/account/libs/client_lib.ex", :unknown_type}
```

Frankly, anywhere that `:no_return` shows up in the `.dialyzer_ignore.exs` I wonder if that's actually a legitimate detection where a function might actually fail to return due to an exception being thrown... I presume that would either hang or eventually return nil or timeout or something...


NortySpock:
Ok, that looks good, I think that's as far as I am going to get in this round of fighting with the dialyzer.

PR #886, opened 2025-12-21
upgrade to elixir 1.19.4

NortySpock:
Upgrades to elixir 1.19.4 for faster compilation, more warnings, and better type checking. 

Disables "error on warning" for now because not all of the new warnings for 1.19 have been addressed. I plan to raise PRs over time to address the ones we can address.

Fixes some of the warnings by running `mix format --migrate`

According to https://hexdocs.pm/elixir/compatibility-and-deprecations.html#deprecations I don't think we need to solve all of the warnings until we attempt to upgrade to Elixir 2.x, which does not exist, so we have some time. :) 

This compiles and runs and passes `mix test --exclude needs_attention` on my machine. 

This would need to be coordinated with https://github.com/beyond-all-reason/ansible-teiserver/pull/18/ 

Not sure how to address the dialyzer failures seen here, I had to delete the dialyzer cache folders on my machine, so possible we need to disable the cache here


Dockerfile
```diff
@@ -20,8 +20,8 @@
 #       teiserver
 #
 
-ARG ELIXIR_VERSION=1.18.4
-ARG OTP_VERSION=26.2.5.14
+ARG ELIXIR_VERSION=1.19.4
+ARG OTP_VERSION=26.2.5.2
```
geekingfrog: why the change on the otp version?

NortySpock: Downgraded for consistency across the rest of the teiserver project , but I see that is out of sync with ansible-teiserver, so I have (in a later commit) adjusted the OTP version to be in sync with ansible-teiserver. 

I have fired up `pg:start` and it works with the tests. I tested the webui and it works fine, I have now smoke-tested Chobby.  

I have **not** had success in setting up SPADS, so I don't have a good way to test that. 

Yep, next time I'll leave it as separate commits. 

I'll be offline for the next two weeks and won't be able to contribute again until January, so this might sit idle for a while. I'll rebase when it's ready to go again (unless you want to fork my branch, smoke-test chobby and SPADS, and take it over the finish line?)


**Edit:** Follow-up, I smoke-tested Chobby and that appears to allow users to login and chat, and the chats appear to be visible both in Chobby and in the webui. 

image

Friend requests between two fake users via Chobby appeared to work.

I have not yet gotten SPADS working locally (still trying to figure out either Incus for containers or Perl locally), so I don't have a good way to test lobbies and such things.

L-e-x-o-n:
Tested, seems to work.

PR #898, opened 2025-12-28
Add two Mise tasks to run tests and server

dmarcoux:
They both also act as developer documentation.

Beside this, I noticed that links to the website in the development guide weren't clickable, so I fixed them.

### DIFF
--- a/documents/guides/development.md
+++ b/documents/guides/development.md
@@ -41,8 +41,8 @@ When successful, the following services should be available:
 
 The website should be available via the URLs below:
 
-- HTTP: [http://localhost:4000/]
-- HTTPS: [https://localhost:8888/]
+- HTTP: http://localhost:4000/
+- HTTPS: https://localhost:8888/
 
 ### Spring Lobby Server
 


scripts/show_test_coverage_for_files_changed_in_last_month.sh
```diff
@@ -18,7 +18,7 @@ git diff HEAD '@{last month}' --name-only > ./cover/files_changed_in_last_month.
 # so the header is also emitted and highlighted
 echo "LINES RELEVANT   MISSED" >> ./cover/files_changed_in_last_month.txt
 
-mix test --exclude needs_attention --cover > ./cover/test_coverage_by_file.txt
+mise run tests --cover > ./cover/test_coverage_by_file.txt
```
dmarcoux: Anything after `mise run tests` is passed as arguments, so this works.

documents/guides/development.md
```diff
@@ -41,8 +41,8 @@ When successful, the following services should be available:
 
 The website should be available via the URLs below:
 
-- HTTP: [http://localhost:4000/]
-- HTTPS: [https://localhost:8888/]
+- HTTP: http://localhost:4000/
+- HTTPS: https://localhost:8888/
```
dmarcoux, line 45: This is how it was before:

Image

And [now with the changes](https://github.com/dmarcoux/teiserver/blob/9623d705042ddd40b5e55f182b509114ecb2c56a/documents/guides/development.md#website).

L-e-x-o-n:
I don't like the replacement of mix commands with mise ones in docs and script, it's effectively just aliasing mix commands behind mise and the benefits of doing so seem minor. IMO it's more beneficial that the mix commands are used directly here because the documentation for those and the arguments can be looked up more easily and mix commands are what new contributors to teiserver will encounter in other projects/documentation/resources. It also assumes mise is installed which, while recommended, might not be the case.

dmarcoux:
True, you are making good points. I pushed changes in order to stick to `mix` in documentation and scripts. Let me know what you think.

L-e-x-o-n:
Thanks

PR #903, opened 2025-12-29
Fix Credo issues

dmarcoux:
When starting to address Credo issues, I thought it would fit in a single PR. How naive of me :smile:. Anyway, I had some help from Claude Code in order to disable Credo issues. I also fixed a few Credo issues.

This PR is the first step towards running Credo in the continuous integration to enforce consistent Elixir code. There are a lot of Credo issues, so for now, we ignore _Readability_ and _Refactor_ checks. _Design_ checks have been addressed or ignored in this PR. In the future, the goal is to get rid of all those `# credo:disable*` comments.

As always, please refer to the commit messages/descriptions for further details.

### DIFF
--- a/.credo.exs
+++ b/.credo.exs
@@ -6,7 +6,10 @@
         {Credo.Check.Warning.MissedMetadataKeyInLoggerConfig,
          [
            metadata_keys: [:request_id, :user_id, :pid, :actor_type, :actor_id]
-         ]}
+         ]},
+        # TODO: Enable this check and fix the issues
+        {Credo.Check.Design.AliasUsage, false}
+        # {Credo.Check.Design.AliasUsage, [if_nested_deeper_than: 2, if_called_more_often_than: 1]}
       ]
     }
   ]
--- a/lib/teiserver/account.ex
+++ b/lib/teiserver/account.ex
@@ -2217,6 +2217,7 @@ defmodule Teiserver.Account do
   @spec update_client(T.userid(), map()) :: nil | :ok
   defdelegate update_client(userid, partial_client), to: ClientLib
 
+  # credo:disable-for-next-line Credo.Check.Design.TagTODO
   # TODO: Remove these in favour of update_client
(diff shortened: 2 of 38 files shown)


dmarcoux:
So after tackling a bunch of Credo issues with the help of Claude Code, I realize now that I need to split this work into multiple PRs for you to easily review this.

Now, all Design checks from Credo are ignored. I see that a lot of them are for [Credo.Check.Design.AliasUsage](https://hexdocs.pm/credo/Credo.Check.Design.AliasUsage.html). What's your opinion on this? Should we use `alias SomeModule` or not? Credo is configured to catch issues when a module function is called more than once in a file, the module should be aliased.

I think the best approach to make this PR mergeable is to have Credo run in the continuous integration, but only for Design checks for now. Later PRs will expand this to run for other Credo checks. Eventually, we run Credo for all checks.

geekingfrog:
> The [Credo.Check.Design.AliasUsage](https://hexdocs.pm/credo/Credo.Check.Design.AliasUsage.html) check has been ignored multiple times. Credo is configured to catch issues when a module function is called more than once in a file, then the module should be aliased.
> 
> What's your opinion on this? Should we use `alias SomeModule` or not?

I'm probably guilty of that, but I think it makes sense. The argument about seeing all the module dependencies at a glance is a good one.
Though perhaps do that in a later PR?

L-e-x-o-n:
> Looking at all the inline ignore for `AliasUsage`, I think I'd rather have the entire check disabled for now, instead of adding all these lines. Then we can enable it and fix everything in a later PR.

Agreed

dmarcoux:
Thank you both for your feedback :slightly_smiling_face:. I removed all the `# credo:disable-*` comments for `Credo.Check.Design.AliasUsage`. This check is now disabled in the Credo configuration. I also dropped the commit which was regenerating the Credo configuration as it's much clearer to only keep around what differs from Credo defaults.

This PR should now be much easier to review.

PR #905, opened 2025-12-30
Prefer Mise over asdf to set up development environment

dmarcoux:
See this [comparison between Mise and asdf](https://mise.jdx.dev/dev-tools/comparison-to-asdf.html). Yes, it's on Mise's website, but it's honest.

It's clear to contributors what they should use now that we only recommend Mise.

In summary, Mise has tasks and environments, better supply chain security, Windows support, and improved UX.

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,2 +1,0 @@
-erlang 26.2.5.1
-elixir 1.19.4-otp-26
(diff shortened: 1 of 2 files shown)


dmarcoux:
@L-e-x-o-n You brought back `.tool-versions` in #604, so I would like to hear what you think about this.

PR #908, opened 2025-12-31
Add warning about mise not being self contained

p2004a:


geekingfrog:
TIL distrobox, that looks neat

PR #909, opened 2026-01-01
Address Credo issues for Refactor checks

dmarcoux:
Some issues have been fixed, while others have been temporarily ignored. Those will be addressed in other PRs. See the commit messages/descriptions for details.

### DIFF
--- a/.dialyzer_ignore.exs
+++ b/.dialyzer_ignore.exs
@@ -3,6 +3,8 @@
   {"lib/teiserver.ex", :invalid_contract},
   {"lib/teiserver/account.ex", :unknown_type},
   {"lib/teiserver/account.ex", :invalid_contract},
+  # This started after inverting a condition in an `if/else` statement. It's hard to fix with all the other ignored warnings here.
+  {"lib/teiserver/account.ex", :guard_fail},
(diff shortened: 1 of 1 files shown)


lib/teiserver/account.ex
```diff
@@ -1928,9 +1928,7 @@ defmodule Teiserver.Account do
   def create_friend_request(from_user_id, to_user_id) do
     {allowed, reason} = can_send_friend_request_with_reason?(from_user_id, to_user_id)
 
-    if not allowed do
-      {:error, reason}
-    else
+    if allowed do
```
dmarcoux, line 1931: I don't really understand why Dialyzer emits a warning here now... this `if/else` and its content were only inverted.

dmarcoux: I decided on ignoring this Dialyzer warning for now. We really need to clean up that list of [Dialyzer ignored warnings](https://github.com/beyond-all-reason/teiserver/blob/master/.dialyzer_ignore.exs).

L-e-x-o-n: I don't see the reason either, strange.

PR #910, opened 2026-01-01
Restore party state after reboot

geekingfrog:
This implement a logic described in [the doc](https://github.com/geekingfrog/teiserver/blob/restore-state-wip/documents/dev_designs/tachyon_system.md#reconnection-and-state-restoration-after-node-restart) to recover some state when we restart a node. The end goal is to be able to redeploy a new version of the server with minimal disruptions: lobbies and matchmaking are kept and players should only notice a few seconds of unresponsiveness at most.

We still need to implement batch write when snapshoting state, otherwise with thousands of players the DB would get overloaded, but for now it's fine.

The logic is rather custom to each process, I can't figure out if a generic solution is even possible :(

### DIFF
--- a/documents/dev_designs/tachyon_system.md
+++ b/documents/dev_designs/tachyon_system.md
@@ -63,6 +63,28 @@ sequenceDiagram
     Session ->> Session: wait until timeout
 ```
 
+## Reconnection and state restoration after node restart
+
+When the node goes down, if the site config `tachyon.should-restore-state` is set to true,
+then processes will attempt to store part of their state when terminating with `:shutdown` reason.
+At startup, these snapshots are then used to recreate the process and bootstrap them.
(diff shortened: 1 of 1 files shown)


lib/teiserver/tachyon/system.ex
```diff
@@ -21,4 +22,46 @@ defmodule Teiserver.Tachyon.System do
...
+        end)
+        |> Task.await_many()
+        |> Enum.count()
+
+      Logger.info("Restored #{n} blobs from #{store_name} snapshots")
+      :ok
+    after
+      Enum.map(blobs, fn b -> {b.store, b.key} end)
+      |> Teiserver.KvStore.delete_many()
+    end
```
geekingfrog, line 65: the `after` block is meant to always run, so that if the snapshots are corrupted, or invalid in some other ways, the server will not stay in a broken state.
We can always restart from a clean and fresh state.

PR #923, opened 2026-01-03
Metric improvements

geekingfrog:
See commit description for wtf is going on.


lib/teiserver/monitoring/buckets.ex
```diff
@@ -0,0 +1,94 @@
...
+    end
+  end
+
+  # hack since :gb_trees.larger/2 is only available from otp 27
+  defp larger(key, tree) do
+    case :gb_trees.iterator_from(key, tree) |> :gb_trees.next() do
+      :none -> :none
+      {k, v, _iter} -> {k, v}
+    end
+  end
```
geekingfrog, line 93: this should just be `:gb_trees.larger` when we are on otp 27

PR #930, opened 2026-01-05
Fix Dialyzer warnings in `Teiserver.Account.AccoladeBotServer`

dmarcoux:
Changes in other modules were needed to fix those warnings.


lib/teiserver/account/accolades/accolade_bot_server.ex
```diff
@@ -167,7 +167,7 @@ defmodule Teiserver.Account.AccoladeBotServer do
     {:noreply, state}
   end
 
-  @spec get_accolade_account() :: Teiserver.Account.CacheUser.t()
+  @spec get_accolade_account() :: Teiserver.Account.User.t()
```
dmarcoux, line 170: `Teiserver.CacheUser.t()` could have been the fix, but it accepts any map while `Teiserver.Account.User.t()` enforces a user.

lib/teiserver/account.ex
```diff
@@ -26,16 +26,16 @@ defmodule Teiserver.Account do
   @spec count_users(list) :: integer
   defdelegate count_users(args), to: UserLib
 
-  @spec get_user!(non_neg_integer()) :: User.t()
+  @spec get_user!(T.userid()) :: User.t()
```
dmarcoux, line 29: Cleaned up those typespecs to be consistent and more accurate while having to change one of them to fix warnings.

lib/teiserver/battle.ex
```diff
@@ -84,8 +84,8 @@ defmodule Teiserver.Battle do
       ** (Ecto.NoResultsError)
 
   """
-  @spec get_match!(Integer.t() | List.t()) :: Match.t()
-  @spec get_match!(Integer.t(), List.t()) :: Match.t()
+  @spec get_match!(T.match_id() | keyword()) :: Match.t()
```
dmarcoux, line 87: Cleaned up those typespecs to be consistent and more accurate while having to change one of them to fix warnings.

lib/teiserver/account.ex
```diff
@@ -26,16 +26,16 @@ defmodule Teiserver.Account do
...
   defdelegate get_user!(user_id), to: UserLib
 
-  @spec get_user!(non_neg_integer(), list) :: User.t() | nil
+  @spec get_user!(T.userid(), keyword()) :: User.t() | nil
   defdelegate get_user!(user_id, args), to: UserLib
 
-  @spec get_user(non_neg_integer()) :: User.t() | nil
+  @spec get_user(T.userid() | nil) :: User.t() | nil
   defdelegate get_user(user_id), to: UserLib
```
L-e-x-o-n: `get_user` should be called with at least user id or some additional arguments by which the user can be found otherwise we run into [this](https://github.com/beyond-all-reason/teiserver/pull/912) issue, therefore `nil` user id shouldn't be allowed in `get_user/1`

dmarcoux: Good catch, this makes sense. I pushed to fix this.

PR #935, opened 2026-01-06
Fix nil lobby when checking for restrictions

geekingfrog:
```
** (BadMapError) expected a map, got:

    nil

    (elixir 1.19.4) lib/map.ex:541: Map.get(nil, :name, nil)
    (teiserver 0.1.0) lib/teiserver/lobby/libs/lobby_restrictions.ex:289: Teiserver.Lobby.LobbyRestrictions.allowed_to_set_restrictions/1
    (teiserver 0.1.0) lib/teiserver/coordinator/consul_commands.ex:758: Teiserver.Coordinator.ConsulCommands.handle_command/2
    (teiserver 0.1.0) lib/teiserver/coordinator/consul_server.ex:362: Teiserver.Coordinator.ConsulServer.handle_info/2
    (stdlib 5.2.3) gen_server.erl:1095: :gen_server.try_handle_info/3
    (stdlib 5.2.3) gen_server.erl:1183: :gen_server.handle_msg/6
    (stdlib 5.2.3) proc_lib.erl:241: :proc_lib.init_p_do_apply/3
```

### DIFF
--- a/lib/teiserver/coordinator/consul_server.ex
+++ b/lib/teiserver/coordinator/consul_server.ex
@@ -925,7 +925,7 @@ defmodule Teiserver.Coordinator.ConsulServer do
 
   def is_on_friendlist?(userid, state, :all) do
     member_ids =
-      Battle.get_lobby(state.lobby_id)
+      (Battle.get_lobby(state.lobby_id) || %{})
       |> Map.get(:players, [])
 
     # If battle has no players it'll succeed regardless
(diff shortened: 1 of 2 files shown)


L-e-x-o-n:
> Because it's ultimately a distributed system. The lobby just went boom/closed but the player process hasn't learn of that yet. The core problem is that the caller (player) does the check, when it should be done within the lobby process. Properly handling that is going to be a large piece of work I'm afraid. Though in this case, it should be fine since when the lobby doesn't exist, we return false and nothing should happen anyway.

The check is done by the consul server which should be tied to the lobby, not the player. Speaking of which, I am not sure why they aren't linked to the lobby process so that when the lobby goes down it brings down consul with it, that way this message to set restriction would never be received and this error wouldn't happen.

PR #937, opened 2026-01-07
Add default redirect uris to localhost for tachyon setup

### LINKED ISSUE #890: [Bug]: Generation of oauth app does not enter redirect_uri
### Describe the Bug

I cloned the repository and perform the setup to be able to locally run the server with the new-client (bar-lobby). If i try to login i was redirected to the website of the server and i see the message "bad request: Cannot use this application to create auth token". 

Solution to fix it was to debug the correct redirect_uri via terminal and enter this into the generic_lobby.
=> This shall do the generation script automatically that developers has not enter the correct uri manually.

### Reproduce the bug

1. Clone repositories (server & new-client)
2. Perform setup instructions
3. Try to login via new-client (Check generic_lobby via webinterface > admin > oauth app: The redirect uri input field should be empty)

### Screenshots

Image

### Additional context

_No response_

TimMeissner:
Closes #890

### DIFF
--- a/lib/teiserver/tachyon/tasks/setup_apps.ex
+++ b/lib/teiserver/tachyon/tasks/setup_apps.ex
@@ -17,9 +17,7 @@ defmodule Teiserver.Tachyon.Tasks.SetupApps do
       owner_id: root.id,
       scopes: ["tachyon.lobby"],
       redirect_uris: [
-        "http://localhost/oauth2callback",
-        "http://127.0.0.1/oauth2callback",
-        "http://[::1]/oauth2callback"
+        "http://localhost/oauth2callback"
       ],
       description: "To support autohost and players using tachyon."
     })


lib/teiserver/tachyon/tasks/setup_apps.ex
```diff
@@ -16,6 +16,11 @@ defmodule Teiserver.Tachyon.Tasks.SetupApps do
       uid: "generic_lobby",
       owner_id: root.id,
       scopes: ["tachyon.lobby"],
+      redirect_uris: [
+        "http://localhost/oauth2callback",
+        "http://127.0.0.1/oauth2callback",
+        "http://[::1]/oauth2callback"
+      ],
```
geekingfrog, line 21: You only need the first one. Every way of specifying localhost are considered the same for oauth. See [this bit](https://github.com/geekingfrog/teiserver/blob/86659ece4e2d2e26625d60e813f1b0a698e1ea2b/lib/teiserver/o_auth.ex#L92-L100)

PR #939, opened 2026-01-07
Fix wrong typespecs for Teiserver.Room

dmarcoux:
See comments below for details on the changes.

### DIFF
--- a/lib/teiserver/chat/chat_room_cache.ex
+++ b/lib/teiserver/chat/chat_room_cache.ex
@@ -9,8 +9,6 @@ defmodule Teiserver.Room do
   @dont_log_room ~w(autohosts)
 
   @spec create_room(map()) :: map()
-  @spec create_room(String.t(), T.userid()) :: map()
-  @spec create_room(String.t(), T.userid(), T.clan_id()) :: map()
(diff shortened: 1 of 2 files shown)


lib/teiserver/chat/chat_room_cache.ex
```diff
@@ -9,8 +9,6 @@ defmodule Teiserver.Room do
   @dont_log_room ~w(autohosts)
 
   @spec create_room(map()) :: map()
-  @spec create_room(String.t(), T.userid()) :: map()
```
dmarcoux, line 12: Those typespecs are defined for the wrong arity.

lib/teiserver/chat/chat_room_cache.ex
```diff
@@ -267,7 +266,7 @@ defmodule Teiserver.Room do
     end
   end
 
-  @spec allow?(map(), String.t()) :: boolean()
+  @spec allow?(T.userid(), String.t()) :: boolean()
```
dmarcoux, line 269: This is clearly a user id, not a map. See the calls to `CacheUser.is_shadowbanned?/1` and `CacheUser.is_restricted?/2`.

PR #940, opened 2026-01-08
Remove `TeiserverWeb.Telemetry` supervisor as it produces nothing 

dmarcoux:
This supervisor runs its function `periodic_measurements/0` every 10 seconds, but this produces nothing as it returns an empty list.

The `TeiserverWeb.Telemetry.metrics/0` must stay as they are used in the live dashboard at /logging/live/dashboard.

The tuples in `Teiserver.Telemetry.TelemetryLib.periodic_measurements/0` have been commented out for 2+ years now, see the commit 80ae15fd784e32fe125bf47e2f6044fad68a84d8.

I also fixed a Dialyzer warning while working in `Teiserver.Telemetry.TelemetryLib`.

### DIFF
--- a/.dialyzer_ignore.exs
+++ b/.dialyzer_ignore.exs
@@ -125,7 +125,6 @@
   {"lib/teiserver/telemetry/libs/simple_lobby_event_type_lib.ex", :invalid_contract},
   {"lib/teiserver/telemetry/libs/simple_match_event_type_lib.ex", :invalid_contract},
   {"lib/teiserver/telemetry/libs/simple_server_event_type_lib.ex", :invalid_contract},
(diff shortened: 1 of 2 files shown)


lib/teiserver_web/telemetry.ex
```diff
@@ -10,9 +10,6 @@ defmodule TeiserverWeb.Telemetry do
   @impl true
   def init(_arg) do
     children = [
```
dmarcoux: There are no children anymore for this supervisor, I wonder if it could be removed. What do you think? This doesn't have to happen in this PR though.

geekingfrog: might as well remove it while you're at it.

PR #946, opened 2026-01-10
Add a season selector to match page

### LINKED ISSUE #945: [Feature]: Add a season selector to match page
### Describe the Feature

When going to [the match page](https://server4.beyondallreason.info/battle/ratings) it shows the matches **for the current season**.
It could be nice to be able to see matches from previous season as well. It's especially jarring when we change the season.

### Expected Behavior

Add a dropdown to select the season to show matches for, which default to the current season.
only matches for the selected season should show up.

dmarcoux:
Closes #945

### DIFF
--- a/assets/scss/custom/_styles.scss
+++ b/assets/scss/custom/_styles.scss
@@ -404,3 +404,7 @@ input[type="color"] {
 .blog-post-content img {
   max-width: 100%;
 }
+
+.w-fit {
+  width: fit-content;
+}
--- a/lib/teiserver_web/live/battles/match/ratings.ex
+++ b/lib/teiserver_web/live/battles/match/ratings.ex
@@ -7,12 +7,14 @@ defmodule TeiserverWeb.Battle.MatchLive.Ratings do
   @impl true
   def mount(params, _ession, socket) do
     game_types = Battle.MatchLib.list_rated_game_types()
+    active_season = MatchRatingLib.active_season()
+    seasons = Enum.to_list(1..active_season)
 
     user_ratings =
       Account.list_ratings(
         search: [
           user_id: socket.assigns.current_user.id,
-          season: MatchRatingLib.active_season()
+          season: active_season
         ],
         preload: [:rating_type]
       )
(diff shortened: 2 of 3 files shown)


lib/teiserver/mix_tasks/fake_data.ex
```diff
@@ -304,119 +306,126 @@ defmodule Mix.Tasks.Teiserver.Fakedata do
...
-          end)
-
-        Ecto.Multi.new()
-        |> Ecto.Multi.insert_all(
-          :insert_all,
-          Battle.MatchMembership,
-          memberships1 ++ memberships2
-        )
-        |> Teiserver.Repo.transaction()
+    match_ids =
```
dmarcoux, line 312: The changes here are much easier to review if you hide whitespace in the GitHub UI:

Image

lib/teiserver/mix_tasks/fake_data.ex
```diff
@@ -304,119 +309,128 @@ defmodule Mix.Tasks.Teiserver.Fakedata do
...
+          )
+          |> Teiserver.Repo.transaction()
+
+          match.id
+        end)
       end)
-    end)
 
-    Teiserver.Game.MatchRatingLib.reset_and_re_rate("all")
+    Teiserver.Game.MatchRatingLib.re_rate_specific_matches(match_ids)
```
L-e-x-o-n, line 429: Why not rerate all like before?

L-e-x-o-n:
> > I'm not very good at liveview either. I think a `<form phx-change="xxx">` with a `<select>` [like so](https://elixirforum.com/t/how-do-i-create-a-select-element-in-a-liveview-form/64348) should get you most of the way?
> 
> This works already, I cannot find though how to refresh the rating types based on the selected season. Depending on the battles played, those rating types might differ. This is why they need to be refreshed.

IMO this is fine even without refreshing the rating type list, maybe that can be added later.

If you really want to do it now I may have an idea but it's a bit tricky. Currently there is no way to tell which rating types the user has logs of in a given season so getting that info will require adding a new query. We now need to recognise when the season filter changed, the simplest solution might be to just add a new event in the season picker form (instead of `filter-update`) and then when handling this new event use the new query to get the rating types, assign them to `:game_types`, pick first if present and pass it to `run_match_query`...

lib/teiserver/mix_tasks/fake_data.ex
dmarcoux: From my understanding of the match ratings, this would have re-rated all matches as if they were in the latest season.

PR #947, opened 2026-01-10
fix dialyzer unknown type warnings

NortySpock:
### LLM Disclosure
I chatted with an LLM to understand more about typespecs, but  also I read the official documentation on typespecs, and all code was written by me.


lib/teiserver/battle/balance/loser_picks.ex
```diff
@@ -21,12 +21,17 @@ defmodule Teiserver.Battle.Balance.LoserPicks do
...
+  @type algorithm_state ::
+          %{
+            group_pairs: list,
+            teams: map,
+            logs: list,
+            solo_players: list,
+            opts: list,
+            remaining_picks: list,
+            solo_players: list,
+            max_teamsize: integer
```
geekingfrog, line 33: the specs are usually written as function call: `list()` `map()` and so on.
max_teamsize is probably also a `non_neg_integer()`

Issue #954, opened 2026-01-14
[Bug]: $min/maxchevlevel restriction has no limits


L-e-x-o-n:
### Describe the Bug

`$minchevlevel` and `$maxchevlevel restrictions have no validation on the entered values so, for example, a chev limit of 1000 or -1000 can be set without issues.

Due to the way the Spring protocol works, chev restriction only have 8 possible values (1 - 8) and values outside this range should not be allowed.

### Reproduce the bug

Use commands `$minratinglevel -1000` or `$maxratinglevel 1000`

PR #955, opened 2026-01-14
adds bounds to minchevlevel command in lobby

ZainMGit:
Links to issue #954 

Only allows minchevlevel command to be within the bounds of 1 - 8. Other values are given the error message "Chev level must be between"

### DIFF
--- a/lib/teiserver/lobby/libs/lobby_restrictions.ex
+++ b/lib/teiserver/lobby/libs/lobby_restrictions.ex
@@ -7,15 +7,11 @@ defmodule Teiserver.Lobby.LobbyRestrictions do
   alias Teiserver.Battle.{BalanceLib, MatchLib}
   alias Teiserver.Battle
 
-  @rank_upper_bound 1000
-  @rank_min_chev_level 1
-  @rank_max_chev_level 8
+  @rank_upper_bound 7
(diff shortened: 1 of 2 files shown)


lib/teiserver/lobby/libs/lobby_restrictions.ex
```diff
@@ -8,10 +8,14 @@ defmodule Teiserver.Lobby.LobbyRestrictions do
...
 
   @rank_upper_bound 1000
+  @rank_min_chev_level 1
+  @rank_max_chev_level 8
   @rating_upper_bound 1000
   @splitter "------------------------------------------------------"
   @spec rank_upper_bound() :: number
   def rank_upper_bound, do: @rank_upper_bound
+  def rank_min_chev_level, do: @rank_min_chev_level
+  def rank_max_chev_level, do: @rank_max_chev_level
```
L-e-x-o-n: `chev` is just `rank + 1` and `rank_upper_bound` already exists, I would just use that then and change it from 1000 to 7. For the lower bound I would either use `rank_lower_bound` or just assume it's always 0 and use 0 directly where needed in consul server.

L-e-x-o-n:
Thanks!

PR #958, opened 2026-01-18
Fix dialyzer warnings 4

NortySpock:
fix two more dialyzer warnings

### DIFF
--- a/lib/teiserver/bridge/bridge_server.ex
+++ b/lib/teiserver/bridge/bridge_server.ex
@@ -41,7 +41,7 @@ defmodule Teiserver.Bridge.BridgeServer do
     send(bridge_pid, message)
   end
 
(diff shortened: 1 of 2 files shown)


lib/teiserver/bridge/bridge_server.ex
```diff
@@ -41,7 +41,7 @@ defmodule Teiserver.Bridge.BridgeServer do
     send(bridge_pid, message)
   end
 
-  @spec send_direct_message(T.user_id(), String.t()) :: :ok | nil
+  @spec send_direct_message(T.userid(), String.t()) :: :ok | nil
```
L-e-x-o-n, line 44: The inconsistency annoys me, half places using `user_id` and the other half `userid`, but that's a topic for another issue.

PR #964, opened 2026-01-23
Autohost session

geekingfrog:
When terminating the server, somehow the websocket process don't terminate with `:shutdown`, so if we want to do graceful shutdown and recovery for autohost, we need the same mechanism as with players: introduce a session process.

This also open the door to track messages sent by the autohost and their ack by battle process, so we can then subscribe to event without missing anything.

### DIFF
--- a/lib/teiserver/autohost/session.ex
+++ b/lib/teiserver/autohost/session.ex
@@ -177,6 +177,32 @@ defmodule Teiserver.Autohost.Session do
     {:keep_state, data}
   end
 
+  def handle_event({:call, from}, {:kill_battle, _battle_id}, _state, data)
+      when data.conn_pid == nil,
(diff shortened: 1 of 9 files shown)


lib/teiserver/autohost/session.ex
```diff
@@ -0,0 +1,269 @@
...
+  alias Teiserver.TachyonBattle
+
+  require Logger
+
+  @default_call_timeout 5000
+
+  @typep battle_data :: %{
+           start_script: Autohost.start_script(),
+           ips: [String.t()],
+           port: non_neg_integer()
```
geekingfrog, line 22: I am not sure we need this data at all in the autohost session's state but for now, I guess it's fine.

Issue #970, opened 2026-01-24
[Feature]: Trottle logins in addition to hard cap on player count


p2004a:
### Describe the Feature

Right now the server has a single tunable, how many people at the same time can be logged in. We don't control how quickly we can get to that number, it looks like it gets there almost instantly. That is problematic because it looks like we suffer from thundering herd issue that prevents server to come up back healthy.

It would be a good feature to be able to control *dynamically* how many new people can join per minute, up to the limit specified in config.

PR #972, opened 2026-01-25
Rework login queue

geekingfrog:
The first commit is the chunky one:

* Use name registration, no need for send/call/cast client and manual
  registration.
* There is only one queue. People put in higher priority queue before
  are now exempt from the queue. There are a few dozen of such people,
  so the impact will be negligible anyway.
* use erlang built in fifo queues
* Do away with the heartbeat logic. Instead, monitor the process and
  purge users that have left when their turn in the queue show up.
  We are dealing with a few thousands users at most anyway, so it's
  unlikely this process will consume all memory.

Also add a rate limiter on the queue so that we don't need to manually raise the limit when we restart the server.

I've tested it with a few local chobby instances and it seems to work fine.

There is still a flaw in this approach: if the login queue process goes down, the client process don't know about it and will not try to re-register. Thus, until the client fully reconnect they will be in limbo, never able to login.

Fix https://github.com/beyond-all-reason/teiserver/issues/970

### DIFF
--- a/test/teiserver/account/login_throttle_server_test.exs
+++ b/test/teiserver/account/login_throttle_server_test.exs
@@ -13,6 +13,9 @@ defmodule Teiserver.Account.LoginThrottleServerTest do
 
   setup _ctx do
     LoginThrottleServer.restart()
+    :ok = Supervisor.terminate_child(Teiserver.Supervisor, Teiserver.ClientRegistry)
+    Supervisor.restart_child(Teiserver.Supervisor, Teiserver.ClientRegistry)
+
     LoginThrottleServer.set_tick_period(:infinity)
     # ensure the rate limiter doesn't interfere with any tests
     LoginThrottleServer.reset_rate_limiter(1000, true)
@@ -98,7 +101,7 @@ defmodule Teiserver.Account.LoginThrottleServerTest do
     assert user2.id == id
   end
 
-  test "respect login throttle" do
+  test "respect rate limiter" do
     set_capacity(0)
     LoginThrottleServer.reset_rate_limiter(1)
     {user1, t1} = {new_user(), oneshot_pid()}


lib/teiserver/libs/teiserver_configs.ex
```diff
@@ -299,6 +299,17 @@ defmodule Teiserver.TeiserverConfigs do
       value_label: ""
     })
 
+    add_site_config_type(%{
+      key: "system.Login rate",
+      section: "Login",
+      type: "integer",
+      permissions: ["Admin"],
+      description: "How many user per minutes should be able to log in",
+      default: 200,
```
geekingfrog, line 308: this may be a bit optimistic, though we can always change it from the web admin UI.

lib/teiserver/account/servers/login_throttle_server.ex
```diff
@@ -1,480 +1,225 @@
...
     cond do
       CacheUser.is_bot?(user) -> :instant
-      CacheUser.has_any_role?(user, ["Server"]) -> :instant
-      CacheUser.is_moderator?(user) -> :moderator
-      CacheUser.has_any_role?(user, ["Core"]) -> :core
-      CacheUser.has_any_role?(user, ["Contributor"]) -> :contributor
-      CacheUser.has_any_role?(user, ["Overwatch", "Reviewer"]) -> :volunteer
-      CacheUser.has_any_role?(user, ["VIP", "BAR+"]) -> :vip
+      CacheUser.is_moderator?(user) -> :instant
+      CacheUser.has_any_role?(user, bypass_roles) -> :instant
```
L-e-x-o-n: All `Server` and `Core` are `Contributors`, `Moderators` as well, a `has_any_role?` check with `Bot`, `Contributor`, `VIP` and `BAR+` should be enough.

mix.exs
```diff
@@ -10,7 +10,7 @@ defmodule Teiserver.MixProject do
       package: package(),
       dialyzer: dialyzer(),
       elixirc_paths: elixirc_paths(Mix.env()),
-      elixirc_options: [warnings_as_errors: true],
+      # elixirc_options: [warnings_as_errors: true],
```
L-e-x-o-n: Does a warning appear? Which one?

lib/teiserver/account/servers/login_throttle_server.ex
geekingfrog: done

mix.exs
geekingfrog: committed by mistake. I often disable warnings when I'm developing/debugging and sometimes forget it's there. I've reverted that.

Issue #847, opened 2025-11-01
[Bug]: Stop dumping passwords, token and other sensitive information in logs.


p2004a:
### Describe the Bug

Sensitive data all across teiserver logs.

It's cool we store hashed passwords in database, but that doesn't help if plenty are just in plaintext in logs.

### Reproduce the bug

Just look at teiserver log.

### Additional context

- When login or registration request fails -> it dumps raw message into logs
- When some request fails -> it often dumps raw sensitive cookies in the logs
- When websocket connects -> e.g. I see `CONNECTED TO TeiserverWeb.UserSocket` it dumps raw access token in the logs

L-e-x-o-n:
When I looked into this briefly in the past I found out it should be possible with [Logger parameter filtering](https://hexdocs.pm/phoenix/Phoenix.Logger.html#module-parameter-filtering) but in my tests it didn't seem to work as expected.

Additionally, in the docs it says:
> Phoenix's default is ["password"]

so I would expect it to filter out `password` field from log messages which is not the case.

L-e-x-o-n:
Today I learned about https://hexdocs.pm/ecto/Ecto.Schema.html#module-redacting-fields

thvl3:
fixed by #977

PR #976, opened 2026-01-28
Initial implementation of SafeLogger framework

thvl3:
## Summary

Step towards resolving #847 - prevents sensitive data from leaking in logs.

Added `redact: true` to sensitive fields so they display as `**redacted**` when inspected or logged:

- `Account.UserToken.value` - user session tokens
- `Account.Code.value` - password reset codes  
- `Account.TOTP.secret` - TOTP 2FA secrets
- `OAuth.Token.value` - OAuth access/refresh tokens
- `OAuth.Code.value` - OAuth authorization codes
- `OAuth.Code.challenge` - PKCE challenges
- `OAuth.Credential.hashed_secret` - client secret hashes

Phoenix Parameter Filtering -
Configured `:filter_parameters` in `config.exs` to filter sensitive parameter names from Phoenix request logs (password, token, secret, api_key, cookie, authorization, etc.)

Reduced WebSocket connection log verbosity to `:debug` level to avoid logging connection params that may contain tokens.

### DIFF
--- a/lib/teiserver/account/schemas/code.ex
+++ b/lib/teiserver/account/schemas/code.ex
@@ -3,7 +3,7 @@ defmodule Teiserver.Account.Code do
   use TeiserverWeb, :schema
 
   typed_schema "account_codes" do
-    field :value, :string
+    field :value, :string, redact: true
     # E.g. password reset
     field :purpose, :string
     field :metadata, :map
--- a/lib/teiserver/account/schemas/user_totp.ex
+++ b/lib/teiserver/account/schemas/user_totp.ex
@@ -6,7 +6,7 @@ defmodule Teiserver.Account.TOTP do
 
   typed_schema "teiserver_account_user_totps" do
     belongs_to :user, Teiserver.Account.User, define_field: false, type: :id
-    field :secret, :binary
+    field :secret, :binary, redact: true
     field :last_used, :utc_datetime, default: nil
     field :wrong_otp, :integer, default: 0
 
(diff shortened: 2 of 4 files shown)


lib/teiserver/logging/lib/safe_logger.ex
```diff
@@ -0,0 +1,202 @@
...
+    Logger.warning(sanitize(message))
+  end
+
+  @doc """
+  Logs an error message after sanitizing sensitive data.
+  """
+  @spec error(String.t() | iodata()) :: :ok
+  def error(message) do
+    Logger.error(sanitize(message))
+  end
```
geekingfrog, line 101: I am pretty sure a wrapper around the logger function is **not** what you want to do. These are actually [macro](https://github.com/elixir-lang/elixir/blob/main/lib/logger/lib/logger.ex#L1123) so that they can access meta information like caller mfa and so on.
If you wrap that with functions, you lose the original caller info (and whatever other things the logging macro is doing, haven't dug too deep there).

lib/teiserver/logging/lib/safe_logger.ex
```diff
@@ -0,0 +1,202 @@
+defmodule Teiserver.Logging.SafeLogger do
```
geekingfrog, line 1: Not a fan of the name, just `Teiserver.Logger` is probably better. If we want to modify it further with things that are not about redacting stuff, it's going to get awkward.

L-e-x-o-n:
Not sure how useful or necessary the `SafeLogger` module is, the main cause of #847 are in error logs due to exceptions which won't go through `SafeLogger`. I think this *should* be fixable by logger config and reactions alone.

Issue #950, opened 2026-01-11
[Bug]: microblog rss feed throwing error and not returning xml


NortySpock:
### Describe the Bug

RSS feed not in xml format, shows red error message

### Reproduce the bug

1) open the microblog ( at https://server4.beyondallreason.info/microblog/rss ) 
2) click the orange RSS feed button
3) read error message

### Screenshots

<img width="945" height="542" alt="Image" src="https://github.com/user-attachments/assets/7ab8147a-5aef-4c78-a978-d8a59785ad66" />

### Additional context

Discovered when manually testing dialyzer fixes on the microblog

PR #981, opened 2026-01-30
Fixx RSS feed parsing issue

thvl3:
Fixes https://github.com/beyond-all-reason/teiserver/issues/950 locally

Added inline parsing and string casting, seemed to be breaking from ampersand signs found in discord mentions for @ l egion. Added a test that creates a post with the characters and checks for proper sanitization. 
On the left is local, right is current site
image

**Note: Claude Code was used to help write the test**


L-e-x-o-n:
Thanks

Issue #818, opened 2025-10-06
[Feature]: Allow creation of very long live oauth credentials through the API


geekingfrog:
### Describe the Feature

We have an admin API that can create test users (see https://github.com/beyond-all-reason/teiserver/issues/676).
This is typically used by the [mix task](https://github.com/beyond-all-reason/teiserver/pull/734) to dump the data into a file that can then be used by other tools.

But by default the credentials are only valid for a short period of time and it doesn't work well with load testing tool for example.

### Expected Behavior

Add an optional parameter to ask for a different duration before token expiration.

PR #982, opened 2026-01-30
Customizable access token TTL in OAuth module added

thvl3:
Ref https://github.com/beyond-all-reason/teiserver/issues/818
- Added support for a configurable access token TTL in minutes, defauts to 30 minutes.
- Updated `create_user_token` function to accept token options, including the new TTL parameter.
- Enhanced user creation and token refresh endpoints to handle `access_token_ttl` from request parameters.
- Added tests using **Claude Code** to verify behavior for custom, default, and invalid TTL values.

New optional parameter to API: `access_token_ttl`(int, seconds) 
Defaults to 30 minutes or 1800 seconds. (Open for suggestions on defaults)
Currently accepts any positive integer input

From a security perspective I can understand the need for this but I would definitely prefer to see some absolute cap on TTL of 30 days or even 2 weeks. Should be simple to add but open to feedback here.

Example request:
```json
{
  "name": "loadtest_user",
  "email": "loadtest@example.com",
  "password": "password123",
  "access_token_ttl": 86400
}
```

**Claude Code used to write and debug tests**

### DIFF
--- a/lib/teiserver/o_auth.ex
+++ b/lib/teiserver/o_auth.ex
@@ -185,8 +185,12 @@ defmodule Teiserver.OAuth do
             :scopes => Application.scopes(),
             optional(:original_scopes) => Application.scopes()
           },
-          create_refresh: boolean() | options(),
-          scopes: Application.scopes()
+          [
+            {:create_refresh, boolean()}
+            | {:scopes, Application.scopes()}
+            | {:now, DateTime.t()}
+            | {:access_token_ttl, Timex.Duration.t()}
+          ]
         ) ::
           {:ok, Token.t()} | {:error, :invalid_scope | Ecto.Changeset.t()}
   def create_token(user_id, application, opts \\ [])
@@ -206,8 +210,7 @@ defmodule Teiserver.OAuth do
   defp do_create_token(owner_attr, application, opts) do
     now = Keyword.get(opts, :now, DateTime.utc_now())
     scopes = opts[:scopes]
-    # Allow custom access token TTL in minutes, default to 30 minutes
-    access_token_ttl = Keyword.get(opts, :access_token_ttl, 30)
+    access_token_ttl = Keyword.get(opts, :access_token_ttl, Timex.Duration.from_minutes(30))
 
     if Enum.empty?(scopes) ||
          not MapSet.subset?(MapSet.new(scopes), MapSet.new(application.scopes)) do
@@ -219,7 +222,7 @@ defmodule Teiserver.OAuth do
           application_id: application.id,
           scopes: scopes,
           original_scopes: Map.get(application, :original_scopes, application.scopes),
-          expires_at: Timex.add(now, Timex.Duration.from_minutes(access_token_ttl)),
+          expires_at: Timex.add(now, access_token_ttl),
           type: :access
         }
         |> Map.merge(owner_attr)
--- a/lib/teiserver_web/controllers/api/admin/user_controller.ex
+++ b/lib/teiserver_web/controllers/api/admin/user_controller.ex
@@ -18,20 +18,22 @@ defmodule TeiserverWeb.API.Admin.UserController do
     :user_not_found => 404,
     :app_not_found => 400,
     :invalid_scope => 400,
+    :invalid_ttl => 400,
     :default => 400
   }
 
   @error_messages %{
     :user_not_found => "User not found",
     :app_not_found => @app_not_found_error,
-    :invalid_scope => "Invalid scope for OAuth application"
+    :invalid_scope => "Invalid scope for OAuth application",
+    :invalid_ttl => "access_token_ttl must be a positive integer (seconds)"
   }
 
   def create(conn, params) do
     params = Map.merge(@default_params, params)
-    token_opts = build_token_opts(params)
 
-    with {:ok, user} <- Account.script_create_user(params),
+    with {:ok, token_opts} <- build_token_opts(params),
+         {:ok, user} <- Account.script_create_user(params),
          :ok <- update_user_stats(user.id, params),
          {:ok, app} <- get_generic_lobby_app(),
          {:ok, token} <- create_user_token(user, app, token_opts) do
@@ -42,9 +44,8 @@ defmodule TeiserverWeb.API.Admin.UserController do
   end
 
   def refresh_token(conn, %{"email" => email} = params) do
-    token_opts = build_token_opts(params)
-
-    with {:ok, user} <- get_user_by_email(email),
+    with {:ok, token_opts} <- build_token_opts(params),
+         {:ok, user} <- get_user_by_email(email),
          {:ok, app} <- get_generic_lobby_app(),
          {:ok, token} <- create_user_token(user, app, token_opts) do
       json(conn, build_user_response(user, token))
@@ -103,15 +104,13 @@ defmodule TeiserverWeb.API.Admin.UserController do
   defp build_token_opts(params) do
     case Map.get(params, "access_token_ttl") do
       nil ->
-        []
+        {:ok, []}
 
       ttl_seconds when is_integer(ttl_seconds) and ttl_seconds > 0 ->
-        # Convert seconds to minutes for OAuth.create_token
-        ttl_minutes = div(ttl_seconds, 60)
-        [access_token_ttl: max(ttl_minutes, 1)]
+        {:ok, [access_token_ttl: Timex.Duration.from_seconds(ttl_seconds)]}
 
       _ ->
-        []
+        {:error, :invalid_ttl}
     end
   end
 
--- a/test/teiserver_web/controllers/api/admin/user_controller_test.exs
+++ b/test/teiserver_web/controllers/api/admin/user_controller_test.exs
@@ -328,27 +328,7 @@ defmodule TeiserverWeb.API.Admin.UserControllerTest do
 
       now = DateTime.utc_now()
       diff_seconds = DateTime.diff(token.expires_at, now)
-      assert diff_seconds >= 3300 and diff_seconds <= 3900
-    end
-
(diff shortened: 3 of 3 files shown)


NortySpock:
As a developer who gets distracted or called away to do housework, I would appreciate the timeout being something more like 2 or 4 hours. Long enough that I can get interrupted once or twice and still loop back to my original task.

That said, 30 minutes still seems like a safer default, so maybe if we made the default Oauth token timeout configurable from  `config/dev.exs` ?

lib/teiserver/o_auth.ex
```diff
@@ -206,6 +206,8 @@ defmodule Teiserver.OAuth do
   defp do_create_token(owner_attr, application, opts) do
     now = Keyword.get(opts, :now, DateTime.utc_now())
     scopes = opts[:scopes]
+    # Allow custom access token TTL in minutes, default to 30 minutes
+    access_token_ttl = Keyword.get(opts, :access_token_ttl, 30)
```
geekingfrog: The comment is not great. It also become useless if you make the ttl a `Duration` and not a simple number.
The typespec for `create_token` should also be updated accordingly.

lib/teiserver_web/controllers/api/admin/user_controller.ex
```diff
@@ -86,18 +89,32 @@ defmodule TeiserverWeb.API.Admin.UserController do
...
 
+  defp build_token_opts(params) do
+    case Map.get(params, "access_token_ttl") do
+      nil ->
+        []
+
+      ttl_seconds when is_integer(ttl_seconds) and ttl_seconds > 0 ->
+        # Convert seconds to minutes for OAuth.create_token
+        ttl_minutes = div(ttl_seconds, 60)
+        [access_token_ttl: max(ttl_minutes, 1)]
```
geekingfrog: Useless comment. Also, this is where you should convert to `Duration`.
the match can also be simplified with:
```elixir
case ... do
  ttl_seconds when is_integer(ttl_seconds) and ttl_seconds > 0 -> ...
  _ -> ...
end
```

Also, why do you require a ttl in seconds in the API to then immediately convert that into minute? I'm fine with seconds or minutes, but it should be consistent without any useless conversion.

test/teiserver_web/controllers/api/admin/user_controller_test.exs
```diff
@@ -308,5 +308,135 @@ defmodule TeiserverWeb.API.Admin.UserControllerTest do
...
+      resp = conn |> post(create_user_path(), user_data) |> json_response(200)
+
+      assert resp["credentials"]["access_token"]
+
+      token = Teiserver.OAuth.TokenQueries.get_token(resp["credentials"]["access_token"])
+      assert token != nil
+
+      now = DateTime.utc_now()
+      diff_seconds = DateTime.diff(token.expires_at, now)
+      assert diff_seconds >= 3300 and diff_seconds <= 3900
```
geekingfrog: There is [assert_in_delta](https://hexdocs.pm/ex_unit/main/ExUnit.Assertions.html#assert_in_delta/4). Also, that's a 300s gap, 10s should be plenty enough. If the test takes longer than that we have other problems.

test/teiserver_web/controllers/api/admin/user_controller_test.exs
```diff
@@ -308,5 +308,135 @@ defmodule TeiserverWeb.API.Admin.UserControllerTest do
...
+      resp = conn |> post(create_user_path(), user_data) |> json_response(200)
+
+      assert resp["credentials"]["access_token"]
+
+      token = Teiserver.OAuth.TokenQueries.get_token(resp["credentials"]["access_token"])
+      assert token != nil
+
+      now = DateTime.utc_now()
+      diff_seconds = DateTime.diff(token.expires_at, now)
+      assert diff_seconds >= 82_800 and diff_seconds <= 90_000
```
geekingfrog: This test is entirely useless, it's exactly the same as the previous one with just slightly different numbers.

test/teiserver_web/controllers/api/admin/user_controller_test.exs
```diff
@@ -308,5 +308,135 @@ defmodule TeiserverWeb.API.Admin.UserControllerTest do
...
+      resp = conn |> post(create_user_path(), user_data) |> json_response(200)
+
+      assert resp["credentials"]["access_token"]
+
+      token = Teiserver.OAuth.TokenQueries.get_token(resp["credentials"]["access_token"])
+      assert token != nil
+
+      now = DateTime.utc_now()
+      diff_seconds = DateTime.diff(token.expires_at, now)
+      assert diff_seconds >= 1500 and diff_seconds <= 2100
```
geekingfrog: Same comment as previous.

test/teiserver_web/controllers/api/admin/user_controller_test.exs
```diff
@@ -308,5 +308,135 @@ defmodule TeiserverWeb.API.Admin.UserControllerTest do
...
+      resp = conn |> post(refresh_token_path(), refresh_data) |> json_response(200)
+
+      assert resp["credentials"]["access_token"]
+
+      token = Teiserver.OAuth.TokenQueries.get_token(resp["credentials"]["access_token"])
+      assert token != nil
+
+      now = DateTime.utc_now()
+      diff_seconds = DateTime.diff(token.expires_at, now)
+      assert diff_seconds >= 6900 and diff_seconds <= 7500
```
geekingfrog: :recycle: same

test/teiserver_web/controllers/api/admin/user_controller_test.exs
```diff
@@ -308,5 +308,135 @@ defmodule TeiserverWeb.API.Admin.UserControllerTest do
...
+    end
+
+    test "ignores invalid access_token_ttl values", %{authed_conn: conn} do
```
geekingfrog: The API should flat out reject invalid value (400), not ignore them and use a default.

test/teiserver_web/controllers/api/admin/user_controller_test.exs
```diff
@@ -308,5 +308,135 @@ defmodule TeiserverWeb.API.Admin.UserControllerTest do
...
+
+      token = Teiserver.OAuth.TokenQueries.get_token(resp["credentials"]["access_token"])
+      assert token != nil
+
+      now = DateTime.utc_now()
+      diff_seconds = DateTime.diff(token.expires_at, now)
+      assert diff_seconds >= 1500 and diff_seconds <= 2100
+    end
+
+    test "ignores non-integer access_token_ttl values", %{authed_conn: conn} do
```
geekingfrog: same comment, this should be rejected.

geekingfrog:
> As a developer who gets distracted or called away to do housework, I would appreciate the timeout being something more like 2 or 4 hours. Long enough that I can get interrupted once or twice and still loop back to my original task.
> 
> That said, 30 minutes still seems like a safer default, so maybe if we made the default Oauth token timeout configurable from `config/dev.exs` ?

This is a case where we don't really care about security. This is for the admin API, where ppl should know what they're doing. Also, the usecase is really for load testing (and even then, it's not really required anymore).

lib/teiserver_web/controllers/api/admin/user_controller.ex
thvl3: Ah yeah.. that is dumb. Thanks for the feedback

thvl3:
Closed, feature no
longer needed

lib/teiserver/o_auth.ex
```diff
@@ -185,8 +185,12 @@ defmodule Teiserver.OAuth do
             :scopes => Application.scopes(),
             optional(:original_scopes) => Application.scopes()
           },
-          create_refresh: boolean() | options(),
-          scopes: Application.scopes()
+          [
+            {:create_refresh, boolean()}
+            | {:scopes, Application.scopes()}
+            | {:now, DateTime.t()}
+            | {:access_token_ttl, Timex.Duration.t()}
```
geekingfrog, line 192: I would rather have the standard lib `Datetime.t()`. Timex is a rather big library, and we are barely using it for things not provided by the standard lib.
Timex causes some issues (warnings) when upgrading otp versions.

lib/teiserver/o_auth.ex
```diff
@@ -206,6 +210,7 @@ defmodule Teiserver.OAuth do
   defp do_create_token(owner_attr, application, opts) do
     now = Keyword.get(opts, :now, DateTime.utc_now())
     scopes = opts[:scopes]
+    access_token_ttl = Keyword.get(opts, :access_token_ttl, Timex.Duration.from_minutes(30))
```
geekingfrog, line 213: Same comment regarding Timex. You can do `Duration.new!(minute: 30)`

lib/teiserver_web/controllers/api/admin/user_controller.ex
```diff
@@ -86,18 +90,30 @@ defmodule TeiserverWeb.API.Admin.UserController do
...
     )
   end
 
+  defp build_token_opts(params) do
+    case Map.get(params, "access_token_ttl") do
+      nil ->
+        {:ok, []}
+
+      ttl_seconds when is_integer(ttl_seconds) and ttl_seconds > 0 ->
+        {:ok, [access_token_ttl: Timex.Duration.from_seconds(ttl_seconds)]}
```
geekingfrog, line 110: Timex again

lib/teiserver/o_auth.ex
NortySpock: Would it make sense to have a task (issue, ticket) to remove Timex?

geekingfrog: Maybe yeah. There may be some places where the standard lib has no equivalent, I'm unsure. Feel free to create a low priority issue for that.

PR #986, opened 2026-01-31
Consider only players for client count in login queue

geekingfrog:
Not super sure how to get the player count though :/


L-e-x-o-n:
Should be called user count (maybe non-bot user count) because `players` means actual in game players in the rest of teiserver.

Issue #962, opened 2026-01-20
[Bug]: Sending a message over web live chat crashes the lobby


L-e-x-o-n:
### Describe the Bug

Selecting "Chat" in the Teiserver web lobby list or "Live chat" from a specific lobby opens up live chat view (`TeiserverWeb.Battle.LobbyLive.Chat`). Moderators and up are allowed to send messages here and it works fine if the user is logged in through lobby client, however if they are not logged in it will crash the lobby.

The reason for this is that SPADS receives a message from an unknown user and disconnects from server.

I see 2 ways to fix this, when the user is not logged in through the lobby client:
 - the message is sent as Coordinator, e.g. "**Coordinator:** Username: hello there"
 - don't allow sending a message at all

### Reproduce the bug

1. Have SPADS hosting a lobby
2. Login to Teiserver web with an account that has at least the Moderator role and navigate to live chat of the hosted lobby
3. Send a message while not being logged in through lobby client
4. Lobby disappears and the SPADS host disconnects

### Additional context

Nothing appears in Teiserver log.
SPADS log:
```
WARNING  - [SPADS] Ignoring SAIDBATTLE command (unknown user: "Lexon")
WARNING  - [SPADS] Disconnecting from lobby server (broken connection detected)
```

PR #991, opened 2026-02-02
962 fix

thvl3:
**Needs Testing** Fix proposal for  #962  - Web live chat messages from moderators crash the lobby when the user is not connected via a game client.

When a moderator sends a message from the web chat interface (/battle/lobbies/:id/chat) without being connected through a game client, `Lobby.say/3` fails because there's no registered client for that user ID. This causes SPADS and the lobby to crash.

Solution
Check if the user has a connected game client before calling `Lobby.say/3`:
If connected: Send the message normally as the user
If not connected: Send via Coordinator with user attribution ( "username (web): message")

I was unable to test this locally with a working SPADS instance. I would like some help verifying if this will fix the issue.

### DIFF
--- a/lib/teiserver_web/live/battles/lobbies/chat.ex
+++ b/lib/teiserver_web/live/battles/lobbies/chat.ex
@@ -144,8 +144,8 @@ defmodule TeiserverWeb.Battle.LobbyLive.Chat do
         if Account.client_exists?(current_user.id) do
           Lobby.say(current_user.id, "web: #{content}", id)
         else
-          coordinator_id = Coordinator.get_coordinator_userid()
-          Lobby.say(coordinator_id, "#{current_user.name} (web): #{content}", id)
+          Coordinator.get_coordinator_userid()
+          |> Lobby.say("#{current_user.name} (web): #{content}", id)
         end
     end
 


lib/teiserver_web/live/battles/lobbies/chat.ex
```diff
@@ -141,7 +141,13 @@ defmodule TeiserverWeb.Battle.LobbyLive.Chat do
         })
 
       true ->
-        Lobby.say(current_user.id, "web: #{content}", id)
+        if Account.client_exists?(current_user.id) do
+          Lobby.say(current_user.id, "web: #{content}", id)
+        else
+          # User not connected via game client, send as Coordinator with attribution. Fixes #962
```
geekingfrog: slop machine detected! this comment is useless. The bug reference could stay, but you can also put it in the commit message.
You should also check if `coordinator_id` isn't nil (and perhaps change the typespec of `get_coordinator_userid`?). The whole thing is racy anyway, so might as well check that.

thvl3: Can try it out, might be a good idea to force a race condition and see if we cant guard against it better.

geekingfrog: how do you force a race?

lib/teiserver_web/live/battles/lobbies/chat.ex
```diff
@@ -141,7 +141,12 @@ defmodule TeiserverWeb.Battle.LobbyLive.Chat do
         })
 
       true ->
-        Lobby.say(current_user.id, "web: #{content}", id)
+        if Account.client_exists?(current_user.id) do
+          Lobby.say(current_user.id, "web: #{content}", id)
+        else
+          coordinator_id = Coordinator.get_coordinator_userid()
+          Lobby.say(coordinator_id, "#{current_user.name} (web): #{content}", id)
```
L-e-x-o-n: You could use `|>` here.

Issue #858, opened 2025-11-13
[Bug]: oauth code and tokens are not hashed


geekingfrog:
### Describe the Bug

OAuth codes, token and refresh token values are stored in plain text in the database.
These should be hashed with Argon2, same way we do for regular password.

PR #995, opened 2026-02-05
Split identifier + verifier pattern proposal

thvl3:
Re:discord -  I've been thinking about a solution to this for about a week now that doesnt cause headaches elsewhere. https://github.com/beyond-all-reason/teiserver/issues/858

Problem 1 is that we need to migrate the old values over with a transition period so we dont invalidate tokens that shouldn't be invalidated. We need to carefully consider a future proof solution to avoid disruption.

Problem 2: Unlike passwords (looked up by username, then verified), tokens need to be looked up by their value. With random salts, you can't query WHERE value = hash(input) because each entry has a different salt. Argon is also slow by hashing standards (compared to sha256 and et al)

My current mockup solution, I went witha  Split identifier + verifier pattern using Argon2 with a fixed salt. New tokens are selector.verifier format - lookup by the indexed selector, then verify the verifier's Argon2 hash. Old tokens remain valid during a 90+ day (or more?) transition period before the legacy column is removed.

All tests are passing locally, migrations will need to be generated and run later after migration procedure is agreed upon.

### DIFF
--- a/lib/teiserver/o_auth/helpers/token_hash.ex
+++ b/lib/teiserver/o_auth/helpers/token_hash.ex
@@ -14,6 +14,16 @@ defmodule Teiserver.OAuth.TokenHash do
   end
 
   @doc """
+  Perform a dummy hash to prevent timing attacks when no record is found.
+  Equivalent in purpose to Argon2.no_user_verify/0 but for SHA-256.
+  """
+  @spec no_user_verify() :: false
+  def no_user_verify do
+    hash_verifier("dummy")
+    false
+  end
+
+  @doc """
   Returns {selector, hashed_verifier, full_token}.
   """
   @spec generate_token() :: {String.t(), String.t(), String.t()}
--- a/lib/teiserver/o_auth/queries/code_query.ex
+++ b/lib/teiserver/o_auth/queries/code_query.ex
@@ -20,7 +20,7 @@ defmodule Teiserver.OAuth.CodeQueries do
 
     case code do
       nil ->
-        TokenHash.hash_verifier(verifier)
+        TokenHash.no_user_verify()
         nil
 
       code ->
--- a/lib/teiserver/o_auth/queries/token_query.ex
+++ b/lib/teiserver/o_auth/queries/token_query.ex
@@ -25,7 +25,7 @@ defmodule Teiserver.OAuth.TokenQueries do
 
     case token do
       nil ->
-        TokenHash.hash_verifier(verifier)
+        TokenHash.no_user_verify()
         nil
 
       token ->
--- a/test/support/fixtures/o_auth/o_auth_fixtures.ex
+++ b/test/support/fixtures/o_auth/o_auth_fixtures.ex
@@ -36,14 +36,14 @@ defmodule Teiserver.OAuthFixtures do
       redirect_uri: List.first(app.redirect_uris),
       challenge: challenge,
       challenge_method: method,
-      _verifier: verifier,
-      _full_code: full_code
+      verifier: verifier,
+      full_code: full_code
     }
   end
 
   def create_code(attrs) do
     code = %Code{} |> Code.changeset(attrs) |> Repo.insert!()
-    %{code | value: attrs[:_full_code]}
+    %{code | value: attrs[:full_code]}
   end
 
   def token_attrs(user_id, application) do
@@ -59,13 +59,13 @@ defmodule Teiserver.OAuthFixtures do
       expires_at: Timex.add(now, Timex.Duration.from_days(60)),
       type: :access,
       refresh_token: nil,
-      _full_token: full_token
+      full_token: full_token
     }
   end
 
   def create_token(attrs) do
     token = %Token{} |> Token.changeset(attrs) |> Repo.insert!()
-    %{token | value: attrs[:_full_token]}
+    %{token | value: attrs[:full_token]}
   end
 
   def credential_attrs(bot, app_id) do
--- a/test/teiserver/o_auth/code_test.exs
+++ b/test/teiserver/o_auth/code_test.exs
@@ -50,7 +50,7 @@ defmodule Teiserver.OAuth.CodeTest do
 
   test "can exchange valid code for token", %{user: user, app: app} do
     assert {:ok, code, attrs} = create_code(user, app)
-    assert {:ok, token} = OAuth.exchange_code(code, attrs._verifier, attrs.redirect_uri)
+    assert {:ok, token} = OAuth.exchange_code(code, attrs.verifier, attrs.redirect_uri)
     assert token.scopes == code.scopes
     assert token.owner_id == user.id
     # the code is now consumed and not available anymore
@@ -60,7 +60,7 @@ defmodule Teiserver.OAuth.CodeTest do
   test "cannot exchange expired code for token", %{user: user, app: app} do
     yesterday = Timex.shift(Timex.now(), days: -1)
     assert {:ok, code, attrs} = create_code(user, app, expires_at: yesterday)
-    assert {:error, :expired} = OAuth.exchange_code(code, attrs._verifier)
+    assert {:error, :expired} = OAuth.exchange_code(code, attrs.verifier)
   end
 
   test "must use valid verifier", %{user: user, app: app} do
--- a/test/teiserver_web/controllers/o_auth/code_controller_test.exs
+++ b/test/teiserver_web/controllers/o_auth/code_controller_test.exs
@@ -10,7 +10,7 @@ defmodule TeiserverWeb.OAuth.CodeControllerTest do
       code: code.value,
       redirect_uri: code.redirect_uri,
       client_id: app.uid,
-      code_verifier: code_attrs._verifier
+      code_verifier: code_attrs.verifier
     }
   end
 


lib/teiserver/o_auth/helpers/token_hash.ex
```diff
@@ -0,0 +1,43 @@
...
+  """
+
+  @spec hash_verifier(String.t()) :: String.t()
+  def hash_verifier(verifier) do
+    Argon2.Base.hash_password(verifier, get_token_salt())
+  end
+
+  @spec verify_verifier(String.t(), String.t()) :: boolean()
+  def verify_verifier(verifier, stored_hash) do
+    Plug.Crypto.secure_compare(hash_verifier(verifier), stored_hash)
```
L-e-x-o-n, line 13: This is fine, but maybe using `Argon2` lib would be better for consistency, it has a function [veirfy_pass](https://hexdocs.pm/argon2_elixir/Argon2.html#verify_pass/2) so you don't even need to provide a hashed verifier.

lib/teiserver/o_auth/queries/code_query.ex
```diff
@@ -1,28 +1,47 @@
...
-    from code in Code, as: :code
+  defp get_code_by_selector_and_verify(selector, verifier) do
+    code =
+      base_query()
+      |> where_selector(selector)
+      |> Repo.one()
+
+    case code do
+      nil ->
+        TokenHash.hash_verifier(verifier)
```
L-e-x-o-n: I would prefer [Argon2.no_user_verify](https://hexdocs.pm/argon2_elixir/Argon2.html#no_user_verify/1) as it makes the purpose clearer.

lib/teiserver/o_auth/queries/token_query.ex
```diff
@@ -1,16 +1,41 @@
...
+      base_query()
+      |> where_selector(selector)
+      |> preload(:application)
+      |> preload(:owner)
+      |> preload(:bot)
+      |> Repo.one()
+
+    case token do
+      nil ->
+        TokenHash.hash_verifier(verifier)
```
L-e-x-o-n: `Argont2.no_user_verify` same comment as above.

priv/repo/migrations/20260204213659_backfill_oauth_token_hashes.exs
```diff
@@ -0,0 +1,74 @@
...
+
+  def down do
+    execute("ALTER TABLE oauth_tokens ALTER COLUMN selector DROP NOT NULL")
+    execute("ALTER TABLE oauth_tokens ALTER COLUMN hashed_verifier DROP NOT NULL")
+    execute("ALTER TABLE oauth_codes ALTER COLUMN selector DROP NOT NULL")
+    execute("ALTER TABLE oauth_codes ALTER COLUMN hashed_verifier DROP NOT NULL")
+
+    execute("UPDATE oauth_tokens SET selector = NULL, hashed_verifier = NULL")
+    execute("UPDATE oauth_codes SET selector = NULL, hashed_verifier = NULL")
+  end
```
L-e-x-o-n: I wouldn't worry about backfilling and supporting old tokens, this is not widely used other than for development purposes, dropping the value column and all the old tokens should be fine IMO. We can inform the few people affected to generate new tokens.

lib/teiserver/o_auth/helpers/token_hash.ex
```diff
@@ -0,0 +1,43 @@
...
+
+  @doc """
+  Returns {selector, verifier, hashed_verifier, full_token}.
+  """
+  @spec generate_token() :: {String.t(), String.t(), String.t(), String.t()}
+  def generate_token do
+    selector = :crypto.strong_rand_bytes(16) |> Base.encode16(case: :lower)
+    verifier = :crypto.strong_rand_bytes(32) |> Base.encode16(case: :lower)
+    {selector, verifier, hash_verifier(verifier), "#{selector}.#{verifier}"}
+  end
```
L-e-x-o-n, line 44: Returning verifier isn't necessary and is only used in tests, I would remove it to prevent accidental misuse.
For the tests a separate function which does return everything could be created (with `test` in name or similar to make the purpose clear), the common generation parts could be extracted to a third private function.

lib/teiserver/o_auth/schemas/code.ex
```diff
@@ -5,7 +5,10 @@ defmodule Teiserver.OAuth.Code do
   alias Teiserver.OAuth
 
   typed_schema "oauth_codes" do
-    field :value, :string, redact: true
+    field :selector, :string
+    field :hashed_verifier, :string, redact: true
+    field :value, :string, redact: true, virtual: true
```
L-e-x-o-n, line 10: Why keep `value`? Even as virtual it shouldn't be necessary.

lib/teiserver/o_auth/helpers/token_hash.ex
```diff
@@ -0,0 +1,48 @@
+defmodule Teiserver.OAuth.TokenHash do
+  @moduledoc """
+  Hashing for OAuth token verifiers using SHA-256.
+  """
+
+  @spec hash_verifier(String.t()) :: String.t()
+  def hash_verifier(verifier) do
+    :crypto.hash(:sha256, verifier) |> Base.encode16(case: :lower)
```
L-e-x-o-n, line 8: Why not Argon2?

test/support/fixtures/o_auth/o_auth_fixtures.ex
```diff
@@ -24,40 +24,48 @@ defmodule Teiserver.OAuthFixtures do
...
       owner_id: user_id,
       application_id: app.id,
       scopes: app.scopes,
       expires_at: Timex.add(now, Timex.Duration.from_minutes(5)),
       redirect_uri: List.first(app.redirect_uris),
       challenge: challenge,
       challenge_method: method,
-      _verifier: verifier
+      _verifier: verifier,
+      _full_code: full_code
```
L-e-x-o-n: Unsure about the `_` here, in Elixir it's used in case something is unused which isn't the case here.

lib/teiserver/o_auth/helpers/token_hash.ex
thvl3: @p2004a  had mentioned maybe using Sha256 instead, Argon2 is relatively slow and could add several hundred ms of delay to the auth pattern

priv/repo/migrations/20260204213659_backfill_oauth_token_hashes.exs
geekingfrog: The one token that will need replacing is the test recoil autohost on the dev server. Everything else is fine (lobby clients will have to re-auth, but that's fine).

lib/teiserver/o_auth/schemas/code.ex
thvl3: From what I think I understand, the virtual `value` field is used in multiple other dowsntream locations to carry the full plaintext token back to callers after creation. IE `authorize_controller.ex` puts `code.value` in the redirect, `code_view.ex` returns `token.value` in the API response, and `gen_token.ex` prints it.

I think removing it here would require refactoring the return type of `create_code/create_token` and all their consumers. Please correct me if im wrong as  I think this is the ecto way to carry non-persistent data in a schema struct.

L-e-x-o-n: You are right, my mistake, I didn't check properly.

>  Please correct me if im wrong as I think this is the ecto way to carry non-persistent data in a schema struct.

You are correct.

lib/teiserver/o_auth/helpers/token_hash.ex
```diff
@@ -0,0 +1,58 @@
...
+  end
+
+  @spec verify_verifier(String.t(), String.t()) :: boolean()
+  def verify_verifier(verifier, stored_hash) do
```
geekingfrog, line 12: I'm not a fan of the naming of this function. If you return a boolean, the convention is more like `valid_verifier?`.
Otherwise, I would do something like `verify_verifier :: :valid | :invalid`

test/teiserver/o_auth/token_hash_test.exs
```diff
@@ -0,0 +1,90 @@
...
+      hash2 = TokenHash.hash_verifier("same_value")
+      assert hash1 == hash2
+    end
+
+    test "different inputs produce different hashes" do
+      hash1 = TokenHash.hash_verifier("value1")
+      hash2 = TokenHash.hash_verifier("value2")
+      refute hash1 == hash2
+    end
+  end
```
geekingfrog, line 41: I'm not sure any of this bring any value tbh. These are really thin wrapper around erlang stdlib functions.
I think the important bit is testing the API: that you can generate and pass the checks.

test/teiserver/o_auth/token_test.exs
```diff
@@ -39,6 +43,27 @@ defmodule Teiserver.OAuth.TokenTest do
...
   end
 
+  test "tokens are created with hashed verifier", %{user: user, app: app} do
+    assert {:ok, token} = OAuth.create_token(user, app, scopes: app.scopes)
+
+    assert token.selector != nil
+    assert token.hashed_verifier != nil
+    assert String.length(token.hashed_verifier) == 64
+    assert String.match?(token.hashed_verifier, ~r/^[0-9a-f]+$/)
+    assert String.contains?(token.value, ".")
```
geekingfrog, line 53: I don't think you should test for that, it's kind of an implementation detail. As long as it's a non empty string and you can use it to auth it's all you need.

geekingfrog:
some small nitpicks but overall good to go

PR #996, opened 2026-02-07
Store script_password in LobbyServer to prevent missing replays

### LINKED ISSUE #438: [Bug]: accountId is not always send to spads
### Describe the Bug

There is a collection of replays (502) that don't have `accountId` set for all players, one example of such replay is https://bar-rts.com/replays/fb72a9663da1ed088c43a3e0f6428f92. It's happening continuously through the years.

### Reproduce the bug

https://bar-rts.com/replays/fb72a9663da1ed088c43a3e0f6428f92 download replay

```
zcat 2024-07-30_23-10-51-524_All\ That\ Glitters\ v2_105.1.1-2511-g747f18b\ BAR105.sdfz | strings | less
```

notice that there is no account id set for player 6 and player 4.

### Screenshots

_No response_

### Additional context

_No response_

thvl3:
Fixes #438

--- 
## Problem
SPADS learns each player's `accountId` through the script_password field sent in the `JOINEDBATTLE` protocol message. When SPADS writes the start script for the engine, it maps `script_password -> accountId`. If a player's script_password is missing, their accountId is not set in the replay.
There are three code paths where script_password was being lost:
1. Login enumeration sends nil for existing players
In `spring_out.ex do_login_accepted`, when any Spring client logs in, all existing lobby players are enumerated with `script_password: nil:`

`spring_out.ex:736-749`
```
      if lobby != nil and Map.has_key?(lobby, :players) do
        lobby.players
        |> Enum.each(fn player_id ->
          msg = %{
            channel: "teiserver_global_user_updates",
            event: :joined_lobby,
            lobby_id: lobby_id,
            client: Account.get_client_by_id(player_id),
            script_password: nil
          }
          send(self(), msg)
        end)
      end
  ```
This meant there was no server-side store to recover passwords from. Problem 1 was unfixable without fixing this first.
3. do_force_add_user_to_lobby overwrites password silently
For non-Spring clients, add_user_to_battle was called twice -- first with the correct password, then again with a new random password. The second call overwrites the stored password in the LobbyServer but doesn't broadcast to the host (the if not Enum.member? guard suppresses it since the user is already added). Result: host has password A, server has password B.

## Fix

`lobby_server.ex`: Add `script_passwords: %{}` to state. Store the password on `:add_user`, delete on `:remove_user`. Expose via `:get_script_passwords` call.
`lobby_lib.ex` / `battle.ex`: Add `get_lobby_script_passwords/1` public API.
`spring_out.ex`: In `do_login_accepted`, look up stored passwords from the LobbyServer instead of sending nil.
`spring_tcp_server.ex`: In the `{:login_event, :add_user_to_battle}` handler, look up the stored password instead of hardcoding nil.
`lobby.ex`: Remove the redundant second `add_user_to_battle` call in `do_force_add_user_to_lobby` that silently corrupted the stored password.

Tests added:
password stored when user joins
password removed when user leaves
distinct passwords for multiple users
password updated on rejoin
auto-generated password stored via 2-arg add_user_to_battle
empty passwords map for new lobby
nil returned for non-existent lob

**Edit for AI Policy: Claude code used for tests and codebase grep.**

### DIFF
--- a/test/teiserver/lobby/script_password_persistence_test.exs
+++ b/test/teiserver/lobby/script_password_persistence_test.exs
@@ -1,10 +1,21 @@
 defmodule Teiserver.Lobby.ScriptPasswordPersistenceTest do
   use Teiserver.ServerCase, async: false
 
-  alias Teiserver.{Battle, Lobby}
+  alias Teiserver.{Battle, Client, Lobby}
+  alias Teiserver.Account.UserCacheLib
+  import Teiserver.Helper.NumberHelper, only: [int_parse: 1]
 
   import Teiserver.TeiserverTestLib,
-    only: [make_lobby: 0, new_user: 0]
+    only: [
+      make_lobby: 0,
+      new_user: 0,
+      auth_setup: 1,
+      auth_setup: 2,
+      _send_raw: 2,
+      _recv_raw: 1,
+      _recv_until: 1,
+      start_spring_server: 1
+    ]
 
   describe "LobbyServer script_password storage" do
     test "stores script_password when user is added to battle" do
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: test/teiserver/lobby/script_password_persistence_test.exs [hunk 1]
```diff
@@ -0,0 +1,106 @@
...
+  import Teiserver.TeiserverTestLib,
+    only: [make_lobby: 0, new_user: 0]
+
+  describe "LobbyServer script_password storage" do
+    test "stores script_password when user is added to battle" do
+      lobby_id = make_lobby()
+      user = new_user()
+
+      Lobby.add_user_to_battle(user.id, lobby_id, "test_password_123")
+      :timer.sleep(100)
```
geekingfrog, line 26: What's the deal with all the 100 ms wait everywhere?
There is `Test.Support.Polling` with a few polling function there that should be preferred do waiting for a fixed duration.

test/teiserver/lobby/script_password_persistence_test.exs
```diff
@@ -0,0 +1,106 @@
...
+  @moduledoc """
+  Tests for issue #438: script_password must be stored in the LobbyServer
+  so it can be recovered when clients (particularly SPADS) reconnect.
+  """
+  use Teiserver.ServerCase, async: false
+
+  alias Teiserver.{Battle, Lobby}
+
+  import Teiserver.TeiserverTestLib,
+    only: [make_lobby: 0, new_user: 0]
```
geekingfrog: It's a bit out of scope, but `make_lobby` doesn't do any cleanup, which can be problematic. There should be a ExUnit callback to terminate the lobby process on test exit.

test/teiserver/lobby/script_password_persistence_test.exs
```diff
@@ -1,10 +1,21 @@
...
-    only: [make_lobby: 0, new_user: 0]
+    only: [
+      make_lobby: 0,
+      new_user: 0,
+      auth_setup: 1,
+      auth_setup: 2,
+      _send_raw: 2,
+      _recv_raw: 1,
+      _recv_until: 1,
+      start_spring_server: 1
```
geekingfrog, line 17: If you are testing protocol related things, the tests should live under `test/teiserver/protocols/spring`.

geekingfrog (commented):
Besides the comments, a few meta ones.

Your code has a distinct llm flavour. We recently added an [AI Policy](https://github.com/beyond-all-reason/teiserver/blob/master/AI_POLICY.md), could you have a read and adjust things accordingly please?

The PR message is good, but I would prefer having a lot of this info into commit messages. Especially the first one, some changes seems mysterious without the context given in the PR description, and this should live in the commit.

I would prefer not having to read commits that merely fix formatting. You should squash/rebase so that the commit history is meaningful, both for reviewers and for historical purposes.

I'm also not very well versed in spring&spads, @L-e-x-o-n maybe you know more about this aspect?

test/teiserver/lobby/script_password_persistence_test.exs
```diff
@@ -0,0 +1,257 @@
...
+      _send_raw(p_socket, "JOINBATTLE #{lobby_id} empty my_script_pw\n")
+      _ = _recv_raw(host_socket)
+      _send_raw(host_socket, "JOINBATTLEACCEPT #{user2.name}\n")
+      _ = _recv_until(p_socket)
+
+      reply = _recv_until(host_socket)
+      assert reply =~ "JOINEDBATTLE #{lobby_id} #{user2.name} my_script_pw"
+    end
+
+    test "script_password persists in LobbyServer after player joins",
```
geekingfrog, line 137: test is failing it seems?

lib/teiserver/lobby.ex
```diff
@@ -182,10 +182,6 @@ defmodule Teiserver.Lobby do
       send(client.tcp_pid, {:force_join_battle, lobby_id, script_password})
     end
 
-    if client != nil and client.protocol != :spring do
-      add_user_to_battle(client.userid, lobby_id)
-    end
-
```
L-e-x-o-n: Why remove this?

PR #1004, opened 2026-02-15
 Addressing issue #1001 lobby name limit

### LINKED ISSUE #1001: [Feature]: Limit the length of lobby names
### Describe the Feature

Add a config or a constant that defines the maximum character count for lobby names.

Should be applied to both Tachyon and Spring protocol lobby names.

Connor-Sandall:
Let me know what you guys think for issue #1001. My attempt at fixing this for both spring and tach. Had a heck of a time with ai trying to teach me elixir for this lol:

made @max_lobby_name_length 100
That's enough to do:
NOOBS v BARBS +50 T4 Eco lemon zest lobby cuh| Unranked | 4v4 | Coop | Min chev: 4 | Max rating: 25
Made changes to the coordinator & protocol changes to utilize lobby length function
Added error checking & testing for it too

### DIFF
--- a/lib/teiserver/tachyon_lobby.ex
+++ b/lib/teiserver/tachyon_lobby.ex
@@ -44,10 +44,22 @@ defmodule Teiserver.TachyonLobby do
     end
   end
 
+  @max_lobby_name_length Teiserver.Lobby.LobbyLib.max_lobby_name_length()
+
   def create(start_params) do
-    with {:ok, %{pid: pid, id: id}} <- TachyonLobby.Supervisor.start_lobby(start_params),
-         {:ok, details} <- Lobby.get_details(id) do
-      {:ok, pid, details}
+    cond do
+      not is_map_key(start_params, :name) or not is_binary(start_params.name) or
+          String.trim(start_params.name) == "" ->
+        {:error, :empty_lobby_name}
+
+      String.length(start_params.name) > @max_lobby_name_length ->
+        {:error, :lobby_name_too_long}
+
+      true ->
+        with {:ok, %{pid: pid, id: id}} <- TachyonLobby.Supervisor.start_lobby(start_params),
+             {:ok, details} <- Lobby.get_details(id) do
+          {:ok, pid, details}
+        end
     end
   end
 
--- a/lib/teiserver/tachyon_lobby/lobby.ex
+++ b/lib/teiserver/tachyon_lobby/lobby.ex
@@ -1461,14 +1461,20 @@ defmodule Teiserver.TachyonLobby.Lobby do
     }
   end
 
+  @max_lobby_name_length Teiserver.Lobby.LobbyLib.max_lobby_name_length()
+
   @spec update_property(atom(), term(), state()) :: {:ok, [event()]} | {:error, String.t()}
   defp update_property(:name, new_name, _state) do
-    # we can expand lobby name validation later
-    if new_name == "" do
-      {:error, "name must not be empty"}
-    end
+    cond do
+      new_name == "" ->
+        {:error, "name must not be empty"}
 
-    {:ok, [{:update_lobby_name, new_name}]}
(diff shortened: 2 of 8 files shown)


lib/teiserver/tachyon_lobby.ex
```diff
@@ -44,10 +44,22 @@ defmodule Teiserver.TachyonLobby do
     end
   end
 
+  @max_lobby_name_length Teiserver.Lobby.LobbyLib.max_lobby_name_length()
```
geekingfrog, line 47: I would not use anything from `LobbyLib` for tachyon lobbies, redefine the same constant there instead.
Because ultimately, the tachyon stuff will completely replace the existing lobby code, and if there is zero sharing it'll be easier to clean up when the time comes.

lib/teiserver/tachyon_lobby/lobby.ex
```diff
@@ -1461,14 +1461,20 @@ defmodule Teiserver.TachyonLobby.Lobby do
     }
   end
 
+  @max_lobby_name_length Teiserver.Lobby.LobbyLib.max_lobby_name_length()
```
geekingfrog, line 1464: This should not be redefined. Actually, the max lobby name length should be on this module, and not `Teiserver.TachyonLobby`.

test/teiserver/tachyon_lobby/lobby_test.exs
```diff
@@ -55,6 +55,24 @@ defmodule Teiserver.TachyonLobby.LobbyTest do
     assert details.engine_version == engine.name
   end
 
+  test "create lobby with name at max length" do
+    max = Teiserver.Lobby.LobbyLib.max_lobby_name_length()
+    params = mk_start_params([1, 1]) |> Map.put(:name, String.duplicate("a", max))
+    {:ok, _pid, details} = Lobby.create(params)
+    assert String.length(details.name) == max
+  end
```
geekingfrog, line 63: this test seems rather useless to me. The next one is more interesting.

lib/teiserver/tachyon_lobby/lobby.ex
```diff
@@ -1461,14 +1461,20 @@ defmodule Teiserver.TachyonLobby.Lobby do
...
+
   @spec update_property(atom(), term(), state()) :: {:ok, [event()]} | {:error, String.t()}
   defp update_property(:name, new_name, _state) do
-    # we can expand lobby name validation later
-    if new_name == "" do
-      {:error, "name must not be empty"}
-    end
+    cond do
+      new_name == "" ->
+        {:error, "name must not be empty"}
```
geekingfrog, line 1470: the errors there are strings, but in `create`, they are atoms. You should pick one.

geekingfrog (commented):
for the tachyon stuff, the lobby name logic is duplicated (and also inconsistent). You should have one function doing the check, something like `:ok | {:error, reason :: term()}` and reuse that.
This function should live in the `Teiserver.TachyonLobby.Lobby` module (or `Teiserver.Lobby` would also work).

test/teiserver/tachyon_lobby/lobby_test.exs
```diff
@@ -787,6 +805,22 @@ defmodule Teiserver.TachyonLobby.LobbyTest do
...
     end
 
+    test "name too long rejected" do
+      max = Teiserver.Lobby.LobbyLib.max_lobby_name_length()
+      {:ok, _pid, %{id: id}} = Lobby.create(mk_start_params([2, 2]))
+      long_name = String.duplicate("x", max + 1)
+      {:error, _} = Lobby.update_properties(id, @default_user_id, %{name: long_name})
+    end
+
+    test "name at max length accepted" do
```
geekingfrog, line 815: similar, I don't think checking "at max length" is useful. Checking "max + 1 => error" should be enough.

Issue #969, opened 2026-01-24
[Feature]: Implement votes when changing map in a lobby


geekingfrog:
### Describe the Feature

For the protocol, see [the tachyon PR](https://github.com/beyond-all-reason/tachyon/pull/86).

### Expected Behavior

When a member of the lobby wants to change the map through `lobby/update`:

If there is only 1 current player: the change goes through.
If there is more than 1 player: start a vote.

If the update is for more than just the map, outright reject the entire update (and no vote is started).
❓ I am actually not sure about that. We could also apply other updates and start the vote for the map. If there are multiple attributes requiring a vote, ignore them all but one and start a vote for that one?

If the member that initiate the change is one of the player, automatically cast their vote as `yes`.
Only players are eligible to vote, players in join queue or spectators cannot vote for that. If one of the voter leaves the lobby, consider they voted `abstain`.
After 30s, cancel the vote.
The vote requires absolute majority.

When a vote is pending, no other update requiring a vote can be made. All these updates should be ignored. For now, it's only about changing the map.

PR #1006, opened 2026-02-15
Lobby votes

geekingfrog:
Implement votes for lobbies: https://github.com/beyond-all-reason/teiserver/issues/969

### DIFF
--- a/lib/teiserver/player/session.ex
+++ b/lib/teiserver/player/session.ex
@@ -528,6 +528,12 @@ defmodule Teiserver.Player.Session do
     GenServer.call(via_tuple(user_id), {:lobby, :update_properties, data})
   end
 
+  @spec lobby_vote_submit(T.userid(), String.t(), TachyonLobby.vote_ballot()) ::
+          :ok | {:error, :invalid_lobby | term()}
+  def lobby_vote_submit(user_id, vote_id, ballot) do
+    GenServer.call(via_tuple(user_id), {:lobby, :vote_submit, vote_id, ballot})
+  end
+
   @spec lobby_update_client_status(T.userid(), TachyonLobby.client_status_update_data()) ::
           :ok | {:error, :invalid_lobby | term()}
   def lobby_update_client_status(user_id, data) do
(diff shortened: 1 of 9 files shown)


lib/teiserver/tachyon_lobby/lobby.ex
```diff
@@ -1573,15 +1720,67 @@ defmodule Teiserver.TachyonLobby.Lobby do
...
+
+        voters =
+          for {_, p} <- state.players, !bot_id?(p.id), into: %{} do
+            if p.id == user_id, do: {p.id, :yes}, else: {p.id, :pending}
+          end
+
+        quorum =
+          if rem(map_size(voters), 2) == 0,
+            do: trunc(map_size(voters) / 2 + 1),
+            else: (map_size(voters) / 2) |> :math.ceil() |> trunc()
```
L-e-x-o-n: I would prefer having a config for this quorum and majority but that can be added later.

It seems unnecessarily complex, this should work:
```ex
quorum = voters * 0.51 |> :math.ceil()  |> trunc()
```

lib/teiserver/tachyon_lobby/lobby.ex
```diff
@@ -1573,15 +1720,67 @@ defmodule Teiserver.TachyonLobby.Lobby do
...
+    broadcast_to_members(fsm_data, nil, {:lobby, fsm_data.id, {:vote_ended, vote.id, outcome}})
+  end
+
+  @spec vote_result(vote_state()) :: :undecided | {:ended, :passed | :failed}
+  defp vote_result(vote) do
+    votes = for {_, v} <- vote.voters, do: v
+
+    cond do
+      Enum.count(votes, &(&1 != :pending)) < vote.quorum -> :undecided
+      Enum.count(votes, &(&1 == :yes || &1 == :abstain)) >= vote.majority -> {:ended, :passed}
```
L-e-x-o-n: Why count abstain as yes here,  what's the purpose of having abstain then if it's effectively an alias for yes?
IMO it should work as discussed on Discord and only yes votes should be counted for majority.

lib/teiserver/tachyon_lobby/lobby.ex
geekingfrog: Ended up with `0.501` since `254 * 0.51 = 129.54` which would lead to incorrect calc.

L-e-x-o-n: Looks like you haven't committed this change.

geekingfrog: oh my bad, it's done now

PR #1013, opened 2026-02-28
Require boss/founder status to add AI bots in lobby

hclivess:
## Summary

- **Spring protocol**: Players could add AI bots (Barbarian AI, Scavenger AI, etc.) without boss status — only spectators were blocked. Changed `:add_bot` from a `player_command` to a `boss_command` that checks the `ConsulServer` `host_bosses` list via `Coordinator.call_consul`. Founders, moderators, and the coordinator bypass this check as before.
- **Tachyon protocol**: Added `creator_id` tracking to lobby state and a permission guard clause that rejects non-creators with `{:error, :insufficient_permissions}`.

## Root Cause

In `lib/teiserver/lobby.ex`, the `allow?/3` function classified `:add_bot` as a `player_command`, which only blocked spectators. Any player in the lobby could add bots — the `host_bosses` list from `ConsulServer` was never consulted.

In `lib/teiserver/tachyon_lobby/lobby.ex`, the `add_bot` handler only checked if the user was in the lobby and if the ally team was valid. No permission check existed.

## Test plan

- [ ] Verify moderators/founders can still add bots
- [ ] Verify players with boss status (via `!boss`) can add bots
- [ ] Verify regular players without boss status are rejected when adding bots
- [ ] Verify spectators are still rejected
- [ ] Verify Tachyon lobby creator can add bots
- [ ] Verify Tachyon lobby non-creators are rejected with `:insufficient_permissions`
- [ ] Verify lobby snapshot restore preserves `creator_id`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/lib/teiserver/lobby.ex
+++ b/lib/teiserver/lobby.ex
@@ -755,7 +755,7 @@ defmodule Teiserver.Lobby do
         cmd
       )
 
-    player_command =
+    boss_command =
       Enum.member?(
         [
           :add_bot
@@ -780,7 +780,8 @@ defmodule Teiserver.Lobby do
       mod_command == true ->
         false
 
-      player_command == true and changer.player == false ->
+      # Boss commands require host boss status
+      boss_command == true and not is_host_boss?(changer.userid, battle.id) ->
         false
 
       # If they're not a member they can't do anything either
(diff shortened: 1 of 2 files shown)


lib/teiserver/tachyon_lobby/lobby.ex
geekingfrog, line 1: why tf is this file changed at all? There is no concept of boss yet in tachyon.
If you want to make change with a slop machine, at the very least take the time to understand what it does.

lib/teiserver/lobby.ex
```diff
@@ -780,7 +780,8 @@ defmodule Teiserver.Lobby do
       mod_command == true ->
         false
 
-      player_command == true and changer.player == false ->
+      # Boss commands require host boss status
```
geekingfrog, line 783: useless comment

PR #1024, opened 2026-03-17
Admin UI button to restart Discord Bridge

NortySpock:
### feature
 - Adds button to kill and restart the Discord Bridge Bot 
   - (only displays button if server is configured to enable Discord)
 - restart process is async and sends back an "assumed success, please manually test chat" message

#### commentary
This only really currently has access to the supervisor restart status, it's not going deep into Nostrum to try to get the bridge link status. I could expand it to do that, but maybe that's out of scope when all that's currently needed is "button to thwack to try it again"

### screenshot

`http://localhost:4000/admin/dashboard`

_on success_
image

_when disabled via config_
image

### logs show clear sign of the DiscordBridgeBot restarting every time I click the button

```
2026-03-17 07:29:16.899 [info] request_id=BridgeServer pid=<0.2691.0>  Client connected: DiscordBridgeBot using Teiserver Internal Client
2026-03-17 07:29:17.131 [debug] pid=<0.2687.0>  :READY
2026-03-17 07:29:17.133 [debug] pid=<0.2687.0>  %{
  user: %{
    flags: 0,
    id: 1361886995403509782,
    username: "nortyspock-teiserver-test-bot",
    email: nil,
    bot: true,
    clan: nil,
    verified: true,
    discriminator: "5491",
    avatar: "50c082c36ad401ce6f591ab05d15b9b7",
    global_name: nil,
    mfa_enabled: true,
    primary_guild: nil
  },

```

(another click of the button)

```
2026-03-17 07:37:31.694 [info] request_id=BridgeServer pid=<0.2893.0>  Client connected: DiscordBridgeBot using Teiserver Internal Client
2026-03-17 07:37:32.006 [debug] pid=<0.2888.0>  :READY


```

(restarted test bot still clearly sees me putting messages in the Discord side of the chat room)
image

```
2026-03-17 07:39:35.935 [debug] pid=<0.1642.0> actor_type=chat_room actor_id=main  QUERY OK source="teiserver_room_messages" db=1.7ms queue=0.4ms idle=1257.1ms
INSERT INTO "teiserver_room_messages" ("user_id","inserted_at","chat_room","content") VALUES ($1,$2,$3,$4) RETURNING "id" [40, ~U[2026-03-17 12:39:35Z], "main", "nortyspock: test test test"]

```

### Merge preference
squash-merge

### LLM disclosure
I consulted Qwen3.5:9b and Google Gemini 3 thinking mode while building this. They wrote small snippets and corrections, while I did the architecture.

### DIFF
--- a/test/teiserver/protocols/spring/spring_party_test.exs
+++ b/test/teiserver/protocols/spring/spring_party_test.exs
@@ -191,6 +191,7 @@ defmodule Teiserver.Protocols.Spring.SpringPartyTest do
       assert username == user2.name
     end
 
+    # https://github.com/beyond-all-reason/teiserver/actions/runs/23672656398/job/68969199329?pr=1024
     @tag :needs_attention
     test "can create party after leaving", %{socket1: sock1} do
       leave_party!(sock1)


lib/teiserver/bridge/discord_system.ex
```diff
@@ -21,4 +21,18 @@ defmodule Teiserver.Bridge.DiscordSystem do
...
+    case Process.whereis(Teiserver.Bridge.DiscordSupervisor) do
+      pid when is_pid(pid) ->
+        Supervisor.stop(pid, :normal)
+
+      nil ->
+        :already_down
+    end
+
+    # and then start the task that starts it back up again (if Discord was enabled)
+    init(nil)
```
geekingfrog: nope, never call a callback function like that. I have no idea what calling init like that is going to do.
If you want to shut it down, you have to use `DynamicSupervisor.terminate_child`, then you should call the function you used to start the discord supervisor again, the one with the async task to shield from discord failure. This ensure the restarted discord system is in its proper place in the supervision tree.

Also, the comments are really useless, tell your slop machines to avoid useless comments that describe what the code is doing.
You also don't need the `:already_down` thingy since you just ignore the return value.

lib/teiserver_web/live/admin_dashboard/index.ex
```diff
@@ -109,6 +109,11 @@ defmodule TeiserverWeb.AdminDashLive.Index do
     {:noreply, socket}
   end
 
+  def handle_event("reinit-discord-bridge", _event, socket) do
+    Teiserver.Bridge.DiscordSystem.restart()
```
geekingfrog: This should carry back some form of return value so we can report to the user what's going on, with a flash message for example.

lib/teiserver_web/live/admin_dashboard/index.html.heex
```diff
@@ -40,6 +40,14 @@ policy_bsname = Teiserver.Game.LobbyPolicyLib.colours() %>
   </div>
 
   <div class="col-md-4">
+    <div
+      class="btn btn-info2 mx-2"
+      style="background-color: #7289DA;"
+      phx-click="reinit-discord-bridge"
```
geekingfrog: call the event `restart-discord-bridge`, less confusion with between the text and the action.

lib/teiserver/bridge/discord_system.ex
NortySpock: Ok, I didn't realize that directing calling `init()` was going to be dangerous. I'll shove all the logic we had in there into a `start()` function and just have `init(_)` call `start()`, and `restart()` call `start()` after it shuts down the supervisor.

Is there a particular reason we want to call DynamicSupervisor.terminate_child ? Ordering?  Preventing non-blocking behavior? Or preventing it from trapping exits? Do we need to do this recursively for all children and grandchildren? 

https://hexdocs.pm/elixir/1.19/DynamicSupervisor.html#terminate_child/2  indicates that both `DynamicSupervisor.stop()` and `DynamicSupervisor.terminate_child()` are blocking calls, and so perhaps we wrap that in `Task.async()` for shutdown? And also it notes that `DynamicSupervisor` children are unordered, so I presume the shutdown order of one child can only have one ordering... (the downstream supervisor would have an explicit child ordering.)

I'll take note to avoid redundant comments. As someone who is sometimes scatterbrained I don't mind them, but I understand they are... redundant.

NortySpock:
I attempted switching to start_link() as @L-e-x-o-n  suggested.

If I use start_link() inside restart(), and run `Teiserver.Bridge.DiscordSystem.restart()` I get
`{:error, {:already_started, #PID<0.1109.0>}}` and that PID maps to Teiserver.Bridge.DiscordSystem. So it attempted to link Teiserver.Bridge.DiscordSystem to itself and failed.


If I continue with `init()` inside `restart()`, it works. (though as noted in my comment to GeekingFrog, I will change it.)

Checking the supervision tree, I can see that the supervision tree is indeed DiscordSystem, DiscordSupervisor,  Nostrum.Application, etc.

iex> SupervisorHelper.supervision_tree(Teiserver.Bridge.DiscordSystem)

```
[
  [
    id: :undefined,
    pid: #PID<0.1770.0>,
    modules: [Teiserver.Bridge.DiscordSupervisor],
    supervised: [
      [
        id: Nostrum.Application,
        pid: #PID<0.1771.0>,
        ...

```

Thwacking the restart function (using `init()` (again, I will change it) ), I can see that the pids change, suggesting that the system works.

iex> SupervisorHelper.supervision_tree(Teiserver.Bridge.DiscordSystem)
```
[
  [
    id: :undefined,
    pid: #PID<0.1867.0>,
    modules: [Teiserver.Bridge.DiscordSupervisor],
    supervised: [
      [
        id: Nostrum.Application,
        pid: #PID<0.1868.0>,
```


Hitting the ui button on the dashboard, I can see the supervision tree pids have changed again

iex> SupervisorHelper.supervision_tree(Teiserver.Bridge.DiscordSystem)
```
[
  [
    id: :undefined,
    pid: #PID<0.2794.0>,
    modules: [Teiserver.Bridge.DiscordSupervisor],
    supervised: [
      [
        id: Nostrum.Application,
        pid: #PID<0.2795.0>,
```

And that the restarted bridge still reads messages from Discord.

```
2026-03-17 21:03:47.259 [debug] pid=<0.1657.0> actor_type=chat_room actor_id=main  QUERY OK source="teiserver_room_messages" db=6.1ms queue=0.3ms idle=1529.2ms
INSERT INTO "teiserver_room_messages" ("user_id","inserted_at","chat_room","content") VALUES ($1,$2,$3,$4) RETURNING "id" [40, ~U[2026-03-18 02:03:47Z], "main", "nortyspock: double test"]

```

I'll look into returning the flash message, that was also in the back of my head but I had not gotten to it yet.

lib/teiserver_web/live/admin_dashboard/index.ex
NortySpock: Instead of making this a basic async thing, I ended up sending the "success" message back via `Process.send(notify_pid, :discord_bridge_restarted, [])` ... That way any page navigation or process failure of Phoenix shouldn't interrupt the Discord Bridge restart process.

Hopefully that's acceptable.

lib/teiserver/bridge/discord_system.ex
```diff
@@ -24,4 +24,44 @@ defmodule Teiserver.Bridge.DiscordSystem do
...
+
+  @spec perform_restart(pid()) :: :ok
+  defp perform_restart(notify_pid) do
+    _termination_result =
+      DynamicSupervisor.terminate_child(
+        __MODULE__,
+        Process.whereis(Teiserver.Bridge.DiscordSupervisor)
+      )
+
+    spec = Supervisor.child_spec(Teiserver.Bridge.DiscordSupervisor, restart: :temporary)
```
geekingfrog: This is the same as what's done in `init`, and it's something that can easily be messed up. You should have a private function that starts the child, called both from `init` and `restart`.

lib/teiserver/bridge/discord_system.ex
```diff
@@ -13,15 +13,34 @@ defmodule Teiserver.Bridge.DiscordSystem do
...
+      end
+    end)
+  end
+
+  @spec restart :: Task.t()
+  def restart do
+    if Process.whereis(Teiserver.Bridge.DiscordSupervisor) do
+      DynamicSupervisor.terminate_child(
+        __MODULE__,
+        Process.whereis(Teiserver.Bridge.DiscordSupervisor)
```
NortySpock: I tried "just let it crash" if the child pid did not exist, but the crash bubbled up to kill the GenServer that called it. So hence the guard check looking for existence of the supervisor.

lib/teiserver/bridge/discord_system.ex
```diff
@@ -13,15 +13,34 @@ defmodule Teiserver.Bridge.DiscordSystem do
...
-      Task.async(fn ->
+    start()
+
+    {:ok, sup_flags}
+  end
+
+  @spec start :: Task.t()
+  def start do
+    Task.async(fn ->
+      if Communication.use_discord?() do
```
NortySpock, line 24: Moved the `if Communication.use_discord?()` check inside `Task.async` so as to always be returning a task.

lib/teiserver/bridge/discord_system.ex
geekingfrog: This is racy, the process may disappear between the first and the second check.

```elixir
case Process.whereis(...) do
  x when is_pid(x) -> ...
  _ -> ...
end
```

lib/teiserver/bridge/discord_system.ex
```diff
@@ -13,15 +13,34 @@ defmodule Teiserver.Bridge.DiscordSystem do
...
-    if Communication.use_discord?() do
-      Task.async(fn ->
+    start()
+
+    {:ok, sup_flags}
+  end
+
+  @spec start :: Task.t()
+  def start do
+    Task.async(fn ->
```
geekingfrog: This **still** creates a task outside of all supervision tree. Read [the doc](https://hexdocs.pm/elixir/Task.html#module-dynamically-supervised-tasks) and amend please.

NortySpock: If I attach it to DiscordSystem, and this restart task fails, wouldn't that risk the restart task causing the crash to bubble up through the entire application? 

Edit: To which supervisor should I attach this task?  

Edit2: Because I thought the whole point for this particular module was to allow the process of starting the supervisor to possibly crash without taking down everything else.

lib/teiserver_web/live/admin_dashboard/index.ex
```diff
@@ -114,6 +115,25 @@ defmodule TeiserverWeb.AdminDashLive.Index do
...
   end
 
+  @spec handle_event(String.t(), map(), Socket.t()) :: {:noreply, Socket.t()}
+  def handle_event("restart-discord-bridge", _event, socket) do
+    restart_status = DiscordSystem.restart() |> Task.await()
+
+    {flash_type, message} =
+      case restart_status do
+        :disabled_by_configuration ->
+          {:error, "Discord Bridge is disabled in your configuration."}
```
L-e-x-o-n: You could just not show the restart button if Discord is disabled in config.

NortySpock: Can do on the UI change.

I was going to keep the return atom (perhaps I should do {:error, :disabled_by_configuration}  ) so as to have a convenient API in case this gets expanded to "periodic bridge restart"

The previous button placement was just "Restart Policies".

My current version is "Restart Discord Bridge, Restart Policies".

My minor concern is that if we just hide the button entirely in the current version, the "Restart Policies" button will move back to the left, causing different button positioning depending on if Discord is enabled. This could cause occasional mis-clicks. 

I'll swap the order to "Restart Policies, Restart Discord Bridge" so that the purple button disappearing won't change the positioning.

NortySpock:
Note, the following test flaked out on me, it worked on my machine and in previous runs...


```
  1) test leaving parties can create party after leaving (Teiserver.Protocols.Spring.SpringPartyTest)
Error:      test/teiserver/protocols/spring/spring_party_test.exs:194
     match (=) failed
     code:  assert {"OK", _args, _msg_id} = _recv_until(socket) |> parse_in_message()
     left:  {"OK", _args, _msg_id}
     right: nil
     stacktrace:
       test/teiserver/protocols/spring/spring_party_test.exs:391: Teiserver.Protocols.Spring.SpringPartyTest.leave_party!/1
       test/teiserver/protocols/spring/spring_party_test.exs:195: (test)
```

test/teiserver/protocols/spring/spring_party_test.exs
```diff
@@ -191,6 +191,7 @@ defmodule Teiserver.Protocols.Spring.SpringPartyTest do
       assert username == user2.name
     end
 
+    @tag :needs_attention
```
geekingfrog, line 195: could you add a link to the failure in CI please?

NortySpock: Sure. Done.

PR #1025, opened 2026-03-17
Added credo check for cond statements

Teifion:
@L-e-x-o-n as promised, here is a present for you.

It should add in a credo check and clear all the issues with it. In each case the output will be the same but there were a few instances of logic being able to be simplified and I took this as an opportunity to improve it.

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,2 +1,0 @@
-elixir 1.19.4-otp-26
-erlang 26.2.5.1


.tool-versions
```diff
@@ -0,0 +1,2 @@
+elixir 1.19.4-otp-26
```
Teifion: Adding this file allows for `asdf install` to be used. I appreciate it is not within the wider scope of the PR and would be happy to remove it if desired.

.credo.exs
```diff
@@ -2,31 +2,30 @@
...
-        # TODO: Enable this check by deleting the line below, then fix the issues
-        {Credo.Check.Readability.AliasOrder, false},
-        # TODO: Enable this check by deleting the line below, then fix the issues
-        {Credo.Check.Readability.ModuleDoc, false},
-        # TODO: Enable this check by deleting the line below, then fix the issues
-        {Credo.Check.Readability.UnnecessaryAliasExpansion, false},
-        # TODO: Enable this check by deleting the line below, then fix the issues
-        {Credo.Check.Readability.PredicateFunctionNames, false}
-      ]
+      checks: %{
```
Teifion, line 5: New credo.exs files use this format and it makes it easier to keep track of different checks.

lib/teiserver/account/libs/accolade_lib.ex
```diff
@@ -246,7 +244,7 @@ defmodule Teiserver.Account.AccoladeLib do
     end
   end
 
-  @spec get_accolade_bot_userid() :: T.userid()
+  @spec get_accolade_bot_userid() :: T.userid() | nil
```
Teifion, line 247: Without this we get a dialyzer error when calling it and catching a nil value.

lib/teiserver/battle/servers/match_monitor_server.ex
```diff
@@ -28,7 +28,7 @@ defmodule Teiserver.Battle.MatchMonitorServer do
     GenServer.start_link(__MODULE__, opts[:data], [])
   end
 
-  @spec get_match_monitor_userid() :: T.userid()
+  @spec get_match_monitor_userid() :: T.userid() | nil
```
Teifion, line 31: Added because without it when trying to catch a nil value we get a Dialyzer error.

.tool-versions
L-e-x-o-n: `.tool-versions` has been removed in favour of [Mise](https://mise.jdx.dev/) and [mise.toml](https://github.com/beyond-all-reason/teiserver/blob/master/mise.toml).

PR #1026, opened 2026-03-17
Fix multi team rating updates

### LINKED ISSUE #434: [Bug]: Players of equal skill can gain infinite FFA rating just playing each other
### Describe the Bug

Three players playing FFA by themselves can gain rating seemingly out of nowhere. Initially these players should start with 25 skill each or 75 skill among them. If they were to play each other for many games, we would expect the total skill to be roughly 75.

We can find it is not. 

### Reproduce the bug

https://bar-rts.com/replays/b517af66db3fb90f3858f90d34bb06cd
TheAnnihilator 36
nanobot 10
iqwert1717 41

These three players only play each other and nobody else, and yet their total rating is too high. We can find TheAnnihilator's FFA games via: 
https://www.beyondallreason.info/replays?page=1&limit=24&preset=ffa&hasBots=false&endedNormally=true&players=TheAnnihilator

He only plays his friends.

Here is a game on July:
https://www.beyondallreason.info/replays?gameId=8ec88866bfc5ab84c52f8643294bb53a
where his rating is 0 and the other ratings are 25, 0.

Since then he has only played with his friends and now they all have higher ratings somehow. The ratings are coming from thin air.

### Screenshots

_No response_

### Additional context

See match_rating_lib
```

    # Build ratings into lists of tuples for the OpenSkill module to handle
    winner_ratings =
      winners
      |> Enum.map(fn membership ->
        rating = rating_lookup[membership.user_id] || BalanceLib.default_rating(rating_type_id)
        {membership.user_id, {rating.skill, rating.uncertainty}}
      end)

    # Now we want to get the best loser to use for the winner's win
    loser_ratings =
      losers
      |> Enum.group_by(
        fn %{team_id: team_id} -> team_id end,
        fn %{user_id: user_id} ->
          rating = rating_lookup[user_id] || BalanceLib.default_rating(rating_type_id)
          {user_id, {rating.skill, rating.uncertainty}}
        end
      )
      |> Map.values()


    # Run the winner calculation
    [win_result | _lose_result] = rate_with_ids([winner_ratings | loser_ratings])
    win_result = Map.new(win_result)
```

It seems here that the winner's rating is calculated by assuming it was a 1 v 2. I.e. we have the winner's ratings versus all the losers' ratings. Teifion mentions in code "Now we want to get the best loser to use for the winner's win" , but it is not the "best" loser - it is all the losers.

```
  # If you lose you just count as losing against the winner
    loss_ratings =
      loser_ratings
      |> Enum.map(fn team_ratings ->
        lose_results = rate_with_ids([winner_ratings, team_ratings], as_map: true)

        team_ratings
        |> Enum.map(fn {user_id, _old_rating} ->
          rating_update = lose_results[user_id]

          user_rating = rating_lookup[user_id] || BalanceLib.default_rating(rating_type_id)
          ratiod_rating_update = apply_change_ratio(user_rating, rating_update, opponent_ratio)
          do_update_rating(user_id, match, user_rating, ratiod_rating_update)
        end)
      end)
      |> List.flatten()
```

However, the loser ratings are calculated as a 1v1 ie. assume the loser lost against the winner. 

So if we have three friends playing A,B,C and A wins then

Pretend A vs B+C and calculate A new rating assuming win
Pretend B vs A and calculate B new rating assuming loss.
Pretend C vs A and calculate C new rating assuming loss.

@geekingfrog is my interpretation of the code correct?

### LINKED ISSUE #656: [Bug]: Asymmetric multi-team OpenSkill updates break μ–σ consistency
### Describe the Bug

In multi-team games (team_count > 2), winners are updated in one big multi-group call while losers are handled by repeated pairwise calls against the winners and then have only their μ (skill) change scaled by an opponent_ratio, leaving σ (uncertainty) untouched. This breaks the core OpenSkill assumption that confidence (σ) should track any damping of skill updates, leading to over-rewarded winners, under-penalized losers, and gradual leaderboard inflation.

A minimal fix is to replace the entire multi-team logic with a single `rate_with_ids([winner_ratings | loser_ratings], as_map: true)` call for all participants, removing the opponent_ratio hack entirely. If any scaling is truly necessary, it must be applied equally to both μ and σ (e.g. {Δµ * r, Δσ * r}) so that uncertainty remains consistent with skill changes, and comments must be updated to accurately reflect the chosen approach.

https://github.com/beyond-all-reason/teiserver/blob/master/lib/teiserver/game/libs/match_rating_lib.ex

### Reproduce the bug

See above. This is based on review of source code.

L-e-x-o-n:
Currently rating matches with more than 2 teams was handled with some custom logic - the winner's rating was updated as if they won against the best loser team and losing teams' ratings were updated as if they lost against the winner.

The openskill system and library supports rating multiple teams, this PR replaces the old custom logic and lets openskill properly rate all the teams.

A noteworthy difference from previous version is that now a team might gain skill if their placement is above expectations without necessarily having to win, for example a second team might still gain some rating.


L-e-x-o-n:
Tested by recalculating FFA ratings on integration server.
This change will probably be followed by partial or full FFA rating recalculation and reset, but that can come as a different PR:

geekingfrog (approved):

PR #1027, opened 2026-03-18
Add batch of credo rules - alias commands

Teifion:
Adds a selection of credo checks, some already pass but a variety of module import ones do not and this PR is to fix that.

There should be no behavioural changes to any of the code.

Claude code was used to write a script to assist and to perform a large portion of the changes. The PR will remain in draft until I have personally reviewed every file changed.

### DIFF
--- a/.credo.exs
+++ b/.credo.exs
@@ -2,6 +2,57 @@
   configs: [
     %{
       name: "default",
+      #
+      # These are the files included in the analysis:
+      files: %{
+        #
+        # You can give explicit globs or simply directories.
+        # In the latter case `**/*.{ex,exs}` will be used.
+        #
+        included: [
+          "credo/",
+          "lib/",
+          "priv/",
+          "scripts/",
+          "test/"
+        ],
+        excluded: [
+          ~r"/_build/",
+          ~r"/deps/",
+          ~r"/node_modules/"
+        ]
+      },
+      #
+      # Load and configure plugins here:
+      #
+      plugins: [],
+      #
+      # If you create your own checks, you must specify the source files for
+      # them here, so they can be loaded by Credo before running the analysis.
+      #
+      requires: [],
+      #
+      # If you want to enforce a style guide and need a more traditional linting
+      # experience, you can change `strict` to `true` below:
+      #
+      strict: true,
+      #
+      # To modify the timeout for parsing files, change this value:
+      #
+      parse_timeout: 5000,
+      #
+      # If you want to use uncolored output by default, you can change `color`
+      # to `false` below:
+      #
+      color: true,
+      #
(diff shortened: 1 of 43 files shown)


lib/teiserver/account.ex
```diff
@@ -1,13 +1,47 @@
 defmodule Teiserver.Account do
   @moduledoc false
   import Ecto.Query, warn: false
-  alias Teiserver.Repo
+
```
Teifion, line 4: This is a massive context file and one with a lot of imports which is why this bit looks so weird. It moves all the aliases to the top instead of scattered throughout the file.

mix.lock
```diff
@@ -17,7 +17,7 @@
...
   "cpu_util": {:hex, :cpu_util, "0.6.0", "dcf064ce2c40d272e2747adb06b311292f2e8627f2b62816a11e09f53f273bd8", [:mix], [], "hexpm", "160cc78325827f82411e21ad6d80de8d8e3349a74492d379e8bbea88b3624246"},
-  "credo": {:hex, :credo, "1.7.12", "9e3c20463de4b5f3f23721527fcaf16722ec815e70ff6c60b86412c695d426c1", [:mix], [{:bunt, "~> 0.2.1 or ~> 1.0", [hex: :bunt, repo: "hexpm", optional: false]}, {:file_system, "~> 0.2 or ~> 1.0", [hex: :file_system, repo: "hexpm", optional: false]}, {:jason, "~> 1.0", [hex: :jason, repo: "hexpm", optional: false]}], "hexpm", "8493d45c656c5427d9c729235b99d498bd133421f3e0a683e5c1b561471291e5"},
+  "credo": {:hex, :credo, "1.7.17", "f92b6aa5b26301eaa5a35e4d48ebf5aa1e7094ac00ae38f87086c562caf8a22f", [:mix], [{:bunt, "~> 0.2.1 or ~> 1.0", [hex: :bunt, repo: "hexpm", optional: false]}, {:file_system, "~> 0.2 or ~> 1.0", [hex: :file_system, repo: "hexpm", optional: false]}, {:jason, "~> 1.0", [hex: :jason, repo: "hexpm", optional: false]}], "hexpm", "1eb5645c835f0b6c9b5410f94b5a185057bcf6d62a9c2b476da971cde8749645"},
```
Teifion, line 20: Credo version bump as part of bringing all the credo checks up to scratch.

.credo.exs
```diff
@@ -2,8 +2,69 @@
   configs: [
```
Teifion, line 2: This brings the credo.exs inline (structure-wise) with a new credo.exs file. It also defaults it to strict so as to be inline with the CI process.

PR #1036, opened 2026-03-19
Added credo map-map check

Teifion:
Adds the "map-map" check to credo.

Also treats warnings as errors in the github CI

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -75,7 +75,7 @@ jobs:
       run: mix format --check-formatted
 
     - name: Compile the project
-      run: mix compile
+      run: mix compile --warnings-as-errors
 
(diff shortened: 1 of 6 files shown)


.github/workflows/lint.yml
```diff
@@ -75,7 +75,7 @@ jobs:
       run: mix format --check-formatted
 
     - name: Compile the project
-      run: mix compile
+      run: mix compile --warnings-as-errors
```
geekingfrog, line 78: Why is this required? we already have `elixirc_options: [warnings_as_errors: true],` in mix.exs, is that not working on the CI?

Teifion: I couldn't find anything in the github files so thought it wasn't being checked; looks like I was mistaken :(

PR #1038, opened 2026-03-20
Added StrictModuleLayout credo check

Teifion:
Added StrictModuleLayout credo check

Claude code was used to perform the changes though I will manually verify them and post a comment to that effect afterwards.

I will also submit a PR subsequent to this to blame ignore the changes here.


lib/teiserver/mix_tasks/ensure_test_users.ex
```diff
@@ -1,15 +1,13 @@
...
-  The token must have `admin.user` permission.
-  "
-
   @moduledoc """
   Generate `n` random users based on predictible names.
   The goal is to end up with `n` users and access token ready to be used
   for load testing.
 
-  #{@usage_str}
+  Usage: `mix teiserver.ensure_test_users --host <host> <tok> <n>`
```
Teifion, line 7: Not like the others but the module attribute isn't used elsewhere so fits.

PR #1043, opened 2026-03-20
Prevent changing map when another vote is in progress

geekingfrog:
The error message is currently a string and we can't translate that. For now it'll do but we likely want to fix that for the release.

### DIFF
--- a/lib/teiserver/tachyon_lobby/lobby.ex
+++ b/lib/teiserver/tachyon_lobby/lobby.ex
@@ -1764,6 +1764,13 @@ defmodule Teiserver.TachyonLobby.Lobby do
       not is_map_key(state.players, user_id) ->
         {:error, "Only players can change the map"}
 
+      state.current_vote ->
+        case state.current_vote.action do
+          # make changing map idempotent, it's just a nicer API this way
+          {:change_map, ^new_name} -> {:ok, []}
+          _ -> {:error, :vote_in_progress}
+        end
+
       Enum.count(state.players, fn {_, p} -> not bot_id?(p.id) end) > 1 ->
         vote_duration_s = 60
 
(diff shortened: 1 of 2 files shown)


lib/teiserver/tachyon_lobby/lobby.ex
```diff
@@ -1764,6 +1764,13 @@ defmodule Teiserver.TachyonLobby.Lobby do
       not is_map_key(state.players, user_id) ->
         {:error, "Only players can change the map"}
 
+      state.current_vote != nil ->
```
Teifion: Could use an implicit not nil there, same with some of the tests.
```suggestion
      state.current_vote ->
```

test/teiserver/tachyon_lobby/lobby_test.exs
```diff
@@ -1052,6 +1063,17 @@ defmodule Teiserver.TachyonLobby.LobbyTest do
...
+
+    test "cannot change map when vote already pending" do
+      %{id: id} = setup_full_lobby([1, 1])
+      :ok = Lobby.join_queue(id, "2")
+      :ok = Lobby.update_properties(id, @default_user_id, %{map_name: "new map"})
+      assert_receive {:lobby, ^id, {:updated, %{players: %{}}}}
+      assert_receive {:lobby, ^id, {:updated, %{current_vote: vote}}}
+      assert vote != nil
+
+      {:error, _} = Lobby.update_properties(id, @default_user_id, %{map_name: "different map"})
```
Teifion, line 1075: Why not match against the expected error atom?

Teifion (approved):
For what it's worth I approve :)

test/teiserver/tachyon_lobby/lobby_test.exs
geekingfrog: Because it doesn't actually [return an atom](https://github.com/geekingfrog/teiserver/blob/d71261901db99e6fa95090328b91c5d8821fc760/lib/teiserver/tachyon_lobby/lobby.ex#L802).
I suspect we are going to want to rework that at some point later, if only because these errors can't be translated by the client. But for now that'll do.

PR #1046, opened 2026-03-21
Centralise the logic to call a lobby

geekingfrog:
Because appart from one case, it's always the same handling with the same errors.


Teifion (approved):
LGTM

PR #1053, opened 2026-03-21
Removes the AccoladeBotServer and AccoladeChatBot server

Teifion:
Removes the AccoladeBotServer and AccoladeChatBot server

### DIFF
--- a/lib/teiserver/libs/teiserver_configs.ex
+++ b/lib/teiserver/libs/teiserver_configs.ex
@@ -76,15 +76,6 @@ defmodule Teiserver.TeiserverConfigs do
     })
 
     add_site_config_type(%{
-      key: "teiserver.Inform of new accolades",
-      section: "Accolades",
-      type: "boolean",
-      permissions: ["Server"],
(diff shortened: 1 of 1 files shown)


lib/teiserver/account/accolades/accolade_chat_server.ex
```diff
@@ -1,229 +0,0 @@
...
-            [{_index, selected_type}] ->
-              Account.create_accolade(%{
-                recipient_id: state.recipient_id,
-                giver_id: state.userid,
-                badge_type_id: selected_type.id,
-                match_id: state.match_id,
-                inserted_at: Timex.now()
-              })
-
-              if Config.get_site_config_cache("teiserver.Inform of new accolades") do
```
L-e-x-o-n, line 114: This config is only used here and can now be deleted from `Teiserver.TeiserverConfigs`.

Teifion: Addressed in https://github.com/beyond-all-reason/teiserver/pull/1053/commits/efd0a80054a54ce31329521e4fd8e6156e52832d

PR #1054, opened 2026-03-21
Re-added some tests, started work on removing Timex

Teifion:


lib/teiserver/data/cache_user.ex
```diff
@@ -964,20 +964,23 @@ defmodule Teiserver.CacheUser do
...
-      bot: Auth.is_bot?(user),
-      country: country,
-      rank: rank,
-      lobby_client: lobby_client,
-      lobby_hash: lobby_hash,
-      last_ip: ip
-    })
+    # These steps are not needed for most tests so we can skip them as they
+    # can cause flakiness
+    if not Application.get_env(:teiserver, Teiserver)[:test_mode] do
```
geekingfrog, line 978: Is this `update_user_stat` a functionality we are planning to keep with tachyon?
If yes I would be keen on finding a different way to fix the flakyness.
More conditional makes the code harder to read. This also introduces the potential for misconfiguration, where important code paths would not run.
And of course, it makes this block untestable, which isn't great.

I don't mind as a stopgap measure, but I would not like to see this pattern spread over the code.

geekingfrog (commented):
I would also remove code instead of just commenting it. It's rather noisy and not very helpful. If you want to recover previous version, git is there for that.

lib/teiserver/data/cache_user.ex
Teifion: It's not something I plan to keep in the PR at all; I completely agree with the concerns you have but want to try different things to identify issues.

Teifion: Related to the earlier comment, the user start part is outside the conditional but the telemetry event and smurf key are not (they are the bit that seems to trigger issues sometimes)

lib/teiserver_web/templates/admin/asset/index.html.heex
L-e-x-o-n: Why not delete this entierly?

Teifion: Missed by mistake, addressed in https://github.com/beyond-all-reason/teiserver/pull/1054/commits/d189e8f0309b722514726163d64b441329c8d100

test/teiserver_web/live/client_live_test.exs
```diff
@@ -1,23 +1,25 @@
...
-  import TeiserverTestLib, only: [_send_raw: 2]
+  import TeiserverTestLib, only: [_send_raw: 2, teiserver_seed: 0]
 
   setup do
     GeneralTestLib.conn_setup(TeiserverTestLib.admin_permissions())
     |> TeiserverTestLib.conn_setup()
   end
 
-  @sleep_time 2100
+  @sleep_time ClientIndexThrottle.update_interval() + 100
```
geekingfrog: maybe out of scope for this PR since the tests are still disabled, but waiting for some time isn't great.
What I tend to do instead is to expose a function like `trigger_update` or something, with a doc mentioning this is only for tests.
This function then sends the message (`:tick` in this case).
And then the test can use some of the [polling functions](https://github.com/geekingfrog/teiserver/blob/ddf768c8152f802c053499ac193a631303bd5e8c/test/support/polling.ex) to wait until the update happened.

geekingfrog (approved):
Also as a meta comment, the commits aren't great. "fixes `n`" isn't terribly helpful.

A better split would be something like:
- disable warning as error + explaining why in the commit message body
- remove/replace some unused timex helpers
- fix permissions on xxx liveviews & associated tests
- better test isolation with cache clear + you can link to a CI flake on github, or explain why these cache were making some tests fails/noisy

and so on.
The commit messages are quite helpful to explain the **why** of a change. The removal of `[warnings_as_errors: true]` is a great example of that. It helps when digging through the history of the project to understand why a change was made.

Having these type of commit also helps a lot when reviewing.

I think reworking the commits there may be quite painful though, so I'm cool merging this, but squash please though, since the commit history is useless, let's not add these to the general history. Something to keep in mind for future PRs.

test/teiserver_web/live/client_live_test.exs
Teifion: It's part of a larger effort to improve tests so I'd rather leave it in since it improves things. I would love to have a chat on the discord over the way we want to approach things like this though.

geekingfrog: As an example, you can look at the [login rate limiter](https://github.com/geekingfrog/teiserver/blob/ddf768c8152f802c053499ac193a631303bd5e8c/lib/teiserver/account/servers/login_throttle_server.ex#L99-L101) and [its test](https://github.com/geekingfrog/teiserver/blob/ddf768c8152f802c053499ac193a631303bd5e8c/test/teiserver/account/login_throttle_server_test.exs) which works reasonably well.

Teifion: That's a better idea, I'll rejig the test to use that even if it has to stay needs attention for now.

lib/teiserver/helpers/timex_helper.ex
```diff
@@ -224,33 +187,14 @@ defmodule Teiserver.Helper.TimexHelper do
   @spec _hms_or_dmy(DateTime.t(), DateTime.t()) :: String.t()
   defp _hms_or_dmy(the_time, today) do
     if the_time |> Timex.to_date() |> Timex.compare(today) == 0 do
-      Timex.format!(the_time, "Today at {h24}:{m}:{s}")
+      # Timex.format!(the_time, "Today at {h24}:{m}:{s}")
+      Calendar.strftime(the_time, "Today at %I:%M:%S")
     else
-      Timex.format!(the_time, "{0D}/{0M}/{YYYY}")
+      # Timex.format!(the_time, "{0D}/{0M}/{YYYY}")
+      Calendar.strftime(the_time, "%d/%m/%Y")
```
L-e-x-o-n: Remove the commented Timex code.

Teifion: Addressed, thanks

PR #1058, opened 2026-03-24
Removes Report Groups

ChocoboCoding:
Removes Report Groups, Report Group Messages and Report Group Votes


priv/repo/migrations/20260324180447_remove_report_groups.exs
```diff
@@ -0,0 +1,21 @@
...
+      remove :report_group_id
+    end
+
+    drop table(:moderation_report_groups), mode: :cascade
+    drop table(:moderation_report_group_votes)
+    drop table(:moderation_report_group_messages)
+  end
+
+  def down do
+    raise "Irreversible migration: cannot un-delete moderation_report_groups"
```
L-e-x-o-n: Not completely irreversible, you can create the tables on rollback even if they would be empty.
Try only [change](https://hexdocs.pm/ecto_sql/Ecto.Migration.html#module-change) instead.

lib/teiserver_web/components/pagination_components.ex
```diff
@@ -31,20 +31,6 @@ defmodule TeiserverWeb.PaginationComponents do
         class="mt-3"
       />
 
-      # Full pagination with search parameter preservation (LiveView)
```
Teifion, line 34: Why are you removing the comments in this component, it has nothing to do with report groups?

test/teiserver_web/live/moderation/overwatch/index_test.exs
L-e-x-o-n, line 1: It looks like only one test here is for report groups, why delete everything?

ChocoboCoding: The /moderation/overwatch page is the :index page for report groups. Since I am removing that page, there is no point keeping the tests around which test for that page

L-e-x-o-n: I see, however it might still be worth reusing the test for testing the Report page then. If you want to, not necessary.

lib/teiserver_web/components/pagination_components.ex
ChocoboCoding: It is a comment for the doc example for a LiveView (as I just realized). I'll restore it

test/teiserver_web/live/moderation/overwatch/index_test.exs
ChocoboCoding: Report page should be tested in `test/teiserver_web/controllers/moderation/report_controller_test.exs` already

lib/teiserver_web/components/pagination_components.ex
ChocoboCoding: Adressed in https://github.com/beyond-all-reason/teiserver/pull/1058/commits/f5fab1560a1f98cb74af477ea6cee59873026964

priv/repo/migrations/20260324180447_remove_report_groups.exs
ChocoboCoding: Adressed in https://github.com/beyond-all-reason/teiserver/pull/1058/commits/f5fab1560a1f98cb74af477ea6cee59873026964

L-e-x-o-n (approved):

PR #1062, opened 2026-03-26
Use user.roles everywhere instead of user.data.roles

L-e-x-o-n:
Another part of an ongoing effort to eventually remove user.data.

### DIFF
--- a/lib/teiserver/protocols/spring/spring_in.ex
+++ b/lib/teiserver/protocols/spring/spring_in.ex
@@ -408,7 +408,6 @@ defmodule Teiserver.Protocols.SpringIn do
         :ok
 
       {:error, reason} ->
-        Coordinator.send_to_user(state.userid, reason)
         reply(:servermsg, reason, msg_id, state)
     end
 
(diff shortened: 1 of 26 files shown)


lib/teiserver/protocols/spring/spring_in.ex
```diff
@@ -408,7 +408,6 @@ defmodule Teiserver.Protocols.SpringIn do
         :ok
 
       {:error, reason} ->
-        Coordinator.send_to_user(state.userid, reason)
```
L-e-x-o-n, line 411: Was causing a foreign key error when running all tests but not when running this test individually.
Didn't seem to work anyway (I didn't receive message from coordinator on invalid name change).

```
* (Ecto.ConstraintError) constraint error when attempting to insert struct:

    * "direct_messages_from_id_fkey" (foreign_key_constraint)

If you would like to stop this constraint violation from raising an
exception and instead add it as an error to your changeset, please
call `foreign_key_constraint/3` on your changeset with the constraint
`:name` as an option.
```

PR #1067, opened 2026-03-31
Added post command

ChocoboCoding:
This adds a command to post reports, actions and the moderation profile link to the current channel

### DIFF
--- a/lib/teiserver/bridge/commands/post_command.ex
+++ b/lib/teiserver/bridge/commands/post_command.ex
@@ -18,26 +18,50 @@ defmodule Teiserver.Bridge.Commands.PostCommand do
   def cmd_definition do
     %{
       name: "post",
-      description: "Post report or action into the current channel",
+      description: "Post something into the current channel",
       options: [
         %{
-          # type3 = String
-          type: 3,
-          name: "type",
-          description: "Report or Action",
+          # type  = sub_command
+          type: 1,
+          name: "report",
+          description: "Post a report",
           required: true,
-          choices: [
-            %{name: "Report", value: "report"},
-            %{name: "Action", value: "action"},
-            %{name: "Profile", value: "profile"}
+          options: [
+            %{
+              name: "id",
+              description: "ID of the report",
+              type: 3,
+              required: true
+            }
           ]
         },
         %{
-          # type3 = String
-          type: 3,
-          name: "args",
-          description: "Additional arguments",
-          required: true
+          type: 1,
+          name: "action",
+          description: "Post an Action",
+          required: true,
+          options: [
+            %{
+              name: "id",
+              description: "ID of the action",
+              type: 3,
+              required: true
+            }
+          ]
+        },
+        %{
+          type: 1,
+          name: "profile",
+          description: "Post the link of a moderation profile",
+          required: true,
+          options: [
+            %{
+              name: "name",
+              description: "Name of the Account",
+              type: 3,
+              required: true
+            }
+          ]
         }
       ],
       nsfw: false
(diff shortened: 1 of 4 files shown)


lib/teiserver/bridge/commands/post_command.ex
```diff
@@ -0,0 +1,94 @@
...
+
+  @impl Teiserver.Bridge.BridgeCommandBehaviour
+  @spec name() :: String.t()
+  def name, do: "post"
+
+  @impl Teiserver.Bridge.BridgeCommandBehaviour
+  @spec cmd_definition() :: map()
+  def cmd_definition do
+    %{
+      name: "post",
```
L-e-x-o-n: Call it something a  bit more specific. "moderation" or even just "user" would be clearer.

lib/teiserver/bridge/commands/post_command.ex
```diff
@@ -0,0 +1,94 @@
...
+            %{name: "Profile", value: "profile"}
+          ]
+        },
+        %{
+          # type3 = String
+          type: 3,
+          name: "args",
+          description: "Additional arguments",
+          required: true
+        }
```
L-e-x-o-n, line 68: I don't like this, the person using the command must somehow know whether the command expects user ID or name and providing the wrong one will result in an exception.

Look into [subcommands and subcommand groups](https://docs.discord.com/developers/interactions/application-commands#subcommands-and-subcommand-groups), it should be possible to structure the command in such a way to offer you a choice of either inputting user name or user ID and then each can be handled properly.

lib/teiserver/bridge/commands/post_command.ex
```diff
@@ -0,0 +1,94 @@
...
+          "profile" ->
+            case Account.get_user_by_name(args) do
+              nil ->
+                "User \"#{args}\" not found"
+
+              user ->
+                "https://#{Application.get_env(:teiserver, TeiserverWeb.Endpoint)[:url][:host]}/moderation/report/user/#{user.id}"
+            end
+        end
+      end
```
L-e-x-o-n: Getting action/report will throw an exception if the args is of the wrong type and/or an action/report doesn't exist.

lib/teiserver/bridge/discord_bridge_bot.ex
```diff
@@ -300,29 +337,57 @@ defmodule Teiserver.Bridge.DiscordBridgeBot do
...
+
+        report.closed == true ->
+          ["**Status:** Closed :file_folder:"]
+
+        true ->
+          ["**Status:** Open"]
+      end
+  end
+
+  def get_channel(type) do
```
L-e-x-o-n: Maybe rename to `get_channel_for_report_type`.

lib/teiserver/bridge/commands/post_command.ex
```diff
@@ -0,0 +1,94 @@
...
+          "report" ->
+            report = Moderation.get_report!(args, preload: [:reporter, :target])
+
+            DiscordBridgeBot.get_report_message(report)
+            |> List.delete_at(-1)
+            |> List.insert_at(
+              -1,
+              "**Original:** https://discord.com/channels/#{Communication.get_guild_id()}/#{DiscordBridgeBot.get_channel(report.type)}/#{report.discord_message_id}"
+            )
+            |> Enum.join("\n")
```
L-e-x-o-n: What if this is posted in a non private channel? What is the purpose here exactly?
If it's to (re)post a message for a missing report then `DiscordBridgeBot.new_report` would be better.

L-e-x-o-n (changes requested):

lib/teiserver/bridge/commands/post_command.ex
ChocoboCoding: The idea is to post existing reports into other channels. That would for example be used to post chat reports in threads for action reports

lib/teiserver/bridge/commands/post_command.ex
ChocoboCoding: Adressed in https://github.com/beyond-all-reason/teiserver/pull/1067/commits/1e74de126af15abd1b88bd6fc5648bf211a506d7

lib/teiserver/bridge/commands/post_command.ex
ChocoboCoding: Adressed in https://github.com/beyond-all-reason/teiserver/pull/1067/commits/ce41dab7184d6f08df2d76b977539abd6061d159

lib/teiserver/bridge/discord_bridge_bot.ex
ChocoboCoding: Adressed in https://github.com/beyond-all-reason/teiserver/pull/1067/commits/2679d3b3f4a9b777bf7b53fac721b322fb8755bd

L-e-x-o-n (approved):

PR #1069, opened 2026-03-31
Refactor of Discord system to better prevent cascading failure

Teifion:
A failure in Nostrum can (despite what we expect) often cause cascading failures in other parts of the server. We have tried a number of approaches to fix it and been met with mixed success. What follows is not what I would consider the ideal first solution but it might be a viable one in this case.

The DiscordSystem is replaced by a GenServer of the same name which handles the management of the Nostrum processes and hopefully completely isolates failures.

ClaudeCode was used to help run through solutions and wrote some of the code but given the generic nature of most of it I am happy with that.


Teifion:
Closing as no longer required.

PR #1077, opened 2026-04-10
Added initial support for Plugins

Teifion:
Adds the first plugin hooks (sending messages) along with documentation for them.


geekingfrog (commented):
I think the idea is good, but I'm not convinced by the API right now.

I see a couple of problems:

1. You can't have multiple plugin reacting to the same "event". This may be a sensible limitation, do you have some thoughts on that? It's simple, so the main system doesn't have to deal with ordering or effectful plugins. They can't really be composable either, but again, that may be for the best.

2. The api currently doesn't allow a plugin to spawn its own stuff, either as a separate otp application, or a more contained supervision tree. If a plugin needs to have some form of cache, or a background cron, it's not possible. I think this is a major limitation.

3. I am not a fan of the fallback function if a plugin isn't present. I think the fallbacks should be implemented as a default plugin themselves and the code responsible to call into the plugins doesn't care either way, just go through the plugin(s) and call them.

4. probably the most minor point: all plugins have to be defined at startup/config time. Maybe we can add later a way of dynamically loading/unloading plugins. The beam is rather suited to this task, though if we end up running teiserver in an immutable container it's going to be awkward.


I think I would implement plugins by defining a behaviour, similar to GenServer. In the config, you specify a list of plugins.

Functions like `send_chat_message` would be callbacks on the behaviour, with a default no-op (or nothing, and the caller checks with `Kernel.function_exported?`).
The default Teiserver would have a plugin (or several, one for each context?) to supply the default behaviour.

The interface would also have something like an `init` like. Perhaps at that point you pass a plugin specific config, where one could define a plugin with a module name (implementing the `TeiserverPlugin` behaviour` alongside an arbitrary map.
If implemented, this should return something that can be added under a supervisor. Or we mimick otp stuff, and if the plugins defines a `child_spec` function, we use that to instantiate a process to be put under the supervisor.
I think you can leave this out for now, this probably requires a bit more thoughts, but I believe it's an important piece that will be required down the road.
Plugins as full otp application may also work, but I have much less experience with that, and that feels overkill.

One problem I see with my approach is how to override only specific parts of the default behaviour? Maybe we can do that by having a bunch of default plugins for different domains, and if you don't want something, you omit that one from the list? I think this needs some further thoughts, for when you want to override instead of add to a default behaviour.

Teifion:
Good feedback thank you.

1 - Multiple plugins. If someone wants multiple things to happen they just need to have multiple things in the function used. It would be easy to extend this to work with a list later if we wanted (and to be able to support single items or lists at that). For now I'd like to leave it as single items.

2 - Spawning own stuff. The solution would be to have a plugin hook for the application module to add more children. I'll add that in a future update; I want to keep this one nice and simple.

3 - Callback function. While I don't agree with having a full behaviour system I do think it would be better to replace certain functions outright instead of the way I have done it here. The result would be the same but the code would be cleaner. I will make this change.

4 - I disagree on this point. The plugins are intended for extending the existing functionality as opposed to dynamically loading/unloading things.

geekingfrog:
> 1 - Multiple plugins. If someone wants multiple things to happen they just need to have multiple things in the function used. It would be easy to extend this to work with a list later if we wanted (and to be able to support single items or lists at that). For now I'd like to leave it as single items.

I was thinking of having multiple different plugins, where one doesn't control all the source code. For example, one may imagine bridges as plugins, one for discord, one for irc and so on.


> 2 - Spawning own stuff. The solution would be to have a plugin hook for the application module to add more children. I'll add that in a future update; I want to keep this one nice and simple.

Agreed, we can tackle that later


> 3 - Callback function. While I don't agree with having a full behaviour system I do think it would be better to replace certain functions outright instead of the way I have done it here. The result would be the same but the code would be cleaner. I will make this change.

I'm not sure I follow there tbh.
Also, I find behaviour are pretty nice to define interface. This would let us have a very clear and documented plugin api, that can also be typed. 


> 4 - I disagree on this point. The plugins are intended for extending the existing functionality as opposed to dynamically loading/unloading things.

That's also fine.

Teifion:
> I'm not sure I follow there tbh.
> Also, I find behaviour are pretty nice to define interface. This would let us have a very clear and documented plugin api, that can also be typed.

I appear to have misunderstood your point here then. How would you define the `send_chat_message` hook in your method?

geekingfrog:
So I was originally thinking of something like

```elixir
defmodule Teiserver.Plugin do
  @optional_callbacks send_chat_message: 3

  @doc """
  send a chat message to the given player
  """
  @callback send_chat_message(name :: String.t(), T.userid(), message :: String.t()) :: :ok

  # add more callbacks as required
end

defmodule Teiserver.DefaultPlugin do
  @behaviour Teiserver.Plugin

  @impl Teiserver.Plugin
  def send_chat_message(name, user_id, message), do: :ok
end

defmodule Teiserver.Plugins do
  def send_chat_message(name, user_id, message) do
    plugins = Application.get_env(:teiserver, Teiserver.Plugins, [])

    Enum.each(plugins, fn plugin_module ->
      if Kernel.function_exported?(plugin_module, :send_chat_message, 3) do
        apply(plugin_module, :send_chat_message, [name, user_id, message])
      end
    end)
  end
end
```

However, looking at it, the behaviour may be overkill since the functions exposed in `Teiserver.Plugins` serve the same role more or less. Though you don't get compiler warnings without the behaviours.

Also, there is a problem with the return value. If/when there are multiple plugins, then the return value is not well defined. Maybe that could be solved by checking there is only one plugin that define a given function when the return value is important. Or we just forbid multiple plugins altogether, but I feel that may damage modularity.

Overall I think it's pretty similar to your proposal. Though I prefer having functions instead of atoms acting as key for the function to use, it leads to more searchable and understandable codebase I feel.

Teifion:
Thank you for the feedback on the implementation. I have experimented and found the [decorator](https://hexdocs.pm/decorator/readme.html) library which I think gives us the best solution.

- Marking a function as overridable is easy and clean
- Little to no overhead for maintainers to make a function extendable
- Existing function interfaces are the same

Downside, chaining multiple implementations is potentially tricky (would need to investigate further).

L-e-x-o-n (approved):

geekingfrog (approved):

PR #1078, opened 2026-04-11
Added setting and check for max_lobby_name_size

krytgames:


krytgames (commented):
First time trying to contribute

config/runtime.exs
```diff
@@ -27,6 +27,12 @@ end
 # used for mailing, checking origins, finding tls certs…
 domain_name = Teiserver.ConfigHelpers.get_env("TEI_DOMAIN_NAME", "beyondallreason.info")
 
+max_lobby_name_size_default = 100
+max_lobby_name_size = Teiserver.ConfigHelpers.get_env("TEI_MAX_LOBBY_NAME_SIZE", max_lobby_name_size_default, :int)
+if max_lobby_name_size <= 0 do
+ max_lobby_name_size = max_lobby_name_size_default
+end
+
```
Teifion: Calculations like this should not be present in the config file.

lib/teiserver/tachyon_lobby/lobby.ex
```diff
@@ -18,6 +18,10 @@ defmodule Teiserver.TachyonLobby.Lobby do
 
   @behaviour :gen_statem
 
+  defp get_max_lobby_name_size() do
+    Application.get_env(:teiserver, Teiserver)[:max_lobby_name_size]
+  end
+
```
Teifion: This is used once and is a single line long, I think it can be inlined

lib/teiserver/tachyon_lobby/lobby.ex
```diff
@@ -1761,11 +1765,18 @@ defmodule Teiserver.TachyonLobby.Lobby do
   @spec update_property(atom(), term(), state(), T.userid()) ::
           {:ok, [event()]} | {:error, String.t()}
   defp update_property(:name, new_name, _state, _user_id) do
-    # we can expand lobby name validation later
-    if new_name == "" do
+    max_name_size = get_max_lobby_name_size()
+    name_is_empty = (new_name == "")
+    name_is_too_long = (String.length(new_name) > max_name_size)
```
Teifion, line 1830: Please talk me through why you wrote the code this way. I'd have expected to see things like `String.length` used; the presence of the parenthesis around each calculation is also very odd to see in Elixir code.

lib/teiserver/tachyon_lobby/lobby.ex
```diff
@@ -1761,11 +1765,18 @@ defmodule Teiserver.TachyonLobby.Lobby do
...
+
+    if name_is_empty do
       {:error, "name must not be empty"}
     end
 
+    if name_is_too_long do
+      {:error, "name must not be greater then #{max_name_size} characters"}
+    end
+
     {:ok, [{:update_lobby_name, new_name}]}
```
Teifion: This will result in the lobby name always being updated regardless of length or any other conditions; I suspect that is not the intent.

Teifion (commented):
I would suggest looking at each part of the code and then referring to documentation/tutorials for Elixir relevant to the specific thing you are trying to do. This code currently reads like it was Javascript converted to Elixir or written by an LLM; it will almost certainly not be doing what you think it is in several of the cases.

Additionally you have not declared the use of any LLM but your messages on discord suggest you have been using one; I would suggest updating the PR description based on what has or has not been used in that regard.

I am happy to be part of helping you learn and put the PR together.

lib/teiserver/tachyon_lobby/lobby.ex
krytgames: I just like to put boolean calculations into (), so its seems like a equality to a solid object
Also in case there will be more hard contidions u can put them in verbal manner and it will be more easy to get what is checked

lib/teiserver/tachyon_lobby/lobby.ex
krytgames: Now i get it, only the last statement automatically return and there is no "return" keyword, i'll fix, thx

lib/teiserver/tachyon_lobby/lobby.ex
krytgames: K, i'll try to use inlines more, cuz for me i'd do something like this
defp ...()
do
   some code
end
but im trying to go with the style around me

config/runtime.exs
krytgames: should i modify the get_env function to allow me for input check then?

krytgames:
also for now im using ai for only asking it questing about elixir
im not writing any code with it

Teifion:
> also for now im using ai for only asking it questing about elixir im not writing any code with it

I would suggest instead follow a tutorial such as https://elixirschool.com/ and then use the LLM for anything in that which doesn't make sense. LLMs often write other languages as if they were javascript which causes many issues.

krytgames:
o7

config/runtime.exs
krytgames: Removed

lib/teiserver/tachyon_lobby/lobby.ex
geekingfrog: Not a fan personally, especially since the vars are used right after and the check is rather simple.
I'd rather see:
```elixir
cond do
  new_name == "" -> ...
  length() > xxx -> ...
  true -> ...
end
```

geekingfrog (commented):
This also lacks tests.
You can add them in two places, either or both.

* a test only touching processes and not doing anything websocket related [over there](https://github.com/beyond-all-reason/teiserver/blob/main/test/teiserver/tachyon_lobby/lobby_test.exs). I think for such a simple change, a test in this module should be enough.

* an integration test, going through the entire websocket stuff. These live [there](https://github.com/beyond-all-reason/teiserver/blob/main/test/teiserver_web/tachyon/lobby_test.exs).

Also, CI is red

krytgames:
Hello again. Im done with the code, now i need help with spring tests, cuz tachyon tests pass
on spring tests i got this error and nobody seems to know to deal with it
["OPENBATTLEFAILED Not a bot", ""]

Also just to clarify, do i just need to run some formatting command to get rid of lint and style checks fail?

krytgames:
also about that
"Not a fan personally, especially since the vars are used right after and the check is rather simple.
I'd rather see:"
i would like to stay it this way, but if u insist, i'll move in to cond. pls answer do u allow me to do it the way i did?

geekingfrog:
> also about that "Not a fan personally, especially since the vars are used right after and the check is rather simple. I'd rather see:" i would like to stay it this way, but if u insist, i'll move in to cond. pls answer do u allow me to do it the way i did?

I don't care either way, it's minor.

> Also just to clarify, do i just need to run some formatting command to get rid of lint and style checks fail?

Look at what commands the CI runs, and do the same locally.

PR #1079, opened 2026-04-11
Vote history

geekingfrog:
Implements the [tachyon changes](https://github.com/beyond-all-reason/tachyon/pull/123) related to expose a voting history for a lobby.

### DIFF
--- a/lib/teiserver/tachyon_lobby/lobby.ex
+++ b/lib/teiserver/tachyon_lobby/lobby.ex
@@ -3,6 +3,26 @@ defmodule Teiserver.TachyonLobby.Lobby do
   Represent a single lobby
   """
 
+  # This module has essentially 3 parts:
+  # 1: the external API as usual. A set of exported function that delegate to
+  # gen_statem.cast/call and so on
+  # 2: the internal handlers: the callbacks for gen_statem
+  # 3: an event sourcing system for the handlers.
+  #
+  # The handlers are responsible to check if the operation is valid, and then
+  # translate the operation into an event. See `@typep event` for the list.
+  # These events are then reduced/folded into an aggregate. This aggregate
+  # is used to
+  # 1: compute the end state after applying the original operation/event See process_events/2
+  # 2: the update events to send to lobby members. See broadcast_updates/1
+  # 3: update the lobby list process when relevant. Also handled in broadcast_updates/1
+  #
+  # The event sourcing approach is used so we can have some "atomic" operations, and more
+  # complex operations can be represented through them. It is mostly useful when dealing
+  # with player movements, to/from spectators or within ally teams.
+  # It also handles update events without having to dispatch them manually on a case by
+  # case basis.
+
   alias Teiserver.Asset
   alias Teiserver.Autohost
   alias Teiserver.Data.Types, as: T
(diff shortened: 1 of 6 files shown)


geekingfrog:
~~Hold on reviewing this, noticed there's a problem about vote id ordering. Fixing that now~~
ok that's resolved now

lib/teiserver/tachyon_lobby/lobby.ex
```diff
@@ -1069,6 +1106,12 @@ defmodule Teiserver.TachyonLobby.Lobby do
       end)
       |> Enum.into(%{})
 
+    vote_history =
```
Teifion, line 1110: Why use a for loop instead of an Enum.map?

Teifion (commented):
First pass review of this looks promising though I need to better familiarise myself with the new Tachyon stuff overall so would suggest a different reviewer for the approval.

lib/teiserver/tachyon_lobby/lobby.ex
geekingfrog: No real reason. Though looking at the code just above it's using map+into, so I'll amend to keep the same style.

Teifion (approved):

lib/teiserver/tachyon_lobby/lobby.ex
```diff
@@ -1318,35 +1364,45 @@ defmodule Teiserver.TachyonLobby.Lobby do
...
+          Enum.map(history, fn {_id, record} -> record.finished_at end)
+          |> Enum.sort()
+
+        cutoff = Enum.at(dates, 4)
+
+        Enum.filter(history, fn {_id, record} -> record.finished_at >= cutoff end)
+        |> Enum.into(%{})
+      else
+        history
+      end
```
L-e-x-o-n, line 1401: Correct me if I am wrong, If there are 10 votes in history and then the 11th one ends this could potentially result in there only being 5 votes in history after the update. Could be seen as somewhat strange behaviour by users.

What not just always keep the last 10 votes in history? It would be slightly more computationally expensive after the 10th vote but that shouldn't be an issue in this case IMO.

lib/teiserver/tachyon_lobby/lobby.ex
```diff
@@ -1318,35 +1364,45 @@ defmodule Teiserver.TachyonLobby.Lobby do
...
   # and it allows us not to worry about storing the tref
-  defp process_event(:vote_timeout = ev, aggregate) do
-    vote_ev = {:vote_ended, aggregate.data.current_vote, :timeout}
+  defp process_event({:vote_ended, ts, outcome}, aggregate) do
+    vote_ev = {:vote_ended, aggregate.data.current_vote, outcome}
+    vote_record = %{vote: aggregate.data.current_vote, finished_at: ts, outcome: outcome}
+    history = Map.put(aggregate.data.vote_history, aggregate.data.current_vote.id, vote_record)
+
+    history =
+      if map_size(history) > 10 do
```
L-e-x-o-n: Extract to a constant please.

lib/teiserver/tachyon_lobby/lobby.ex
geekingfrog: This history is only used when new clients join. Existing client will have all the events and are free to keep the entire history if they like.
This logic is similar to how heap-allocated vectors work in many programming languages, to avoid having to reallocate for every insertion deletion.

lib/teiserver/tachyon_lobby/lobby.ex
geekingfrog: done

PR #1080, opened 2026-04-12
Fix: Terminate connections exceeding Spring protocol message buffer limit

thvl3:
Summary

  - Adds a 64KB cap on the per-connection message_part buffer in the Spring TCP protocol handler
  - Connections exceeding the limit receive s.system.disconnect Message too large and are terminated
  - Normal TCP fragmentation is unaffected

  Problem

  The Spring protocol parser in SpringIn.data_in/2 accumulates partial TCP data in state.message_part until a newline arrives. There is no upper bound on this buffer. A remote, unauthenticated attacker can open connections to port 8200 and send data without newlines, growing the GenServer state without limit until server memory is exhausted.

This was validated against a local test harness and confirmed reachable on production:                                                 
```bash
  $ nc server4.beyondallreason.info 8200                                                                                                 
  TASSERVER 0.38-33-ga5f3b28 * 8201 0   
```                          
                                                                                                                                         
  Combined with max_connections: :infinity on the Ranch listener, this is a trivial remote DoS.                                          
                                                
  64KB is ~100x larger than any legitimate Spring protocol command.                                                                      
                                                                                                                                                                                                                                                                                                              
  Tested against BAR-Devtools local harness:                                                                                             
  Overflow rejected:                                                                                                                     
  Sent 100KB without newline                                      
  Server response: s.system.disconnect Message too large
  PASS: Connection dead after overflow                  
                                      
  Normal fragmentation preserved:
  Fragmented PING (sent as "PI" + "NG\n"): PONG                                                                                          
  PASS: Normal TCP fragmentation works         
                                                                                                                                         
  Server log:                                                                                                                            
 ` [warning] Disconnecting client 172.20.0.1: message buffer exceeded 65536 bytes`
 
 LLM Usage: Claude Opus 4.6 for bug validation and tests


lib/teiserver/protocols/spring/spring_in.ex
```diff
@@ -74,7 +80,19 @@ defmodule Teiserver.Protocols.SpringIn do
...
-        %{state | message_part: state.message_part <> data}
+        new_buffer = state.message_part <> data
+
+        if byte_size(new_buffer) > @max_message_part_size do
+          Logger.warning(
+            "Disconnecting client #{state.ip}: message buffer exceeded #{@max_message_part_size} bytes"
+          )
+
+          SpringOut.reply(:disconnect, "Message too large", nil, state)
+          send(self(), :terminate)
```
Teifion: Suggestion, we don't disconnect the person only clear the buffer.

lib/teiserver/protocols/spring/spring_in.ex
```diff
@@ -52,6 +52,12 @@ defmodule Teiserver.Protocols.SpringIn do
     c.auth.login_queue_heartbeat
   )
 
+  # Maximum size of the partial message buffer per connection. If a client sends
+  # more than this many bytes without a newline, the connection is terminated.
+  # Without this limit, an attacker can grow message_part without bound by
+  # sending data without newlines, exhausting server memory.
+  @max_message_part_size 64 * 1024
```
Teifion: Could we move this to a runtime config, probably in `Teiserver.Config.SiteConfigTypes.SystemConfigs`.

lib/teiserver/protocols/spring/spring_in.ex
```diff
@@ -74,7 +80,19 @@ defmodule Teiserver.Protocols.SpringIn do
...
       else
-        %{state | message_part: state.message_part <> data}
+        new_buffer = state.message_part <> data
+
+        if byte_size(new_buffer) > @max_message_part_size do
+          Logger.warning(
+            "Disconnecting client #{state.ip}: message buffer exceeded #{@max_message_part_size} bytes"
+          )
+
+          SpringOut.reply(:disconnect, "Message too large", nil, state)
```
Teifion: It would help for well meaning users hitting this limit to know how big their message is if we told them. Bad meaning users would know the limit already and be working with that info.

```suggestion
          SpringOut.reply(:disconnect, "Message too large (#{byte_size(new_buffer)})", nil, state)
```

Teifion (commented):
Good observation, I like this PR and with some minor changes I would be happy to approve it.

lib/teiserver/protocols/spring/spring_in.ex
NortySpock: I liked the Teifion's Discord suggestion of `Message exceeds max length of n (message was m long)`

lib/teiserver/protocols/spring/spring_in.ex
```diff
@@ -63,6 +63,9 @@ defmodule Teiserver.Protocols.SpringIn do
       end
     end
 
+    max_buffer_size =
+      Config.get_site_config_cache("teiserver.Spring max message buffer size")
+
```
L-e-x-o-n: We only really need this in the else block, this code is ran quite often so I would like to avoid unnecessary calls.

thvl3: Good call

NortySpock: ~~I admit I'm not quite following how the "case with newline" works here, what happens if someone sends a message that is 9MB in size but it contains a newline? It would have a multi-megabyte buffer to process that gets split on the newline? What constrains the buffer size in that case?~~

edit: nevermind, I think I can see it's accumulating parts of a line. 
 1) if a newline is encountered, it splits on the newline and it attempts to treat it as a message to process
 2) otherwise, it continues accumulating a single line of text until it hits max_buffer_size, because that accumulator is the only way multiple packets would add up to a payload that exceeds the max_buffer_size. Thus this is the only code path that needs a check.

L-e-x-o-n (approved):

Teifion (approved):
Nice work, if you fix the merge conflicts and ping me here or discord I'll hit the merge button.

PR #1083, opened 2026-04-12
Some safety fixes

geekingfrog:
Some minor fixes because when you tell a llm to find security bug, they will find some, even when they're minor/bogus.


Teifion (approved):

Teifion:
LGTM so far :D

Edit, missed that it was a draft and approved, sorry.

PR #1085, opened 2026-04-12
Refine flood protection logic to exempt heartbeat for authenticated connections

thvl3:
The flood protection in spring_tcp_server.ex only counts commands that contain a newline character. Partial TCP data bypasses the counter entirely. c.auth.login_queue_heartbeat is also explicitly exempt, even for unauthenticated connections.

This means an unauthenticated client can send data to the server without ever triggering flood protection.

Relevant code: flood_protect?/2 at spring_tcp_server.ex:1167-1183

Fix: count all incoming data toward the flood limit regardless of newline presence, and remove the heartbeat exemption for unauthenticated connections


Teifion (approved):
Good catch.

FWIW I think the newline was with the intent to flood protect on commands not messages but I think messages is the better idea.

PR #1086, opened 2026-04-12
Add telemetry burst rate limiting for unauthenticated users

thvl3:
Several c.telemetry.* commands are in the @unauthenticated_commands whitelist in spring_in.ex and write directly to the database without any auth check or rate limiting. This includes simple_client_event, complex_client_event, update_client_property, and upload_infolog.

The _test variants of these commands correctly deny unauthenticated users -- the non-test variants don't have the same check.

Confirmed locally via BAR-Devtools and verified the commands are accepted on prod.

Fix: either remove the telemetry commands from @unauthenticated_commands or add per-IP rate limiting for anonymous writes.

### DIFF
--- a/lib/teiserver/libs/teiserver_configs.ex
+++ b/lib/teiserver/libs/teiserver_configs.ex
@@ -217,6 +217,17 @@ defmodule Teiserver.TeiserverConfigs do
     })
 
     add_site_config_type(%{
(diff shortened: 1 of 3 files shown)


Teifion (approved):
Also a good catch; thank you.

thvl3:
@Teifion  FYI I don't have write permissions to merge these PRs so a maintainer will need to

PR #1089, opened 2026-04-13
Extend plugin system

Teifion:
Adds more plugin decorators

### DIFF
--- a/lib/teiserver/application.ex
+++ b/lib/teiserver/application.ex
@@ -5,10 +5,14 @@ defmodule Teiserver.Application do
 
   alias Phoenix.PubSub
   alias Teiserver.Helper.ObanLogger
+  alias Teiserver.Plugins
   alias Teiserver.Startup
   alias TeiserverWeb.Endpoint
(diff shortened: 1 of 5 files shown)


L-e-x-o-n (approved):

lib/teiserver/application.ex
```diff
@@ -211,6 +216,11 @@ defmodule Teiserver.Application do
     Startup.startup()
   end
 
+  @decorate Plugins.plugin(:additional_application_children)
```
geekingfrog, line 219: I think this is where the approach of having keys from the config starts to fall apart.
Because now, if a plugin requires a background process (a cache perhaps), and a callback somewhere else, you need to define the set the two things in the config. There is no logical link between the two.

That was the main reason behind having a list of modules instead, that can export some of these functions to be picked up by the main code.

Though this does raise the question of what to do when multiple plugins are configured for the same hook. I think we can accept many when the result doesn't matter, and raise an exception when we need the result and more than one plugin implements a given hook.

PR #1095, opened 2026-04-15
fix: extend max length for complex client events

bruno-dasilva:
Complex_client_events were silently dropped if they were above the 1024 length limit. Benchmark results are in the neighborhood of ~1500 characters.

Added a log for when they're dropped, and increased the drop threshold to 4096 for JUST complex events. Other payloads remain at a threshold of 1024.

Tested: data makes it into my local db! 
image

### DIFF
--- a/lib/teiserver/protocols/spring/spring_telemetry_in.ex
+++ b/lib/teiserver/protocols/spring/spring_telemetry_in.ex
@@ -12,6 +12,8 @@ defmodule Teiserver.Protocols.Spring.TelemetryIn do
   import SpringIn, only: [_no_match: 4]
   # import Teiserver.Helper.NumberHelper, only: [int_parse: 1]
 
+  @complex_client_event_max_length 4096
+
   # TODO: Less nested hackyness
   @spec do_handle(String.t(), String.t(), String.t() | nil, map()) :: map()
   def do_handle("upload_infolog", data, msg_id, state) do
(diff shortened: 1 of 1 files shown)


lib/teiserver/protocols/spring/spring_telemetry_in.ex
```diff
@@ -145,8 +145,10 @@ defmodule Teiserver.Protocols.Spring.TelemetryIn do
     end
   end
 
+  @complex_client_event_max_length 4096
```
Teifion: This should be near the top of the file with any other module attributes

lib/teiserver/protocols/spring/spring_telemetry_in.ex
```diff
@@ -169,6 +171,11 @@ defmodule Teiserver.Protocols.Spring.TelemetryIn do
           "no match"
       end
     else
+      Logger.warning(
+        "log_client_event exceeds max_length: length=#{String.length(data)} " <>
+          "max=#{@complex_client_event_max_length}"
+      )
```
Teifion, line 176: No string concatenation needed

```suggestion
      Logger.warning(
        "log_client_event exceeds max_length: length=#{String.length(data)} max=#{@complex_client_event_max_length}"
      )
```

Teifion (commented):
Minor tweaks but overall good and will be happy to merge once these are addressed.

lib/teiserver/protocols/spring/spring_telemetry_in.ex
bruno-dasilva: was doing it to keep line length lower but this is fine too :)

Teifion (approved):
Looks good to me, once the CI goes green I will merge it :)

PR #1104, opened 2026-04-21
Lobby bosses

geekingfrog:
* can enable bosses for lobbies
* can appoint new bosses
* can unboss players. Currently there is no restriction, any boss can unboss any other boss. This will likely change
* only boss can change the map. We'll extend this to other operations later (they also need protocol support).
* leaving the lobby unboss

### DIFF
--- a/test/teiserver/tachyon_lobby/lobby_test.exs
+++ b/test/teiserver/tachyon_lobby/lobby_test.exs
@@ -1242,6 +1242,17 @@ defmodule Teiserver.TachyonLobby.LobbyTest do
       assert_receive {:lobby, ^id, {:updated, data}}
       assert data.bosses == %{"2" => %{}}
     end
+
+    test "leaving lobby unboss" do
+      %{id: id} = setup_full_lobby([1, 1], boss_enabled?: true)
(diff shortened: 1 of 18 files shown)


test/teiserver/tachyon_lobby/lobby_test.exs
```diff
@@ -1123,6 +1143,115 @@ defmodule Teiserver.TachyonLobby.LobbyTest do
...
+
+    test "unboss - must be in lobby" do
+      %{id: id} = setup_full_lobby([1, 1], boss_enabled?: true)
+      {:error, :not_in_lobby} = Lobby.unboss(id, "not-in-lobby", @default_user_id)
+    end
+
+    test "unboss - target must be boss" do
+      %{id: id} = setup_full_lobby([1, 1], boss_enabled?: true)
+      {:error, :not_a_boss} = Lobby.unboss(id, @default_user_id, "2")
+    end
```
L-e-x-o-n: Appointing an existing boss is a no-op, IMO unbossing a non boss should be as well.

geekingfrog: done (amended the commit)

PR #1105, opened 2026-04-22
Removed client index list page

Teifion:
This page was added when we had 50 or so concurrent users and was helpful for debugging state. These days it is impractical to use it and the tests for it are not even run so it makes sense to remove it entirely.


L-e-x-o-n:
I still find it useful, especially when testing locally or on integration server where there are fewer clients.

Teifion:
Oooh, that's useful to know. I will refactor this to instead fix the test and put some limits on the number displayed if needed etc.

Teifion:
Closing to replace with something to fix the test and keep using the page.

PR #1106, opened 2026-04-22
Remove lobby policies

Teifion:
Lobby policies are not currently used, this will remove them from the codebase.


priv/repo/migrations/20260422093438_remove_lobby_policies.exs
```diff
@@ -0,0 +1,15 @@
...
+  def up do
+    alter table(:teiserver_battle_matches) do
+      remove :lobby_policy_id
+    end
+
+    drop table(:lobby_policies)
+  end
+
+  def down do
+    raise "Cannot revert removal of lobby policies"
```
L-e-x-o-n, line 13: I usually prefer `change/0` instead of just raising an error.
Seems cleaner in any case: if it can revert it will do so and only one function is necessary, if it can't it will raise an error anyway.

Teifion: Yeah, in this case I wanted to make it abundantly clear you cannot revert the migration.

Teifion:
@L-e-x-o-n I have rebased to remove merge conflicts, ready for review.

PR #1107, opened 2026-04-22
Fixes test for client index

Teifion:
Fixes a bug and the relevant tests for Client index.

### DIFF
--- a/lib/teiserver/account/client_index_throttle.ex
+++ b/lib/teiserver/account/client_index_throttle.ex
@@ -5,7 +5,6 @@ defmodule Teiserver.Account.ClientIndexThrottle do
   player_changes lists players that have changed (added, updated or removed!)
   """
   alias Phoenix.PubSub
-  alias Teiserver.Account.ClientIndexThrottle
   alias Teiserver.Client
 
   use GenServer
@@ -73,7 +72,7 @@ defmodule Teiserver.Account.ClientIndexThrottle do
   end
 
   def start_link(opts) do
-    GenServer.start_link(ClientIndexThrottle, opts[:data], [])
+    GenServer.start_link(__MODULE__, opts[:data], [])
   end
 
   def init(_opts) do
@@ -85,7 +84,7 @@ defmodule Teiserver.Account.ClientIndexThrottle do
 
     Horde.Registry.register(
       Teiserver.ThrottleRegistry,
-      "ClientIndexThrottle",
+      "__MODULE__",
       :index
     )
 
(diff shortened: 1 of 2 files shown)


lib/teiserver/account/client_index_throttle.ex
```diff
@@ -5,7 +5,9 @@ defmodule Teiserver.Account.ClientIndexThrottle do
   player_changes lists players that have changed (added, updated or removed!)
   """
   alias Phoenix.PubSub
+  alias Teiserver.Account.ClientIndexThrottle
```
geekingfrog: hum, you are aliasing the current module? I find that a bit weird tbh, and later I prefer `__MODULE__`.
I find that clearer. Where I see an alias, I assume it's another module

test/teiserver_web/live/client_live_test.exs
```diff
@@ -12,14 +12,23 @@ defmodule TeiserverWeb.Live.ClientTest do
...
 
   setup do
-    GeneralTestLib.conn_setup(TeiserverTestLib.admin_permissions())
-    |> TeiserverTestLib.conn_setup()
+    {:ok, setup_result} =
+      TeiserverTestLib.admin_permissions()
+      |> GeneralTestLib.conn_setup()
+      |> TeiserverTestLib.conn_setup()
+
+    {:ok, throttle_pid} = ClientIndexThrottle.start_link(data: %{})
```
geekingfrog: Prefer [start_link_supervised/2](https://hexdocs.pm/ex_unit/main/ExUnit.Callbacks.html#start_link_supervised!/2). This also allow you to skip the `on_exit` callback

lib/teiserver/account/client_index_throttle.ex
Teifion: I used to prefer `__MODULE__` but have since moved to preferring aliases as I find that clearer.

test/teiserver_web/live/client_live_test.exs
Teifion: Good idea, addressed in https://github.com/beyond-all-reason/teiserver/pull/1107/commits/c913c4c4cc76f252c15fd84da93856b00b735ef6

lib/teiserver/account/client_index_throttle.ex
L-e-x-o-n: We are using `__MODULE__` everywhere else so let's use it here as well for consistency.
If we agree on changing it, that can be done everywhere as part of a separate PR.

Teifion: Fair, addressed in https://github.com/beyond-all-reason/teiserver/pull/1107/commits/653cbe26bb5e6de080e71f1e7f418627a0599687

L-e-x-o-n (approved):

PR #1110, opened 2026-04-23
Addressed flakey tests

Teifion:
This PR addressed a number of flakey tests and also fixes an issue where permissions could allow a moderator to do something but not an admin.

### DIFF
--- a/lib/teiserver/account/auth.ex
+++ b/lib/teiserver/account/auth.ex
@@ -43,9 +43,6 @@ defmodule Teiserver.Account.Auth do
   def is_bot?(%{roles: roles}), do: Enum.member?(roles, "Bot")
   def is_bot?(_user), do: false
 
-  @spec admin_or_moderator?(T.userid() | T.user()) :: boolean()
-  def admin_or_moderator?(u), do: moderator?(u) or admin?(u)
-
   @spec moderator?(T.userid() | T.user()) :: boolean()
   def moderator?(nil), do: false
 
--- a/lib/teiserver/account/auth.ex
+++ b/lib/teiserver/account/auth.ex
@@ -43,7 +43,9 @@ defmodule Teiserver.Account.Auth do
   def is_bot?(%{roles: roles}), do: Enum.member?(roles, "Bot")
   def is_bot?(_user), do: false
 
-  # credo:disable-for-lines:5 Credo.Check.Readability.PredicateFunctionNames
+  @spec admin_or_moderator?(T.userid() | T.user()) :: boolean()
+  def admin_or_moderator?(u), do: moderator?(u) or admin?(u)
+
   @spec moderator?(T.userid() | T.user()) :: boolean()
   def moderator?(nil), do: false
 
--- a/lib/teiserver/application.ex
+++ b/lib/teiserver/application.ex
@@ -184,8 +184,11 @@ defmodule Teiserver.Application do
     start_result = Supervisor.start_link(children, opts)
 
     # If the server seems to not start up we can check the info log to see
-    # if this is present
-    Logger.info("Teiserver.Supervisor start result: #{Kernel.inspect(start_result)}")
+    # if this is present, we don't want it appearing in our tests though
+    # TODO: Replace this with something having less of a code-smell
+    if not Application.get_env(:teiserver, Teiserver)[:test_mode] do
+      Logger.error("Teiserver.Supervisor start result: #{Kernel.inspect(start_result)}")
+    end
 
     startup_sub_functions(start_result)
 
--- a/lib/teiserver/application.ex
+++ b/lib/teiserver/application.ex
@@ -183,9 +183,9 @@ defmodule Teiserver.Application do
     opts = [strategy: :one_for_one, name: Teiserver.Supervisor]
     start_result = Supervisor.start_link(children, opts)
 
-    # We use a logger.error to ensure something appears even on the error logs
-    # and we can be sure they're being written to
-    Logger.error("Teiserver.Supervisor start result: #{Kernel.inspect(start_result)}")
+    # If the server seems to not start up we can check the info log to see
+    # if this is present
+    Logger.info("Teiserver.Supervisor start result: #{Kernel.inspect(start_result)}")
 
     startup_sub_functions(start_result)
 
--- a/lib/teiserver/coordinator/consul_server.ex
+++ b/lib/teiserver/coordinator/consul_server.ex
@@ -199,7 +199,7 @@ defmodule Teiserver.Coordinator.ConsulServer do
     new_approved = [userid | state.approved_users] |> Enum.uniq()
 
     username = Account.get_username(userid)
-    ChatLib.persist_system_message("#{username} joined the lobby", state.lobby_id)
+    maybe_persist_system_message("#{username} joined the lobby", state.lobby_id)
 
     {:noreply,
      %{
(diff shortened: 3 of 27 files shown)


lib/teiserver/account/auth.ex
```diff
@@ -43,7 +43,9 @@ defmodule Teiserver.Account.Auth do
   def is_bot?(%{roles: roles}), do: Enum.member?(roles, "Bot")
   def is_bot?(_user), do: false
 
-  # credo:disable-for-lines:5 Credo.Check.Readability.PredicateFunctionNames
+  @spec admin_or_moderator?(T.userid() | T.user()) :: boolean()
+  def admin_or_moderator?(u), do: moderator?(u) or admin?(u)
```
geekingfrog: That seems quite specific. I think using `Auth.admin?(u) or Auth.moderator?(u)` works just fine.
My issue with that kind of function is that they don't compose well. What's next? `admin_or_moderator_or_contributor?`, `contributor_but_no_admin?`, there is just no end to these, but you can easily express the same with the regular `or` and `and`

test/teiserver/battle/lobby_server_test.exs
```diff
@@ -7,7 +7,7 @@ defmodule Teiserver.Battle.LobbyServerTest do
   # Cannot be async because some other tests will call for a list of all lobbies
   use Teiserver.DataCase, async: false
 
-  @tag :needs_attention
+  # @tag :needs_attention
```
geekingfrog: just remove the line

lib/teiserver/account/auth.ex
L-e-x-o-n: This also seems to be a consequence of issues elsewhere, specifically `moderator?` and `admin?` checking roles instead of permissions (I am working on improving this elsewhere) every Admin should have the Moderator permissions even if not necessarily the Moderator role.

lib/teiserver/coordinator/consul_server.ex
```diff
@@ -1437,6 +1437,14 @@ defmodule Teiserver.Coordinator.ConsulServer do
     |> queue_size_changed()
   end
 
+  # In tests this can lead to generating foreign key issues so we don't do this in tests
+  # but the function we are wrapping is tested elsewhere
+  defp maybe_persist_system_message(message, lobby_id) do
+    if not Application.get_env(:teiserver, Teiserver)[:test_mode] do
+      ChatLib.persist_system_message(message, lobby_id)
+    end
+  end
```
L-e-x-o-n, line 1451: Which foreign key was the issue and why is it a problem here?
Is there some other way this can be resolved without test mode specific code?

lib/teiserver/application.ex
```diff
@@ -183,9 +183,9 @@ defmodule Teiserver.Application do
     opts = [strategy: :one_for_one, name: Teiserver.Supervisor]
     start_result = Supervisor.start_link(children, opts)
 
-    # We use a logger.error to ensure something appears even on the error logs
-    # and we can be sure they're being written to
-    Logger.error("Teiserver.Supervisor start result: #{Kernel.inspect(start_result)}")
+    # If the server seems to not start up we can check the info log to see
+    # if this is present
+    Logger.info("Teiserver.Supervisor start result: #{Kernel.inspect(start_result)}")
```
L-e-x-o-n: I got used to this being in the error log and would like to keep it at this point, it's a convenient separator of pre and post server (re)start errors.

lib/teiserver_web/live/clients/index.html.heex
```diff
@@ -59,7 +59,7 @@
               <%= if Enum.member?(@client_ids, userid) do %>
                 <% user = @users[userid]
                 client = @clients[userid] %>
-                <%= if client != nil do %>
+                <%= if client != nil and user != nil do %>
```
L-e-x-o-n, line 62: How can an `user`  be `nil`  in the first place? That seems like a possible bigger issue.

lib/teiserver/account/auth.ex
Teifion: > I think using Auth.admin?(u) or Auth.moderator?(u) works just fine.

I think that's fair, I will change it over

> every Admin should have the Moderator permissions even if not necessarily the Moderator role.

That was my read too, e.g. Admins won't get a green name in the client because they won't show up as Moderators to the protocol which I took to be intentional. Is that not the case?

lib/teiserver/coordinator/consul_server.ex
Teifion: It was the lobby or match key for the message; the cause was when a test was being torn down the consul would see someone left the lobby and persist the message but the DB teardown was already in progress thus causing the error but not messing with the test itself. I will extend the comment.

lib/teiserver/application.ex
Teifion: I'd really rather the error wasn't present for tests so I will put a wrapper around it for prod :)

lib/teiserver_web/live/clients/index.html.heex
Teifion: It can be nil in certain specific test situations from my understanding, the additional code in the liveview file itself allowed this to sometimes happen and this seemed the easier fix. If it was a bigger issue we'd be seeing it in many other places too.

lib/teiserver/account/auth.ex
Teifion: I have refactored the function in https://github.com/beyond-all-reason/teiserver/pull/1110/commits/9ec5b4864cf32234ea9df53db48acab0c9c532e7

test/teiserver/battle/lobby_server_test.exs
Teifion: Ooopsie, addressed in https://github.com/beyond-all-reason/teiserver/pull/1110/commits/9ec5b4864cf32234ea9df53db48acab0c9c532e7

lib/teiserver/coordinator/consul_server.ex
Teifion: Comment added of:
```elixir
# In tests this can lead to generating foreign key issues so we don't do this in tests
  # but the function we are wrapping is tested elsewhere.
  # The error happens because when a test is closed down the relevant match
  # entry from the database is removed and we still try to insert this generating:
  # ** (Ecto.ConstraintError) constraint error when attempting to insert struct:
  #
  #   * "teiserver_lobby_messages_user_id_fkey" (foreign_key_constraint)
  defp maybe_persist_system_message(message, lobby_id) do
    if not Application.get_env(:teiserver, Teiserver)[:test_mode] do
      ChatLib.persist_system_message(message, lobby_id)
    end
  end
```

lib/teiserver/application.ex
Teifion: Now wrapped so not to appear in tests in https://github.com/beyond-all-reason/teiserver/pull/1110/commits/9ec5b4864cf32234ea9df53db48acab0c9c532e7

geekingfrog (approved):

lib/teiserver/account/auth.ex
L-e-x-o-n: > That was my read too, e.g. Admins won't get a green name in the client because they won't show up as Moderators to the protocol which I took to be intentional. Is that not the case?

That is the case and the way to achieve this IMO is with permission checks for command usage and role check for protocol moderator flag. This should be addressed in a different PR however, I already have some work in progress on a related topic.

L-e-x-o-n (approved):

PR #1121, opened 2026-04-28
Expanded lobby naming logic

Teifion:
- Expands `allowed_to_set_restrictions/1` to also take an atom declaring if the limit check is a min, max or either
- Setting minimum limits is not allowed in noob games
- Setting maximum limits is not allowed in pro games
- Removed some consul commands related to these which are not used/supported
- Fixed a bug where using `setrating` would allow you to set a minimum rating even in a noob game


L-e-x-o-n (commented):
`minimum_uncertainty_to_play` and `maximum_uncertainty_to_play` are now unused and all mentions of it can be removed completely.

L-e-x-o-n (approved):

PR #1123, opened 2026-04-28
More log tidying

geekingfrog:


config/config.exs
```diff
@@ -157,6 +157,10 @@ config :teiserver, Teiserver,
     battle_minimum_seconds: 120
   }
 
+config :nostrum,
+  youtubedl: false,
+  streamlink: false
+
```
L-e-x-o-n, line 164: I remember seeing `ffmpeg` as well, maybe add that one as well?

PR #1133, opened 2026-05-06
Upgraded version of plug_cowboy

Teifion:
A [vulnerability in plug_cowboy](https://github.com/advisories/GHSA-q8x4-x7mp-5vg2) has been flagged by the CI process.


L-e-x-o-n (approved):

NortySpock:
According to the link, it only affected 2.8.0 and below, so I didn't get the impression we were vulnerable when I looked it up. 

(Otherwise I would have said something earlier)

Teifion:
We had a CI fail which is what caused this to have to be done. I was also surprised but CI is CI :)

Sent from [Proton Mail](https://proton.me/mail/home) for Android.

-------- Original Message --------

PR #1136, opened 2026-05-07
shuffle team assignments after balancing

### LINKED ISSUE #1000: [Bug]: Lower rated player always in first team when balancing Duel matches
### Describe the Bug

Default balancer for Duel rating type (`respect_avoids` in this case) seems to always put the lower rated player in team 0 (shown as team 1 in Chobby) and the higher rated one in team 1 (shown as team 2 in Chobby).

This is an issue because map starting positions depend on team order.

The fix should always randomise team order after the balance algorithm is ran, this should probably be done in `Teiserver.Game.BalancerServer` so that it''s independent on the result of a specific balance algorithm.

### Reproduce the bug

User `respect_avoids` balance algorithm (I suspect `split_noobs` might have the same issue) to balance 2 players of different ratings. Lower rated player should always be on first team.

BElluu:


test/teiserver/game/balancer_server_test.exs
```diff
@@ -181,4 +181,41 @@ defmodule Teiserver.Game.BalancerServerTest do
     poll_until_nil(fn -> GenServer.call(pid, :get_current_balance) end)
     assert GenServer.call(pid, :report_state) |> Map.get(:last_balance_hash, :not_found) == nil
   end
+
+  test "swap_teams swaps all team-keyed maps between team 1 and team 2" do
```
Teifion: This is nice, can we also see a test for Team FFA to ensure we don't break on N teams please.

Teifion (changes requested):
Nice work, I think this is a good PR and aside from a possible Team FFA issue I think this is ready. Add the test and I'll approve/merge it :)

Teifion:
@BElluu I agree this prevents it breaking but can we find a way to shuffle the teams in a team ffa? One of the reasons for the change is better evaluation of map balance and it would be a shame to only be able to do that for 2 team games. I am happy to jump into a voice chat on the discord to chat and brainstorm ways to do better if it would help :)

BElluu:
> @BElluu I agree this prevents it breaking but can we find a way to shuffle the teams in a team ffa? One of the reasons for the change is better evaluation of map balance and it would be a shame to only be able to do that for 2 team games. I am happy to jump into a voice chat on the discord to chat and brainstorm ways to do better if it would help :)

Ahhh, I just didn't understand. I don't know why I thought we didn't want to swap team positions for FFA. I'll fix that :)

BElluu:
@Teifion After changes should be better :)

L-e-x-o-n (approved):
Thanks for fixing this

Teifion (approved):
Thanks for fixing this, approved and merged!

Issue #1068, opened 2026-03-31
Tidy overly verbose logs


geekingfrog:
We are getting quite a few logs withe the following shape. They don't seem particularly useful, so it would be nice to either silence them or make them a lot less verbose.
I have no idea how to reproduce/trigger that however :/

```
2026-03-31 17:11:15.519 [error] pid=<0.54902007.0>  Ranch listener TeiserverWeb.Endpoint.HTTPS, connection process #PID<0.54902007.0>, stream 1 had its request process #PID<0.50429732.0> exit with reason {{{%Body
guard.NotAuthorizedError{
     message: "not authorized",
     status: 403,
     reason: {:error,
      :unauthorized}
   },
   [
     {Bodyguard,
      :permit!,
      5,
      [
        file: ~c"lib/bodyguard.ex",
        line: 64
      ]},
     {Bodyguard.Plug.Authorize,
      :call,
      2,
      [
        file: ~c"lib/bodyguard/plug/authorize.ex",
        line: 116
      ]},
     {TeiserverWeb.Admin.UserController,
      :phoenix_controller_pipeline,
      2,
      [
        file: ~c"lib/teiserver_web/controllers/admin/user_controller.ex",
        line: 1
      ]},
     {Phoenix.Router,
      :__call__,
      5,
      [
        file: ~c"lib/phoenix/router.ex",
        line: 484
      ]},
     {TeiserverWeb.Endpoint,
      :plug_builder_call,
      2,
      [
        file: ~c"lib/teiserver_web/endpoint.ex",
        line: 1
      ]},
     {TeiserverWeb.Endpoint,
      :call,
      2,
      [
        file: ~c"lib/teiserver_web/endpoint.ex",
        line: 1
      ]},
     {Plug.Cowboy.Handler,
      :init,
      2,
      [
        file: ~c"lib/plug/cowboy/handler.ex",
        line: 11
      ]},
     {:cowboy_handler,
      :execute,
      2,
      [
        file: ~c"/build/deps/cowboy/src/cowboy_handler.erl",
        line: 37
      ]}
   ]},
  {TeiserverWeb.Endpoint,
   :call,
   [
     %Plug.Conn{
       adapter: {Plug.Cowboy.Conn,
        :...},
       assigns: %{},
       body_params: %Plug.Conn.Unfetched{
         aspect: :body_params
       },
       cookies: %Plug.Conn.Unfetched{
         aspect: :cookies
       },
       halted: false,
[...]

PR #1139, opened 2026-05-08
Silence Bodyguard not-authorized log noise

EeroVakiparta:
### Silence Bodyguard `NotAuthorizedError` noise in the production error log
Issue: [#1068](https://github.com/beyond-all-reason/teiserver/issues/1068)

#### Problem
When an authenticated user without the required role hits a Bodyguard-protected URL, Phoenix returns 403 cleanly (`Bodyguard.NotAuthorizedError` implements `Plug.Exception`), but the request process still exits with the exception as its reason. Cowboy then logs the abnormal request-process exit at `[error]` level, dumping a multi-line stack trace into the error log for every failed authorization. That's the noise #1068 is about.

#### Root cause
The log is emitted from [`cowboy_stream_h.erl:149`](https://github.com/ninenines/cowboy/blob/2.14.2/src/cowboy_stream_h.erl#L149) via `cowboy:log/2`. Plug.Cowboy does not set the cowboy `logger` protocol option, so cowboy falls through to its default path: [`cowboy.erl:241`](https://github.com/ninenines/cowboy/blob/2.14.2/src/cowboy.erl#L241) calls `:error_logger.error_msg(Format, Args)`. In OTP 26 that maps to `logger:log(error, Report, Meta)` ([kernel/src/error_logger.erl:241](https://github.com/erlang/otp/blob/OTP-26.2.5.14/lib/kernel/src/error_logger.erl#L241)) with the report being:

```elixir
%{label: {:error_logger, :error_msg},
  format: ~c"... had its request process ~p exit with reason ~0p~n",
  args:   [Ref, ConnPid, StreamID, ReqPid, Reason]}
```

So the noise arrives at `:logger` as a structured `:report` event, not a `{Format, Args}` event.

#### Fix
A `:logger` primary filter in `TeiserverWeb.LoggerFilter` that drops the event only when **all four** of these match:

- `level: :error`
- the `:msg` is the structured `error_logger:error_msg/2` report (or, as a fallback, the legacy `{Format, Args}` shape used if cowboy is configured with `logger: :logger`)
- the format string contains cowboy's signature substring `"had its request process"`
- exactly five args, with a `%Bodyguard.NotAuthorizedError{}` found while walking only the fifth (`Reason`) arg structurally

The four-condition match keeps the filter narrow enough that an unrelated log carrying the same exception in some other shape (for example application code logging an auth failure manually) still flows through normally. The 403 response itself is unchanged.

Installed from `Teiserver.Application.start/2` via `TeiserverWeb.LoggerFilter.install_primary!/0` rather than `config/*.exs`, because configs can be read before user modules are loaded into the BEAM and the function reference would fail to resolve.

#### Verification
This is the load-bearing part. To confirm the filter actually fires against the real production path and not against a synthetic event shape:

1. I commented out `install_primary!()` in `Application.start/2` and re-ran the filter test file.
2. The end-to-end test (`real cowboy path via :error_logger.error_msg/2 is suppressed end-to-end`) failed exactly as expected — captured log contained the cowboy `"Ranch protocol ... terminated, exit reason ... %Bodyguard.NotAuthorizedError{...}"` string.
3. Restoring the registration brought all 14 tests back to green.

The end-to-end test calls `:error_logger.error_msg/2` directly, mirroring what cowboy does internally, so the regression test exercises the actual shape that arrives at `:logger` in production rather than a more convenient synthetic shape.

#### Test coverage
- 7 unit tests over the report shape (positive matches; negative matches for different exception, unrelated format, wrong label, wrong arg count, non-error level; recursion into nested reasons)
- 2 unit tests over the legacy `{Format, Args}` shape
- 1 unit test for malformed events (filter ignores them rather than crashing)
- 3 integration tests via `ExUnit.CaptureLog`: filter is registered at startup, real cowboy path is suppressed end-to-end, unrelated cowboy-shaped reports still flow through
- 1 controller-level auth-boundary test (non-admin Verified user gets 403 on `/teiserver/admin/user/:id`). It's annotated as documentary about the filter, since `Plug.Test` does not go through cowboy and so the noise log doesn't fire under the test transport.

#### Notes
- The filter is active in `MIX_ENV=test` as well. Given the four-condition match, accidental matches from real test failures are effectively impossible.
- Continues the recent log-tidying effort (#1123) but in a different file/path so there's no overlap.
- `mix format`, `mix dialyzer`, and the affected-modules tests all pass locally.

> **AI usage disclosure**: AI was used to assume roles of three repository's top contributors. Fix was iterated countless times untill all three approved it without nitpicks. All documentation is AI generated to achieve higher quality english but is human verified. Claude Opus 4.7 & GPT-5.5.

### DIFF
--- a/lib/teiserver/application.ex
+++ b/lib/teiserver/application.ex
@@ -8,6 +8,7 @@ defmodule Teiserver.Application do
   alias Teiserver.Plugins
   alias Teiserver.Startup
   alias TeiserverWeb.Endpoint
+  alias TeiserverWeb.LoggerFilter
   alias TeiserverWeb.Monitoring.Router, as: MonitoringRouter
 
   use Plugins
@@ -20,6 +21,8 @@ defmodule Teiserver.Application do
 
   @impl Application
   def start(_type, _args) do
+    LoggerFilter.install_primary!()
+
     LoggerBackends.add(LoggerBackends.Console)
     LoggerBackends.add({LoggerFileBackend, :error_log})
     LoggerBackends.add({LoggerFileBackend, :notice_log})
--- a/lib/teiserver_web/logger_filter.ex
+++ b/lib/teiserver_web/logger_filter.ex
@@ -1,0 +1,102 @@
+defmodule TeiserverWeb.LoggerFilter do
+  @moduledoc "Custom `:logger` primary filters."
+
+  @primary_filter_name :teiserver_bodyguard_not_authorized
+
+  @doc """
+  Installs every primary `:logger` filter defined in this module.
+
+  Called from `Teiserver.Application.start/2`. Doing this at app start (rather
+  than in `config/*.exs`) guarantees that this module is loaded by the time
+  `:logger` is asked to call into it.
+  """
+  @spec install_primary!() :: :ok
+  def install_primary! do
+    case :logger.add_primary_filter(
+           @primary_filter_name,
+           {&__MODULE__.drop_bodyguard_not_authorized/2, []}
+         ) do
+      :ok -> :ok
+      {:error, {:already_exists, @primary_filter_name}} -> :ok
+    end
+  end
+
+  @doc """
+  Drops Ranch "request process exit" logs whose exit reason is a
+  `Bodyguard.NotAuthorizedError`.
+
+  These fire whenever an authenticated user without the required role hits a
+  protected URL. Phoenix has already returned a 403 to the client by the time
+  the process exits, so the stack trace in the error log is just noise.
+  See #1068.
+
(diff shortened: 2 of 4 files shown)


EeroVakiparta:
Side note for after/if this merges: the "Bodyguard raises → cowboy logs the exit" pattern is supressed here, not really fixed. Could be worth a followup to switch the ~47 Bodyguard.Plug.Authorize calls to use :fallback so the exception never raises. Happy to open an issue/PR for that if there's interest, but maybe someone else should do such wide fix, not me.

geekingfrog:
As you mentioned, you are dropping the logs, but this is an ostrich move. The error is somewhere in our code, we need to find the root cause and fix it.
No amount of roleplay with your slop machines pointed the fundamentally flawed approach?

PR #1142, opened 2026-05-08
Keep session alive during battle when player disconnect

### LINKED ISSUE #920: [Feature]: Keep session if player is in game
In case the player disconnect abruptly (that is, without sending `system/disconnect` request), the session will then stop a short time after.

But if the player is currently in game, this timeout logic should instead be tied to the battle in progress, not to the connection. This would avoid a player crashing then having major trouble restarting the PC and game, so they can always rejoin the battle.

If there is still no connection when the battle ends, the usual timeout should be resumed.

BElluu:
Closed #920


lib/teiserver/player/session.ex
```diff
@@ -1744,11 +1744,17 @@ defmodule Teiserver.Player.Session do
...
-      Logger.debug("Player timed out, stopping session")
-      {:stop, :normal, state}
-    else
-      {:noreply, state}
+    cond do
+      not is_nil(state.conn_pid) ->
+        {:noreply, state}
+
+      not is_nil(state.battle) ->
+        {:ok, _tref} = :timer.send_after(@connection_timeout, :connection_timeout)
```
geekingfrog: Instead of periodically sending this message, I think it would be better to not send it in the first place when the connection goes down when there is an ongoing battle.
And when the battle ends (or goes down), only then send this message if there is no connection alive at that point.

This would lead to more predictable behaviour. With this polling system, depending on the timing, the session could end right when the battle stops, or up to 30s later.

test/teiserver/player/session_test.exs
```diff
@@ -36,6 +36,49 @@ defmodule Teiserver.Player.SessionTest do
     end
   end
 
+  describe "connection timeout" do
+    setup [:setup_session]
+
+    test "session stops when timed out and not in battle", %{sess_pid: sess_pid} do
+      :sys.replace_state(sess_pid, fn state -> %{state | conn_pid: nil} end)
```
geekingfrog: I am not a fan of `replace_state`, because it breaks the encapsulation of the state. If the session changes how it handle this bit of state, this test will still pass (though most likely because of an unhandled exception crashing the genserver).

I think it a better (but more complex) approach would be instead to simulate the connection with a separate process that you can kill whenever, and use the public API of the module as much as possible.
There is `Session.battle_start` which will do all of that, no need to muck around the internal state.
To spawn processes when you need a pid, you can have a look at the [lobby test](https://github.com/beyond-all-reason/teiserver/blob/main/test/teiserver/tachyon_lobby/lobby_test.exs#L126) where this method is used a lot.

test/teiserver/player/session_test.exs
```diff
@@ -36,6 +36,49 @@ defmodule Teiserver.Player.SessionTest do
...
+
+      send(sess_pid, :connection_timeout)
```
geekingfrog: Also not terribly keen on sending messages from outside the module. If you need to trigger this event, you should instead create another public function like `trigger_connection_timeout` or something like that, with a doc mentioning this should only be used for test, and then call this function from the test.

Once you start sending "internal" messages from other modules, it becomes a lot trickier to make change to these messages, whereas it's generally easier to track call sites for the functions emitting these messages.

test/teiserver/player/session_test.exs
```diff
@@ -36,6 +36,49 @@ defmodule Teiserver.Player.SessionTest do
...
+
+      Polling.poll_until(fn -> Process.alive?(sess_pid) end, fn alive -> not alive end)
```
geekingfrog: I think polling is fine there. But when you have a pid, you can also monitor and wait for the DOWN message like so:

```elixir
ref = Process.monitor(sess_pid)
# do something
assert_receive {:DOWN, ^ref, _, _, _}
```

which is snappier than the polling

test/teiserver/player/session_test.exs
```diff
@@ -13,14 +15,49 @@ defmodule Teiserver.Player.SessionTest do
...
   def setup_config(_context) do
     TachyonLib.enable_state_restoration()
     Callbacks.on_exit(fn -> TachyonLib.disable_state_restoration() end)
   end
 
+  defp disconnect(fake_conn, user_id) do
+    conn_ref = Process.monitor(fake_conn)
+    Process.exit(fake_conn, :kill)
+    assert_receive {:DOWN, ^conn_ref, :process, _, _}
+    Session.conn_state(user_id)
```
geekingfrog: you don't seem to do anything with that and can be removed

test/teiserver/player/session_test.exs
```diff
@@ -13,14 +15,49 @@ defmodule Teiserver.Player.SessionTest do
...
+
+    Session.battle_start(user_id, {"battle-id", fake_battle}, %{
+      ips: ["127.0.0.1"],
+      port: 1234,
+      engine: %{version: "v1"},
+      game: %{springName: "game"},
+      map: %{springName: "map"}
+    })
+
+    Session.conn_state(user_id)
```
geekingfrog: likewise, you don't need this I believe

test/teiserver/player/session_test.exs
```diff
@@ -36,6 +73,52 @@ defmodule Teiserver.Player.SessionTest do
...
+      disconnect(fake_conn, user.id)
+      Session.trigger_connection_timeout(sess_pid)
+      assert_receive {:DOWN, ^ref, :process, _, _}
+    end
+
+    test "session survives when player disconnects during battle",
+         %{sess_pid: sess_pid, user: user, fake_conn: fake_conn} do
+      _fake_battle = start_battle(sess_pid, user.id)
+      disconnect(fake_conn, user.id)
+      assert Process.alive?(sess_pid)
```
geekingfrog: this test passes even without any of your changes.
This lead me to discover that there is a subtle problem in the disconnection logic. Consider the following sequence of event:

* player connects
* connection drops -> timeout timer starts
* player reconnects on the same session
* player joins a battle
* connection drops again
* `:timeout` event arrive -> session is dropped

It is a bit far fetch, but the fix is simple: `if is_nil(state.conn_pid) && is_nil(state.battle) do` when checking if the session should stop.

With this you can amend this test and trigger the connection timeout then `refute_receive {:DOWN...` on the session pid.

test/teiserver/player/session_test.exs
```diff
@@ -36,6 +73,52 @@ defmodule Teiserver.Player.SessionTest do
...
+    test "session stops after battle ends and player is still disconnected",
+         %{sess_pid: sess_pid, user: user, fake_conn: fake_conn} do
+      fake_battle = start_battle(sess_pid, user.id)
+      disconnect(fake_conn, user.id)
+
+      sess_ref = Process.monitor(sess_pid)
+      battle_ref = Process.monitor(fake_battle)
+      Process.exit(fake_battle, :kill)
+      assert_receive {:DOWN, ^battle_ref, :process, _, _}
+      Session.conn_state(user.id)
```
geekingfrog: Is this call there to ensure the `DOWN` message was processed by the session? If yes this should instead be polled repeatedly until you get `:disconnected`, otherwise it's going to lead to races.

test/teiserver/player/session_test.exs
```diff
@@ -36,6 +73,52 @@ defmodule Teiserver.Player.SessionTest do
...
+    test "session survives when player reconnects before battle ends",
+         %{sess_pid: sess_pid, user: user, fake_conn: fake_conn} do
+      _fake_battle = start_battle(sess_pid, user.id)
+      disconnect(fake_conn, user.id)
+
+      {:ok, new_conn} = Task.start(:timer, :sleep, [:infinity])
+      Session.replace_connection(sess_pid, new_conn)
+
+      assert Process.alive?(sess_pid)
+      assert :connected == Session.conn_state(user.id)
```
geekingfrog, line 124: similar here, I think there should also be a `trigger_connection_timeout` followed by a `refute_receive` to make sure old timeout events aren't processed in this case.

geekingfrog (commented):
Nearly there. This turned out to be a bit more tricky than I expected, there can be subtle races. I think with a slight change I suggested in the comment it should be good to go.

BElluu:
Thank you for review. I'll check it and make corrections after work :)

geekingfrog (approved):

PR #1145, opened 2026-05-09
Added some typo checks around the all-welcome title check

Teifion:
Addresses a loophole in the name check which can be exploited by spelling mistakes. We won't be trying to stop all of them but this should stop some of the more obvious ones.

### DIFF
--- a/lib/teiserver/lobby/libs/lobby_restrictions.ex
+++ b/lib/teiserver/lobby/libs/lobby_restrictions.ex
@@ -347,19 +347,19 @@ defmodule Teiserver.Lobby.LobbyRestrictions do
 
   defp allwelcome_title?(nil), do: false
 
-  defp allwelcome_title?(name) do
-    name
-    |> String.downcase()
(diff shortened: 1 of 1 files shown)


lib/teiserver/lobby/libs/lobby_restrictions.ex
```diff
@@ -347,19 +347,19 @@ defmodule Teiserver.Lobby.LobbyRestrictions do
...
+    title = leet_replace(title)
+
+    # Matches against all welcome and a couple of typo'd versions
+    Regex.match?(~r/all?\s?well?come/, title)
   end
 
   @spec pro_title?(String.t()) :: boolean()
-  def pro_title?(name) do
-    name = String.replace(name, "0", "o")
+  def pro_title?(title) do
```
Teifion, line 358: Renamed from `name` to `title` for consistency across the module.

L-e-x-o-n (approved):

PR #1151, opened 2026-05-12
When blocked by a boss, cannot join lobby

Teifion:
If the lobby has bosses then performs an additional block check, if any of the bosses are blocking you then you are unable to join the lobby (as they can kick without a vote).


lib/teiserver/account/libs/auth_lib.ex
```diff
@@ -100,13 +100,14 @@ defmodule Teiserver.Account.AuthLib do
   end
 
   def allow?(permissions_held, permission_required) do
+    permissions_held = List.wrap(permissions_held)
```
Teifion, line 103: We use `Enum.member` for the test so this should in theory be a list. LLMs flag it as a concern though I do not believe it's actually an attack vector.

L-e-x-o-n:
This used to be a thing but was intentionally removed. It was working initially for the first boss but once people started to appoint a new one it caused issues. When appointing someone it could lead to a potentially large number of people being blocked from entering the lobby which might not be the intention of those appointing the new boss. Bosses are often be appointed to help with lobby management, not to have full control over who can enter and I don't really see an easy way to address this. A manual kickban on the players can still be used.

Teifion:
Thank you Lexon, I will close the PR. I agree with your points and can't think of a solution either.

PR #1154, opened 2026-05-13
Add `data` field to CacheUser as first stage of removal

Teifion:
This is a step towards removing the CacheUser, putting a data field into it as a mirror of the User. Given the cache sizes being so small I do not anticipate memory issues while we transition.

This PR changes over `print_client_messages` and `print_server_messages` to the new access method.

Edit: Thinking about an alternate approach.


Teifion:
Closing as Lexon already working on the issue.

PR #1155, opened 2026-05-13
Implemented a login queue for Tachyon protocol

### LINKED ISSUE #922: [Feature]: Implement a login queue for tachyon
### Describe the Feature

Having the ability to limit the number of active player saved our bacon several time so we want the same feature but for tachyon. Similar as to how there is a `system.User limit` site setting, there should be one for tachyon.

### Expected Behavior

After authenticating the user:
1) if it's an autohost: no limit apply, they can always connect
2) if it's a player and the number of connected player is ≥ system limit, then wait in a queue. When the number of connected player drops below the limit (or the limit is increased), the player should be considered "online".
When that happen, send the `user/self` event.

While the user is in the login queue, no event should be sent, and any request from the client should be rejected with `invalid_request`

### Additional context

We could have a different queue pre-authentication, but that can come later.
We will likely also expand the protocol a bit to add some events to let the player know they are in a login queue, and how many are in front (roughly).

The first implementation of the queue can be a simple, centralized process. We can create partition later on (probably before we go fully live)

BElluu:
Closed #922 

**AI Usage**
I used AI (Claude Code - Sonnet 4.6) to describe how login queue works for Spring protocol and which files implements that. I got advise where I should handle queue.


lib/teiserver/tachyon/login_queue.ex
```diff
@@ -0,0 +1,142 @@
...
+  end
+
+  @spec set_limit(non_neg_integer()) :: :ok
+  def set_limit(limit) do
+    GenServer.call(__MODULE__, {:set_limit, limit})
+  end
+
+  @impl GenServer
+  def init(_args) do
+    Logger.metadata(actor_id: "Tachyon.LoginQueue")
```
geekingfrog: generally, at least in the tachyon stuff, the actor id is an atom, `:login_queue` should do.

lib/teiserver/tachyon/login_queue.ex
```diff
@@ -0,0 +1,142 @@
...
+    {:reply, :queue.len(state.queue), state}
+  end
+
+  def handle_call({:attempt_login, pid, user_id}, _from, state) do
+    Process.monitor(pid)
+    capacity = available_capacity(state)
+
+    {new_state, admitted?} =
+      if capacity > 0 && :queue.is_empty(state.queue) do
+        {put_in(state, [:monitors, pid], :admitted), true}
```
geekingfrog: why do you need to monitors admitted players?

lib/teiserver/tachyon/login_queue.ex
```diff
@@ -0,0 +1,142 @@
...
+        state
+
+      {{:value, member}, rest} ->
+        new_state = Map.replace!(state, :queue, rest)
+
+        case Map.fetch(new_state.monitors, member.pid) do
+          {:ok, :waiting} ->
+            # Promote to admitted — no new monitor needed, we already have one
+            new_state = put_in(new_state, [:monitors, member.pid], :admitted)
+            send(member.pid, {:login_accepted, member.user_id})
```
geekingfrog: Don't send messages like that, this should be a function call from the relevant module and the actual message is an implementation details.
I already gave the same feedback on [your other PR](https://github.com/beyond-all-reason/teiserver/pull/1142#discussion_r3211337415)

lib/teiserver/tachyon/login_queue.ex
geekingfrog, line 1: I am not a fan of fully relying on `:DOWN` messages to manage the queue.
The problem is that if there is a lot of churn, it'll generate a lot of messages, and all of them are only changing the queue by 1, with also some racy behaviour due to the eventual consistency of the registry.

The way the spring queue works seems simpler:
* have a (configurable) timer
* on each tick, check available capacity and dequeue `n` users at once
* when a queued player disconnect, just remove it from the queue, but otherwise just wait for the next tick to enqueue players.

lib/teiserver/player/tachyon_handler.ex
```diff
@@ -24,7 +25,8 @@ defmodule Teiserver.Player.TachyonHandler do
 
   @type state :: %{
           user: T.user(),
-          sess_monitor: reference(),
+          status: :waiting | :admitted,
+          sess_monitor: reference() | nil,
```
geekingfrog: I prefer having the state be a union type. Because effectively, when `status = :admitted`, then the session is there, but this type does allow such thing. Whereas a union like `%{user: T.user(), status: :waiting} | %{user: T.user(), status: :admitted, sess_monitor: ref()}` is more explicit

lib/teiserver/tachyon/system.ex
```diff
@@ -28,6 +29,7 @@ defmodule Teiserver.Tachyon.System do
       Matchmaking.System,
       TachyonLobby.System,
       Party.System,
+      LoginQueue,
```
geekingfrog: I think the login queue should belong in the `Player.System` tree since it's directly relevant to the tachyon handler, the registry and the session, and not much else.

test/teiserver/player/login_queue_test.exs
geekingfrog, line 1: I'll have a look at the tests once the queue is reworked to use ticks.

BElluu:
Thank you for such a comprehensive review. I'll respond to comments and make corrections tomorrow.

lib/teiserver/tachyon/login_queue.ex
```diff
@@ -0,0 +1,142 @@
...
+  def handle_call(:queue_size, _from, state) do
+    {:reply, :queue.len(state.queue), state}
+  end
+
+  def handle_call({:attempt_login, pid, user_id}, _from, state) do
+    Process.monitor(pid)
+    capacity = available_capacity(state)
+
+    {new_state, admitted?} =
+      if capacity > 0 && :queue.is_empty(state.queue) do
```
geekingfrog: If a session already exists the player should bypass the login queue since they are reconnecting

Teifion:
Question, why not re-use or extend `LoginThrottleServer`? If we bring this in we will now have two different login queues meaning rate limiting becomes that much harder.

geekingfrog:
Reusing the `LoginThrottleServer` is also an option, though you would need to change how to get the capacity and also how to communicate back to the waiting process that they are now logged in.
This thing is simple enough that I think it's fine to have 2 slightly different implementations. And because tachyon and spring are incompatible and using different game servers, I don't think having a shared queue makes much sense. We can also have a queue for tachyon for testing in parallel with the spring queue, say, for a limited beta or something like that.

Teifion:
The reason for the login queue for spring is to prevent the thundering herd problem and creating too many messages too fast. Unless we are going to turn off spring at the same time as Tachyon we will need to have the spring login throttle be aware of the tachyon one and the tachyon one use the same limits as the spring one.

geekingfrog:
I don't think we'll have this problem. I'm not expecting to have much dual use of the 2 protocols since it would split the player base. I'd like to keep the 2 systems separate as much as possible. If we run into problems, we can add the player count of the other system to each queue, it would be a trivial change.

Teifion:
Given the nature of such a policy decision I've posted to [discord](https://discord.com/channels/549281623154229250/943582636679520256/1505604956340752616) for those who might want to weigh in on it but would not otherwise see this PR and to prevent us derailing the PR discussion with this topic.

geekingfrog:
Clarified on discord: tachyon is completely separate from spring, so it's fine to have a different login queue process.

lib/teiserver/player/login_queue.ex
```diff
@@ -0,0 +1,156 @@
...
+
+  @spec set_tick_period(non_neg_integer() | :infinity) :: :ok
+  def set_tick_period(new_period) do
+    GenServer.cast(__MODULE__, {:set_tick_period, new_period})
+  end
+
+  @doc false
+  def tick do
+    send(__MODULE__, :tick)
+    :ok
```
geekingfrog: you don't need the `:ok`, `send` already returns that. Also, if you're using that for test purpose, there should at least be a comment

lib/teiserver/player/login_queue.ex
```diff
@@ -0,0 +1,156 @@
...
+  alias Teiserver.Config
+  alias Teiserver.Data.Types, as: T
+  alias Teiserver.Player
+  alias Teiserver.Player.TachyonHandler
+
+  use GenServer
+
+  require Logger
+
+  @config_key "system.User limit"
```
geekingfrog: Let's put that under a different config key under the tachyon config. There are already things [like these](https://github.com/geekingfrog/teiserver/blob/19e919c1235b05ab9dad28d6905b8a7575bb8d9e/lib/teiserver/party.ex#L98-L116)

lib/teiserver/player/tachyon_handler.ex
```diff
@@ -22,11 +24,18 @@ defmodule Teiserver.Player.TachyonHandler do
...
-  @type state :: %{
-          user: T.user(),
-          sess_monitor: reference(),
-          pending_responses: Handler.pending_responses()
-        }
+  @type state ::
+          %{
+            user: T.user(),
+            status: :waiting,
+            pending_responses: Handler.pending_responses()
```
geekingfrog: how can there be pending responses when the user isn't connected yet?

lib/teiserver/player/tachyon_handler.ex
```diff
@@ -51,19 +60,43 @@ defmodule Teiserver.Player.TachyonHandler do
...
-    user = initial_state.user
-    Logger.metadata(actor_type: :connection, actor_id: to_string(user.id))
+    event = build_user_self_event(state.user, sess_state)
+    {:event, "user/self", event, new_state}
+  end
 
-    event = build_user_self_event(user, sess_state)
-    {:event, "user/self", event, state}
+  @doc false
+  def notify_login_accepted(pid, user_id) when is_pid(pid) do
```
geekingfrog: you don't need the `user_id` there. If you have the pid then it's enough, the target process (the handler) has the user in its state.
Also, don't just put `@doc false` especially since it's part of the public API of the module.

lib/teiserver/player/tachyon_handler.ex
```diff
@@ -78,6 +111,10 @@ defmodule Teiserver.Player.TachyonHandler do
   end
 
   @impl Handler
+  def handle_info({:login_accepted, _user_id}, state) do
```
geekingfrog: the user id being ignored is a telltale sign that it's useless.

test/teiserver/player/login_queue_test.exs
```diff
@@ -0,0 +1,154 @@
...
+  use Teiserver.DataCase, async: false
+
+  import Teiserver.TeiserverTestLib, only: [new_user: 0]
+
+  @moduletag :tachyon
+
+  setup _context do
+    :ok = Supervisor.terminate_child(Teiserver.Player.System, LoginQueue)
+    {:ok, _pid} = Supervisor.restart_child(Teiserver.Player.System, LoginQueue)
+    LoginQueue.set_tick_period(:infinity)
```
geekingfrog, line 12: there is already a `Tachyon.System.restart` that is preferred since it'll take care of also restarting the registry used to get the count of connected players.

lib/teiserver/player/login_queue.ex
```diff
@@ -0,0 +1,156 @@
...
+          TachyonHandler.notify_login_accepted(member.pid, member.user_id)
+          dequeue_members(n - 1, new_state)
+        else
+          dequeue_members(n, new_state)
+        end
+    end
+  end
+
+  defp available_capacity(%{total_limit: limit}) do
+    limit - Player.Registry.connected_count()
```
geekingfrog: You should look at the number of session, not the number of connection. Since you are looking at sessions for the `do_admit` logic.
We could make an argument to have a different limit for connection, but for now, we don't need such complexity.

test/teiserver/player/login_queue_test.exs
```diff
@@ -0,0 +1,154 @@
...
+  end
+
+  test "tick on empty queue does nothing" do
+    set_capacity(1)
+    LoginQueue.tick()
+    assert LoginQueue.get_queue_length() == 0
+  end
+
+  defp set_capacity(n) do
+    LoginQueue.set_limit(Player.Registry.connected_count() + n)
```
geekingfrog: adding `n` to the connected count is fishy, why is that?

test/teiserver/player/login_queue_test.exs
```diff
@@ -0,0 +1,154 @@
...
+  end
+
+  defp fake_conn do
+    parent = self()
+    spawn(fn -> fake_conn_loop(parent) end)
```
geekingfrog: Don't use spawn, it creates completely unsupervised processes. `spawn_link` is better. If you need to kill the process, you can [unlink](https://hexdocs.pm/elixir/main/Process.html#unlink/1).
Alternatively, you can also have the echo loop and manually call spawn/spawn_link in the test

lib/teiserver/player/login_queue.ex
```diff
@@ -0,0 +1,156 @@
...
+
+  @config_key "system.User limit"
+  @default_tick_period 1_000
+
+  def start_link(opts) do
+    GenServer.start_link(__MODULE__, opts, name: __MODULE__)
+  end
+
+  @spec attempt_login(pid(), T.userid()) :: boolean()
+  def attempt_login(pid, user_id) do
```
geekingfrog: This is more of a nitpick, but the `pid` is kind of an implementation detail. I tend to allow an override but otherwise just put a default, for example: `def create_party(user_id, pid \\ self()) do`
There should also be a doc mentioning what function will be called when the login is accepted. With event driven system, the control flow can get confusing really quickly.

test/teiserver/player/login_queue_test.exs
geekingfrog, line 1: I see there that you rely on the user id in the `:login_accepted` message. This isn't ideal to have code that only makes sense for test, this really should be kept to an absolute minimum.

What you can do instead is to have "tagged echo" processes. So the child process looping over `receive` are spawned with a `tag`, and they send to the parent {tag, msg}. This way you can still keep track of which process get which message without the need for the additional parameter.

test/teiserver/player/login_queue_test.exs
```diff
@@ -130,24 +123,25 @@ defmodule Teiserver.Player.LoginQueueTest do
   end
 
   defp set_capacity(n) do
-    LoginQueue.set_limit(Player.Registry.connected_count() + n)
+    LoginQueue.set_limit(SessionRegistry.count() + n)
```
geekingfrog: why do you offset the capacity by the session count? That seems fishy

BElluu: Yes, there must be some leftover from my previous implementation or test run.

lib/teiserver/player/login_queue.ex
```diff
@@ -0,0 +1,175 @@
...
+  end
+
+  def setup_site_configs do
+    Config.add_site_config_type(%{
+      key: @limit_config_key,
+      section: "Tachyon",
+      type: "integer",
+      permissions: ["Admin"],
+      description: "Maximum number of concurrent Tachyon player sessions",
+      default: 1000
```
geekingfrog: There should be an `update_callback` key there to `set_limit`. Right now the process' limit is set at creation and will never change.

lib/teiserver/player/tachyon_handler.ex
```diff
@@ -265,6 +300,16 @@ defmodule Teiserver.Player.TachyonHandler do
...
           state()
         ) :: WebSock.handle_result()
+  def handle_command(
+        _command_id,
+        _message_type,
+        _message_id,
+        _message,
+        %{status: :waiting} = state
+      ) do
+    {:error_response, :invalid_request, state}
```
geekingfrog: I think it would be helpful to have a short message there, like "waiting for login". Until we figure out if/how we need a better feedback mechanism to the client.

PR #1156, opened 2026-05-13
Moved Privacy Policy and CoC pages to refer to main website

Teifion:
The version of GDPR on the server is out of date, now it is hosted on the main site it makes more sense to direct people there instead.


lib/teiserver_web/controllers/admin/user_controller.ex
```diff
@@ -1170,6 +1170,9 @@ defmodule TeiserverWeb.Admin.UserController do
     page == 0 && Enum.count(users) > 20 && search_term != ""
   end
 
+  @doc """
+  Removes all PIP from a user in accordance with GDPR.
```
L-e-x-o-n: What's PIP? Typo of PII (personally identifiable information) or something else?

Teifion: That's a typo, it should be PII. I'll fix that.

PR #1159, opened 2026-05-14
Add download replay button to match page

ChocoboCoding:
This PR adds a download replay button to the match page.

This will need testing on the integration server, as I can not properly test it locally

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -221,6 +221,11 @@ config :logger, :notice_log, path: "#{log_root_path}/notice.log"
 
 config :logger, :info_log, path: "#{log_root_path}/info.log"
 
+config :teiserver, TeiserverWeb,
+  api_url: "https://api.bar-rts.com/replays/",
+  storage_url:
+    "https://storage.uk.cloud.ovh.net/v1/AUTH_10286efc0d334efd917d476d7183232e/BAR/demos/"
+
 enable_discord_bridge =
   Teiserver.ConfigHelpers.get_env("TEI_ENABLE_DISCORD_BRIDGE", false, :bool)
 
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -222,8 +222,16 @@ config :logger, :notice_log, path: "#{log_root_path}/notice.log"
 config :logger, :info_log, path: "#{log_root_path}/info.log"
 
 config :teiserver, TeiserverWeb,
-  api_url: Teiserver.ConfigHelpers.get_env("BAR_LIVE_SERVICES_API_URL"),
-  storage_url: Teiserver.ConfigHelpers.get_env("BAR_REPLAY_STORAGE_URL")
+  api_url:
+    Teiserver.ConfigHelpers.get_env(
+      "BAR_LIVE_SERVICES_API_URL",
+      "https://api.bar-rts.com/replays/"
+    ),
+  storage_url:
+    Teiserver.ConfigHelpers.get_env(
+      "BAR_REPLAY_STORAGE_URL",
+      "https://storage.uk.cloud.ovh.net/v1/AUTH_10286efc0d334efd917d476d7183232e/BAR/demos/"
+    )
 
 enable_discord_bridge =
   Teiserver.ConfigHelpers.get_env("TEI_ENABLE_DISCORD_BRIDGE", false, :bool)
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -222,9 +222,8 @@ config :logger, :notice_log, path: "#{log_root_path}/notice.log"
 config :logger, :info_log, path: "#{log_root_path}/info.log"
 
 config :teiserver, TeiserverWeb,
-  api_url: "https://api.bar-rts.com/replays/",
-  storage_url:
-    "https://storage.uk.cloud.ovh.net/v1/AUTH_10286efc0d334efd917d476d7183232e/BAR/demos/"
+  api_url: Teiserver.ConfigHelpers.get_env("BAR_LIVE_SERVICES_API_URL"),
+  storage_url: Teiserver.ConfigHelpers.get_env("BAR_REPLAY_STORAGE_URL")
 
 enable_discord_bridge =
   Teiserver.ConfigHelpers.get_env("TEI_ENABLE_DISCORD_BRIDGE", false, :bool)
--- a/lib/teiserver_web/live/battles/match/match_components.ex
+++ b/lib/teiserver_web/live/battles/match/match_components.ex
@@ -12,7 +12,26 @@ defmodule TeiserverWeb.Battle.MatchComponents do
   attr :match_id, :integer, default: nil
   attr :replay, :string, default: nil
 
+  defp build_download_link(nil), do: nil
+
+  defp build_download_link(match_id) do
+    url = "https://api.bar-rts.com/replays/" <> match_id
+
+    json =
+      HTTPoison.get!(url).body
+      |> Jason.decode!()
+
+    filename =
+      json["fileName"]
+      |> String.replace(" ", "%20")
+
+    "https://storage.uk.cloud.ovh.net/v1/AUTH_10286efc0d334efd917d476d7183232e/BAR/demos/" <>
+      filename
+  end
+
   def section_menu(assigns) do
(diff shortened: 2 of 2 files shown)


ChocoboCoding:
As you wanted a mention, @StanczakDominik

lib/teiserver_web/live/battles/match/match_components.ex
```diff
@@ -3,6 +3,23 @@ defmodule TeiserverWeb.Battle.MatchComponents do
...
+
+    json =
+      HTTPoison.get!(url).body
+      |> Jason.decode!()
+
+    filename =
+      json["fileName"]
+      |> String.replace(" ", "%20")
+
+    "https://storage.uk.cloud.ovh.net/v1/AUTH_10286efc0d334efd917d476d7183232e/BAR/demos/" <>
```
L-e-x-o-n: Where is this from? Will it change in the future?

lib/teiserver_web/live/battles/match/match_components.ex
```diff
@@ -3,6 +3,23 @@ defmodule TeiserverWeb.Battle.MatchComponents do
   use TeiserverWeb, :component
   import TeiserverWeb.NavComponents, only: [section_menu_button: 1]
 
+  defp build_download_link(nil), do: nil
+
+  defp build_download_link(match_id) do
+    url = "https://api.bar-rts.com/replays/" <> match_id
```
L-e-x-o-n: This should be read from configuration.

lib/teiserver_web/live/battles/match/match_components.ex
```diff
@@ -62,6 +81,18 @@ defmodule TeiserverWeb.Battle.MatchComponents do
       </.section_menu_button>
     <% end %>
 
+    <div class="float-end">
+      <.section_menu_button
+        :if={@download_link != nil}
+        bsname={@view_colour}
+        icon={StylingHelper.icon(:replay)}
```
L-e-x-o-n: The same icon is used for the replay link button so it should be changed.

lib/teiserver_web/live/battles/match/match_components.ex
ChocoboCoding: It is the server where we store the replay files on. I don't think it will change but can also be read from configuration

ChocoboCoding: Addressed in https://github.com/beyond-all-reason/teiserver/pull/1159/commits/37afa4a1f2e4105f600df2efc64d953d28302b6c

lib/teiserver_web/live/battles/match/match_components.ex
ChocoboCoding: Addressed in https://github.com/beyond-all-reason/teiserver/pull/1159/commits/37afa4a1f2e4105f600df2efc64d953d28302b6c

lib/teiserver_web/live/battles/match/match_components.ex
ChocoboCoding: Addressed in https://github.com/beyond-all-reason/teiserver/pull/1159/commits/37afa4a1f2e4105f600df2efc64d953d28302b6c

config/runtime.exs
```diff
@@ -221,6 +221,11 @@ config :logger, :notice_log, path: "#{log_root_path}/notice.log"
 
 config :logger, :info_log, path: "#{log_root_path}/info.log"
 
+config :teiserver, TeiserverWeb,
+  api_url: "https://api.bar-rts.com/replays/",
+  storage_url:
+    "https://storage.uk.cloud.ovh.net/v1/AUTH_10286efc0d334efd917d476d7183232e/BAR/demos/"
```
L-e-x-o-n: Should be read from environment variables like the majority of configs here, it would allow changing the config without the need to recompile.
To make it work you would also need to set the env variables in [ansible teiserver](https://github.com/beyond-all-reason/ansible-teiserver).

L-e-x-o-n (commented):
On second look, I have some additional concerns.

I don't expect it to actually work, match id is internal to teiserver while replay server uses a different format (game id).
Not a fan of extracting the filename like this either, the moment something on that site changes this can stop working, what happens then or if building the link fails?

Honestly, this entire thing feels like a lot of work to save just one click.

ChocoboCoding:
> On second look, I have some additional concerns.
> 
> I don't expect it to actually work, match id is internal to teiserver while replay server uses a different format (game id). Not a fan of extracting the filename like this either, the moment something on that site changes this can stop working, what happens then or if building the link fails?
> 
> Honestly, this entire thing feels like a lot of work to save just one click.

It should work, as it gets the fileName from our [api](https://github.com/beyond-all-reason/bar-db/blob/master/src/rest-api/routes/replays.ts) `https://api.bar-rts.com`. The API returns a json object including the fileName if you query it with the match id.
The download link is constructed the same way it is on the [replay website](https://github.com/beyond-all-reason/bar-live-services/blob/main/pages/replays/_gameId.vue#L13) using the same [storage url](https://github.com/beyond-all-reason/bar-live-services/blob/main/nuxt.config.ts#L17-L19).
The only only way how this can stop working is if we discontinue our own API and stop using our own storage server, but by that point the replay site needs to change, and we can just construct the download link the same way here as we do there.

Other than that, I would hope that we don't just discontinue our REST api, but rather add a new version and discontinue it at a later point after enough time has passed to migrate things

L-e-x-o-n:
By not expecting it to work I was referring to the use of match id; teiserver match id and replay server id are 2 different things.

This is easily solvable because we also store `gameId` on teiserver which is what the replay server uses.

ChocoboCoding:
> By not expecting it to work I was referring to the use of match id; teiserver match id and replay server id are 2 different things.
> 
> This is easily solvable because we also store `gameId` on teiserver which is what the replay server uses.

Isn't the link to the replay website (and in turn the gameId) just the `match.game_id` in most cases?

https://github.com/beyond-all-reason/teiserver/blob/main/lib/teiserver_web/live/battles/match/show.ex#L239-L250

L-e-x-o-n:
> > By not expecting it to work I was referring to the use of match id; teiserver match id and replay server id are 2 different things.
> > This is easily solvable because we also store `gameId` on teiserver which is what the replay server uses.
> 
> Isn't the link to the replay website (and in turn the gameId) just the `match.game_id` in most cases?
> 
> https://github.com/beyond-all-reason/teiserver/blob/main/lib/teiserver_web/live/battles/match/show.ex#L239-L250

Yes, that's exactly what I was referring to.

config/runtime.exs
ChocoboCoding: Resolved in https://github.com/beyond-all-reason/teiserver/pull/1159/commits/1655a06add1e3554fcad66af0748e232e71e7853

ChocoboCoding: Ansible PR https://github.com/beyond-all-reason/ansible-teiserver/pull/27

L-e-x-o-n (commented):
You are still using `match_id` instead of `game_id`.

`get!` and `decode!` will raise an exception if they fail for any reason, use `get` and `decode` and handle the error instead. Don't show the button in case of error.

ChocoboCoding:
Is the missing ENV error something I should fix somehow for example by setting a default, or will that be done with the repo vars or some other way?

L-e-x-o-n:
> Is the missing ENV error something I should fix somehow for example by setting a default, or will that be done with the repo vars or some other way?

Yes it should be fixed by setting the default.

lib/teiserver_web/live/battles/match/match_components.ex
```diff
@@ -1,8 +1,26 @@
...
+  defp build_download_link(nil), do: nil
+
+  defp build_download_link(game_id) do
+    url = Application.get_env(:teiserver, :replay)[:api_url] <> game_id
+
+    with {:ok, response} <- HTTPoison.get(url).body,
+         {:ok, json} <- Jason.decode(response) do
+      filename =
+        json["fileName"]
+        |> String.replace(" ", "%20")
```
geekingfrog, line 16: this is incomplete. If you want to build a link, you should use the [URI](https://hexdocs.pm/elixir/URI.html) module, construct the url, and turn it into a string. It'll take care of proper percent encoding and other stuff like that.

PR #1163, opened 2026-05-16
Defense in depth preventing admin.user scope holders from assigning privileged roles

thvl3:
Hardened Teiserver `Web.API.Admin.UserController` so callers holding the `admin.user` OAuth scope can no longer assign privileged roles or arbitrary permission strings: create/2 now filters roles to a non-staff allowlist and recomputes permissions from those roles (ignoring caller-supplied permission strings), and refresh_token/2 returns 403 when the target user has any role outside that allowlist.

Added three regression tests covering both paths

### DIFF
--- a/lib/teiserver_web/controllers/api/admin/user_controller.ex
+++ b/lib/teiserver_web/controllers/api/admin/user_controller.ex
@@ -14,15 +14,9 @@ defmodule TeiserverWeb.API.Admin.UserController do
   # allowlist is silently dropped from create requests and blocks
   # refresh_token requests against pre-existing users.
   @api_allowed_roles ~w(
-    Default
     Verified
     Trusted
     BAR+
-    Armada
-    Cortex
-    Legion
-    Raptor
-    Scavenger
     Bot
   )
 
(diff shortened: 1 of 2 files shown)


lib/teiserver_web/controllers/api/admin/user_controller.ex
```diff
@@ -51,6 +72,34 @@ defmodule TeiserverWeb.API.Admin.UserController do
...
   end
 
+  # Drop any caller-supplied role outside the allowlist and recompute
+  # permissions from the surviving roles. Caller-supplied permission
+  # strings are discarded entirely.
+  defp sanitize_roles_and_permissions(params) do
+    roles =
+      params
+      |> Map.get("roles", [])
+      |> List.wrap()
```
geekingfrog: why is `wrap` needed?

geekingfrog (commented):
the changes looks fine, but I am not fully sold on the reasoning behind.
If someone has access to this admin API, they should already be an admin, so limiting what this API can do seems a bit counterproductive.
Though I don't have a use case right now that would warrant this API to craft admin users.

geekingfrog:
After sleeping on it, I think it's fine. If/when we have a need for this API to create users with arbitrary roles we can revisit that. Just not sure about this `List.wrap` thingy.

lib/teiserver_web/controllers/api/admin/user_controller.ex
thvl3: ah it isnt needed, roles should already pass thru as a list, will fix

PR #1164, opened 2026-05-17
Check oauth scopes with roles

geekingfrog:
Whenever an OAuth scope is involved, either at app creation or when getting tokens, these should be checked against the user's roles.
This prevent someone to get a token granting them admin scope when they would otherwise not be able to do so.

### DIFF
--- a/lib/teiserver/o_auth.ex
+++ b/lib/teiserver/o_auth.ex
@@ -114,7 +114,7 @@ defmodule Teiserver.OAuth do
   The token scopes are the same as the application
   """
   @spec create_code(
-          User.t() | T.userid(),
+          User.t(),
           %{
             id: integer(),
             scopes: Application.scopes(),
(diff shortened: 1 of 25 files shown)


lib/teiserver/o_auth.ex
```diff
@@ -144,7 +139,8 @@ defmodule Teiserver.OAuth do
       # the code for a token
       attrs = %{
         value: :crypto.strong_rand_bytes(32) |> Base.hex_encode32(),
-        owner_id: user_id,
+        owner: user,
+        owner_id: user.id,
```
L-e-x-o-n: Why do we need both owner and owner id, would just owner not be enough?

geekingfrog: the changeset requires owner_id, but yeah, I should probably move that in the changeset logic.

PR #1165, opened 2026-05-17
Add option to require MFA for privilaged accounts

Teifion:
- Adds a report for listing users with MFA guarded roles without MFA enabled
- Setting set from an elixir config value (to prevent it being disabled without observation)
- A user without MFA will not pass any permission checks for `Server` or `Admin` tests

When you have MFA enabled and an MFA guarded role, the security screen now shows that:
image

When MFA is not enabled, on the homepage you will see this warning:
image

MFA usage report (names left out intentionally):
image

### DIFF
--- a/lib/teiserver/account/libs/auth_lib.ex
+++ b/lib/teiserver/account/libs/auth_lib.ex
@@ -75,7 +75,7 @@ defmodule Teiserver.Account.AuthLib do
           map() | Conn.t() | Socket.t() | Teiserver.Account.User.t(),
           String.t() | [String.t()]
         ) :: boolean
-  def allow?(nil, []), do: false
+  def allow?(nil, _any), do: false
   def allow?(_permissions, nil), do: true
   def allow?(_permissions, ""), do: true
   def allow?(_permissions, []), do: true
@@ -199,8 +199,8 @@ defmodule Teiserver.Account.AuthLib do
   Returns true if the user_id in question has an active totp
   """
   @spec has_active_mfa?(Teiserver.Account.User.id()) :: boolean()
-  def has_active_mfa?(userid) do
-    Account.get_user_totp_status(userid) == :active
+  def has_active_mfa?(user_id) do
+    user_id != nil and Account.get_user_totp_status(user_id) == :active
   end
 
   @doc """
@@ -208,16 +208,20 @@ defmodule Teiserver.Account.AuthLib do
   are possessed then also checks for MFA. If no MFA is present then the
   roles are removed.
   """
-  @spec maybe_remove_mfa_roles(Teiserver.Account.User.t()) :: Teiserver.Account.User.t()
-  def maybe_remove_mfa_roles(user) do
-    if contains_mfa_role?(user.roles) do
-      IO.puts("")
-      IO.inspect("CONTAINS", label: "#{__MODULE__}:#{__ENV__.line}")
-      IO.puts("")
+  @spec maybe_remove_mfa_roles(Teiserver.Account.User.id()) :: :removed | :nochange
+  def maybe_remove_mfa_roles(user_id) do
+    user = Account.get_user(user_id)
+    mfa_required = Application.get_env(:teiserver, Teiserver)[:require_mfa_for_privileged_roles]
 
+    if mfa_required and contains_mfa_role?(user.roles) do
+      # If the user has an MFA role then do an empty update
+      # and the update code will decide if how to alter the user
       user
+      |> Account.script_update_user(%{})
+
+      :removed
     else
-      user
+      :nochange
     end
   end
 
--- a/lib/teiserver/account/libs/auth_lib.ex
+++ b/lib/teiserver/account/libs/auth_lib.ex
@@ -1,6 +1,6 @@
 defmodule Teiserver.Account.AuthLib do
   @moduledoc """
-
+  Module with functions for working with authorisation/permission checks
   """
 
   alias Phoenix.LiveView
@@ -12,6 +12,10 @@ defmodule Teiserver.Account.AuthLib do
 
   @spec icon :: String.t()
   def icon, do: "fa-solid fa-address-card"
+
+  def mfa_roles do
+    ~w[Server Admin]s
+  end
 
   @spec get_all_permission_sets() :: list()
   def get_all_permission_sets do
--- a/lib/teiserver/account/libs/auth_lib.ex
+++ b/lib/teiserver/account/libs/auth_lib.ex
@@ -224,16 +224,6 @@ defmodule Teiserver.Account.AuthLib do
     Application.get_env(:teiserver, Teiserver)[:require_mfa_for_privileged_roles]
   end
 
-  @doc """
-  Given a list of roles, remove any MFA guarded roles from the list
-  """
-  @spec remove_mfa_roles_from_list([String.t()]) :: [String.t()]
-  def remove_mfa_roles_from_list(roles) do
-    Enum.reject(roles, fn role ->
-      Enum.member?(mfa_roles(), role)
-    end)
-  end
-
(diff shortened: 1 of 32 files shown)


lib/teiserver_web/templates/account/security/totp.html.heex
```diff
@@ -26,6 +26,11 @@
 
         <div class="flex-grow-1"></div>
 
+        <div :if={@totp_status == :active and @show_mfa_warning} class="alert alert-warning mb-4">
+          <strong>Warning:</strong>
+          Disabling or resetting your MFA will remove any privilaged roles. You will need to have them re-added after your MFA is setup again.
```
ChocoboCoding: I think it's writing "privileged"

geekingfrog:
>     * Setting set from an elixir config value (to prevent it being disabled without observation)

I think that's fine for now. But if we need that, we may want to add an audit table and log changes to selected (or all?) configs.

>     * All editing of a user will remove guarded roles if MFA is not enabled

I am not too enthusiastic about this idea. I find the side effect/punishment odd. If we want to ensure mfa is enabled for some users, I'd rather have this modeled directly on the `teiserver_account_user_totps` table, or derive it from their roles. And perhaps have an option for exempting someone?
So some users won't be able to login at all unless they also pass the mfa check. I find that more clear than changing how some operations work depending on some unrelated property.

L-e-x-o-n:
I too am not a fan of removing roles like that.

I have an alternative proposal, while MFA is disabled for a role that should require it:
- Show warning everywhere to enable MFA
- Disable access to anything requiring the role for which MFA is required

lib/teiserver_web/templates/account/security/totp.html.heex
Teifion: Resolved with https://github.com/beyond-all-reason/teiserver/pull/1165/commits/088fda696ced91f3e320ecda6361c9bf7dae5908

Teifion:
I started by having `allow?` check for mfa for certain roles but it felt like it was circumventing the actual problem of requiring admins etc to have mfa enabled. I am happy to change it to that if you both want it set that way. If you both let me know your thoughts on this I can make the change and re-request the review.

> May need an audit table

It is set in the config.exs, it cannot be changed at runtime so there's no audit table relevance for this one. Though I do agree an audit table would be good in general.

L-e-x-o-n:
I don't see how it circumvents the problem, on the contrary, requiring MFA to access admin features seems like the direct goal of this change. The current PR behaviour is just a bit confusing and inconvenient. 

Alternatively, if changing all the necessary permission checks to work with MFA requires significant effort, this PR could just show a warning and add the report, a separate PR can then implement the actual MFA requirements.

Teifion:
I'm happy to make the permission check validate the presence of MFA. Given it can trigger multiple times per page I think I'd like to either add a flag to the user or a cache. Do you have a preference?

L-e-x-o-n:
> I'm happy to make the permission check validate the presence of MFA. Given it can trigger multiple times per page I think I'd like to either add a flag to the user or a cache. Do you have a preference?

No preference.

Teifion:
@L-e-x-o-n I have altered it to use the cache and do it at the check stage.

lib/teiserver/account/libs/auth_lib.ex
```diff
@@ -184,6 +198,51 @@ defmodule Teiserver.Account.AuthLib do
...
+
+  @doc """
+  Given a list of roles, remove any MFA guarded roles from the list
+  """
+  @spec remove_mfa_roles_from_list([String.t()]) :: [String.t()]
+  def remove_mfa_roles_from_list(roles) do
+    Enum.reject(roles, fn role ->
+      Enum.member?(mfa_roles(), role)
+    end)
+  end
```
L-e-x-o-n: This seems unused.

lib/teiserver/account/reports/mfa_report.ex
```diff
@@ -0,0 +1,67 @@
...
+      ON totps.user_id = users.id
+    WHERE
+      users.roles && $1::varchar[]
+      AND totps.user_id IS NULL -- Remove anybody with a TOTPS entry
+    ORDER BY
+      users.name ASC
+    LIMIT 50;
+    """
+
+    IO.puts(query)
```
L-e-x-o-n: I assume you missed this.

test/teiserver/account/auth_test.exs
```diff
@@ -0,0 +1,34 @@
+defmodule Teiserver.Account.AuthTest do
+  alias Teiserver.Account
+  alias Teiserver.Account.AuthLib
+  alias Teiserver.AccountFixtures
+
+  use Teiserver.DataCase, async: false
+
+  setup %{} do
```
L-e-x-o-n: Minor detail but can you remove the empty map? It would look cleaner and be consistent with other tests.

test/teiserver/account/auth_test.exs
```diff
@@ -0,0 +1,34 @@
...
+
+    on_exit(fn ->
+      Application.put_env(:teiserver, Teiserver, config)
+    end)
+
+    %{user: user}
+  end
+
+  describe "MFA functions" do
+    test "allow when no MFA", %{user: user} do
```
L-e-x-o-n: The test actually checks if access is denied without MFA so the description should be updated.

lib/teiserver/account/reports/mfa_report.ex
```diff
@@ -0,0 +1,67 @@
...
+      account_users AS users
+    LEFT JOIN
+      teiserver_account_user_totps AS totps
+      ON totps.user_id = users.id
+    WHERE
+      users.roles && $1::varchar[]
+      AND totps.user_id IS NULL -- Remove anybody with a TOTPS entry
+    ORDER BY
+      users.name ASC
+    LIMIT 50;
```
L-e-x-o-n: I think the limit can be raised a bit, maybe 200. There generally won't be that many users with elevated roles, eventually the largest groups would be contributors and overwatch, that's currently about ~ 200 users total and the report wouldn't be accessed often for an increased limit to be an issue.

L-e-x-o-n (commented):
A few minor comments but generally looks good.

lib/teiserver/account/libs/auth_lib.ex
Teifion: Yep, addressed in https://github.com/beyond-all-reason/teiserver/pull/1165/commits/92094112c12f547847af0dc3e3c3be2723771e91

lib/teiserver/account/reports/mfa_report.ex
Teifion: Yes, addressed in https://github.com/beyond-all-reason/teiserver/pull/1165/commits/92094112c12f547847af0dc3e3c3be2723771e91

test/teiserver/account/auth_test.exs
Teifion: Yeah, thanks spicy autocomplete for "helping", addressed in https://github.com/beyond-all-reason/teiserver/pull/1165/commits/92094112c12f547847af0dc3e3c3be2723771e91

test/teiserver/account/auth_test.exs
Teifion: Renamed them in https://github.com/beyond-all-reason/teiserver/pull/1165/commits/92094112c12f547847af0dc3e3c3be2723771e91

lib/teiserver/account/reports/mfa_report.ex
Teifion: Limit upped to 200 in https://github.com/beyond-all-reason/teiserver/pull/1165/commits/92094112c12f547847af0dc3e3c3be2723771e91

L-e-x-o-n (approved):

PR #1170, opened 2026-05-20
Oauth userinfo endpoint

### LINKED ISSUE #1166: Expose new userinfo endpoint protected by OAuth2
This is a feature request to support minimal userinfo endpoint, as documented in https://openid.net/specs/openid-connect-core-1_0.html#UserInfo with a minimal set of claims needed for https://dexidp.io/docs/connectors/oauth/.

Per https://datatracker.ietf.org/doc/html/rfc8414 it would be good if `.well-known/oauth-authorization-server`:
- `userinfo_endpoint` field contained URL of the endpoint
- I see we don't set `scopes_supported` either, it might be good to populate

The userinfo endpoint must based on the provided access token return `application/json` JSON with claims set dependend on the selected scopes. The minimal JSON document can be:

```json
{
 "sub": "248289761001"
}
```

where `sub` claim must be an unique user identifier (yes, string), I assume for us it will be just user id.

Depending on the selected scope additional claims can be set. Theoretically userinfo endpoint could *always* return all claims unconditionally, but it is a good practice to make it granular and let user see "This OAuth2 application will get access to see your email address" and so on.

### Scope `profile`

Claim `preferred_username`: set to in-game username

### Scope `email`

Claim `email` and maybe `email_verified` exactly as defined in https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims

### Scope `groups`

A not exactly standard defined, but many providers do and we need it, `groups` claim, that contains a JSON list with the groups user is member of.

## Full example

Assuming that we requested access token for all 3 scopes: `profile`, `email`, `groups`, example response to userinfo endpoint can look like:

```json
{
 "sub": "248289761001",
 "preferred_username": "Bob132",
 "email": "test@example.com",
 "email_verified": true,
 "groups": ["overwatch", "moderator"]
}
```

geekingfrog:
https://github.com/beyond-all-reason/teiserver/issues/1166

Add an endpoint to get some user info through a JSON API. The info are based on the scopes tied to the bearer token.

### DIFF
--- a/lib/teiserver/o_auth/libs/scope_lib.ex
+++ b/lib/teiserver/o_auth/libs/scope_lib.ex
@@ -9,7 +9,19 @@ defmodule Teiserver.OAuth.Libs.ScopeLib do
 
   @spec allowed_scopes() :: [String.t()]
   def allowed_scopes do
-    ["tachyon.lobby", "admin.map", "admin.engine", "admin.user"]
+    [
+      # this is for blobby/new lobby
+      "tachyon.lobby",
+      # some scopes for bar specific administration and testing
+      "admin.map",
+      "admin.engine",
+      "admin.user",
+      # OpenID connect standard scopes (https://openid.net/specs/openid-connect-core-1_0.html#ScopeClaims)
+      "profile",
+      "email",
+      # that one isn't truly standard but is widely used in practice
+      "groups"
+    ]
   end
 
   @spec all_scopes_allowed?(Account.User.t() | Bot.t(), [String.t()]) ::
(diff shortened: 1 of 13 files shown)


L-e-x-o-n (approved):

p2004a (commented):
That's not true, it's configurable via `userIDKey` field per https://dexidp.io/docs/connectors/oauth/.

Eh, that was supposed to be to https://github.com/beyond-all-reason/teiserver/pull/1170/changes/716c1faee49beb6c6293bc334e704bb39200b767 commit

PR #1178, opened 2026-05-22
Refactored GDPR clean into GDPR forget

Teifion:
A recent audit found our implementation of the GDPR right to be forgotten has some areas we could improve such as removing relationships.

GDPR clean is now renamed into GDPR forget to better mirror the actual function and is placed into a separate task though still written to use the `Conn` struct.

It is advised that after this change is deployed you also run the following code to re-apply this updated function to all previously forgotten users.

```elixir
alias Ecto.UUID
alias Plug.Conn
alias Teiserver.Account
alias Teiserver.Account.User
alias Teiserver.Account.Tasks.GdprForgetTask

admin_id = xyz
conn = %Conn{assigns: %{current_user: %User{id: admin_id, roles: ["Server"]}}, remote_ip: {127, 0, 0, 1}}

# List all users with a dash in them, most users do not have dashes in their
# name so we can probably filter things down a little
(
  Account.list_users(where: [name_like: "-"], limit: :infinity)
  |> Enum.filter(fn user ->
    valid_name? =
      case UUID.cast(user.name) do
        {:ok, _valid} -> true
        _not_valid -> false
      end

    [email_num | _rest] = String.split(user.email, "@")

    valid_email? =
      user.email == "#{email_num}@#{email_num}" or
        user.email == "#{email_num}@#{email_num}.#{email_num}"

    # By checking for the role we can re-start it and not have to
    # re-forget users we have forgotten with the updated
    # function
    valid_name? and valid_email? and not Enum.member?(user.roles, "GDPR forgotten")
  end)
  |> Enum.map(fn user ->
    IO.puts "Re-forgetting #{user.id}"
    GdprForgetTask.forget_user(conn, user)
    
    # The transaction results in deleting from several tables so to
    # give things a chance to calm down we pause for a second
    :timer.sleep(1000)
  end)
  |> Enum.count()
)
```


lib/teiserver_web/templates/component/icon.html.heex
```diff
@@ -1,6 +1,6 @@
 <% icon_class =
   if assigns[:icon_class] do
-    assigns[:icon]
+    assigns[:icon] || ""
```
Teifion: This came up because after forgetting the user trying to view the profile to verify it worked hit an error.

lib/teiserver/account/tasks/gdpr_forget_task.ex
```diff
@@ -0,0 +1,143 @@
+defmodule Teiserver.Account.Tasks.GdprForgetTask do
```
Teifion, line 1: I have tried to ensure this is as documented as possible for a potential future review/audit.

L-e-x-o-n:
Looks good, my only concern is about the way the special role was added here, I think it should be added to `Teiserver.Account.RoleLib` so we have a better overview of which roles exist.

L-e-x-o-n (approved):

geekingfrog:
I'm not a fan of marking the operation through a role. As lexon mentioned, it has some side effects with roles. I think it would be better to have a soft delete `deleted_at`, or a separate table with log events related to gdpr and link that to the anonymised user id.

PR #1184, opened 2026-05-24
Migrate to OTP 27

### LINKED ISSUE #1073: Upgrade to otp 27
* Check compilation warnings/errors when upgrading to otp27. If required, can upgrade the dependencies, but do check changelogs/test that it doesn't break anything.
* need to update `mise.toml` as well as the Dockerfile. For the dockerfile, check the correct builder image, we can't change the debian image from docker since it's ran bare metal.

BElluu:
Closed #1073


NortySpock:
@BElluu  There needs to be an equivalent PR in Ansible-Teiserver to upgrade the OTP version there, so that Ansible deploys an image with the expected  OTP version. I would recommend you leave comments both on this PR and the other PR, noting that they depend on each other and need to be deployed together.

I was also asked, when I updated the Elixir version, what parts I had tested on, so that Lexon / BlackMelon know what areas to watch out for.

See previous PR https://github.com/beyond-all-reason/ansible-teiserver/pull/18

I note that attempting to pull the podman container appears to succeed, so at least you did better than me the first time! 

```
$ podman pull docker.io/hexpm/elixir:1.19.4-erlang-27.3.4-debian-trixie-20251208-slim

Trying to pull docker.io/hexpm/elixir:1.19.4-erlang-27.3.4-debian-trixie-20251208-slim...
Getting image source signatures
Copying blob a48838287428 skipped: already exists  
Copying blob 1733a4cd5954 skipped: already exists  
Copying blob 8e8b8f0b4783 done   | 
Copying blob 71c4f29abe91 done   | 
Copying config 59fe483c33 done   | 
Writing manifest to image destination
59fe483c3368ffddb248e4b973a1e271a0dbac3619b2bcf925fef04cd98d14cc

```

BElluu:
@NortySpock I think we don't need change anything in ansible-teiserver repo, because mechanism was changed in https://github.com/beyond-all-reason/ansible-teiserver/pull/19  - information about image is fetching from teiserver repo.

NortySpock:
@BElluu Ah, yes, that sounds correct. Good catch. Carry on!

geekingfrog (commented):
Did you check the [release notes](https://erlang.org/download/otp_src_27.0.readme) and checked none of the noted potential incompatibilities could catch us?

BElluu:
After reviewing the OTP 27 changelog, I do not see critical breaking changes that would block the migration.
One opportunity worth talking: `config/config.exs` contains a legacy TLS cipher list with 38 entries, most of which are CBC-baed or use weak primities (3DES, SHA1 MAC, static ECDH). OTP 27 tightens its stance on these - CBC ciphers are now explicitly excluded from the default TLS 1.2 cipher list (OTP-19025), and `DES-CBC3-SHA` in particular carries a high risk of failing on systems running OpenSSL 3.x. Since `runtime.exs` overrides the `:https` endpoint config entirely in production, this cipher list only applies to the dev environment, so the blast radius is minimal. The list can safely be reduced to the four strong suites:

```
  ciphers: [
    ~c"ECDHE-ECDSA-AES256-GCM-SHA384",
    ~c"ECDHE-RSA-AES256-GCM-SHA384",
    ~c"ECDHE-ECDSA-AES128-GCM-SHA256",
    ~c"ECDHE-RSA-AES128-GCM-SHA256"
  ],
```

I was still wondering about the compatibility of ranch 1.8 with OTP 27, but I found this source [Ranch](https://ninenines.eu/articles/ranch-2.2.0/) which says directly that version 1.8.1 is compatible with OTP 28, so there should be no problem with our 27.

geekingfrog:
That should be good. I'd like to hold off this upgrade until we fix and deploy https://github.com/beyond-all-reason/teiserver/issues/1198
then do a deploy with hopefully only this change. As it can be a rather big one, it'll be good to have it in isolation.

# DO NOT MERGE (yet)

geekingfrog (approved):

BElluu:
@geekingfrog #1198 is closed, so ready for merge ? :)

PR #1185, opened 2026-05-25
Removed navigation banners with redundant links.

Zonkodonko:
Removed redundant nav banner from admin pages and battle pages


L-e-x-o-n:
Looked nice but was redundant, thanks.

PR #1189, opened 2026-05-25
MaxMembers in PartyState

### LINKED ISSUE #1074: Add `maxMember` to party state
Implement the changes from https://github.com/beyond-all-reason/tachyon/pull/120
do check the schema in tachyon master at time of implementing is still the same.

BElluu:
Closed #1074

### DIFF
--- a/lib/teiserver/player/tachyon_handler.ex
+++ b/lib/teiserver/player/tachyon_handler.ex
@@ -5,12 +5,14 @@ defmodule Teiserver.Player.TachyonHandler do
 
   alias Teiserver.Account
   alias Teiserver.CacheUser
+  alias Teiserver.Config
   alias Teiserver.Data.Types, as: T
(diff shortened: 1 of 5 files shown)


lib/teiserver/player/tachyon_handler.ex
```diff
@@ -1131,6 +1133,7 @@ defmodule Teiserver.Player.TachyonHandler do
             joinedAt: DateTime.to_unix(m.joined_at, :microsecond)
           }
         end),
+      maxMembers: Config.get_site_config_cache(Server.max_size_key()),
```
geekingfrog: This has to come from the party state. The handler should not have any party specific logic.

geekingfrog (commented):
This logic should live in the party process and nowhere else.

geekingfrog (approved):

PR #1192, opened 2026-05-25
Handle lobby tags

### LINKED ISSUE #1140: add support for lobby tags
introduced in https://github.com/beyond-all-reason/tachyon/pull/131

BElluu:
Closed #1140

### DIFF
--- a/lib/teiserver/tachyon_lobby/lobby.ex
+++ b/lib/teiserver/tachyon_lobby/lobby.ex
@@ -77,7 +77,8 @@ defmodule Teiserver.TachyonLobby.Lobby do
           optional(:game_version) => String.t(),
           optional(:engine_version) => String.t(),
           optional(:boss_enabled?) => boolean(),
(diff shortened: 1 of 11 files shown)


lib/teiserver/tachyon_lobby/lobby.ex
```diff
@@ -77,7 +77,8 @@ defmodule Teiserver.TachyonLobby.Lobby do
           optional(:game_version) => String.t(),
           optional(:engine_version) => String.t(),
           optional(:boss_enabled?) => boolean(),
-          optional(:game_options) => %{String.t() => String.t()}
+          optional(:game_options) => %{String.t() => String.t()},
+          optional(:tags) => %{String.t() => map()}
```
geekingfrog: I don't think it make sense to have this representation in the lobby state. Because of json merge patch we have this `%{String.t() => %{}}` but for the internals, a list or a set would be better.

geekingfrog (approved):

PR #1199, opened 2026-05-26
Added BannedPhrase functionality

Teifion:
- Adds functionality to start making use of BannedPhrases
- Uses a new style of Query module I would like to start moving towards
- Removes use of BannedPhrases from WordLib
- Adds the option to have a default value when getting something from the Cache
- Fixes a sometimes occurring duplicate component ID error with flashes
- Improves form error message styling

Includes error messages for incorrectly formatted phrases:
image

image

### DIFF
--- a/lib/teiserver/moderation/queries/banned_phrase_queries.ex
+++ b/lib/teiserver/moderation/queries/banned_phrase_queries.ex
@@ -1,0 +1,23 @@
+defmodule Teiserver.Moderation.BannedPhraseQueries do
+  @moduledoc false
+  alias Ecto.Query
+  alias Teiserver.Moderation.BannedPhrase
+
+  use TeiserverWeb, :queries
+
+  @type t :: Query.t()
+
+  @spec banned_phrases() :: t()
+  def banned_phrases do
+    from(banned_phrases in BannedPhrase, as: :banned_phrases)
+  end
+
+  @spec order_by_severity(t(), :asc | :desc) :: t()
+  def order_by_severity(query, direction \\ :asc) do
+    if direction == :asc do
+      from(banned_phrases in query, order_by: [asc: banned_phrases.severity])
+    else
+      from(banned_phrases in query, order_by: [desc: banned_phrases.severity])
+    end
+  end
+end
--- a/lib/teiserver/moderation/schemas/banned_phrase.ex
+++ b/lib/teiserver/moderation/schemas/banned_phrase.ex
@@ -1,8 +1,26 @@
 defmodule Teiserver.Moderation.BannedPhrase do
   @moduledoc """
-  A phrase banned in the game which would warrant an action if detected.
+  A phrase banned in the game which could warrant an action if detected.
+
+  All tests are checked as case sensitive.
+
+  Types:
+    raw - if the phrase is present in the message it will trigger
+    fuzzy - uses * as wildcards for as many characters as needed
+            executed as a regex
+    regex - a regex, if the pattern matches it will trigger
+
+  Score threshold is not currently used but is intended to be used
+  in the future for partial match types.
   """
+  alias Ecto.Changeset
+  alias Teiserver.Moderation
+  alias Teiserver.Moderation.BannedPhrase
+
   use TeiserverWeb, :schema
+
+  @type severity :: :low | :medium | :high
+  @type search_type :: :raw | :fuzzy | :regex
 
   typed_schema "banned_phrases" do
     field :phrase, :string
@@ -26,5 +44,127 @@ defmodule Teiserver.Moderation.BannedPhrase do
     |> cast(attrs, [:phrase, :score_threshold, :type, :severity])
     |> validate_required([:phrase, :score_threshold, :type, :severity])
     |> unique_constraint([:phrase])
+    |> validate_phrase()
+  end
+
+  defp validate_phrase(%Changeset{} = changeset) do
+    case get_field(changeset, :type) do
+      :fuzzy -> validate_fuzzy_phrase(changeset)
+      :regex -> validate_regex_phrase(changeset)
+      _any_other -> changeset
+    end
+  end
+
+  defp validate_fuzzy_phrase(%Changeset{} = changeset) do
+    phrase = get_field(changeset, :phrase) || ""
+
+    {regex_result, possible_regex_error} =
+      phrase
+      |> Regex.escape()
+      |> String.replace("\\*", ".*")
+      |> Regex.compile()
+
+    cond do
+      not String.contains?(phrase, "*") ->
+        add_error(changeset, :phrase, "Fuzzy phrase needs to contain at least one wildcard (*)")
+
+      regex_result == :error ->
+        {error_chars, pos} = possible_regex_error
+
+        error_str = List.to_string(error_chars)
+        add_error(changeset, :phrase, "Invalid regex - #{error_str}, pos: #{pos}")
+
+      true ->
+        changeset
+    end
+  end
+
+  defp validate_regex_phrase(%Changeset{} = changeset) do
+    pattern = get_field(changeset, :phrase)
+
+    case Regex.compile(pattern) do
+      {:ok, _regex} ->
+        changeset
+
+      {:error, {error_chars, pos}} ->
+        error_str = List.to_string(error_chars)
+        add_error(changeset, :phrase, "Invalid regex - #{error_str}, pos: #{pos}")
+    end
+  end
+
+  @doc """
+  Given a BannedPhrase loaded from the database cache the loaded_phrase
+  field.
+  """
+  @spec load_phrase(BannedPhrase.t()) :: BannedPhrase.t()
+  def load_phrase(%BannedPhrase{type: :regex, phrase: phrase} = banned_phrase) do
+    {:ok, compiled_phrase} = Regex.compile(phrase)
+    %BannedPhrase{banned_phrase | loaded_phrase: compiled_phrase}
+  end
+
+  def load_phrase(%BannedPhrase{type: :fuzzy, phrase: phrase} = banned_phrase) do
+    pattern =
+      phrase
+      |> Regex.escape()
+      |> String.replace("\\*", ".*")
+
+    {:ok, compiled_phrase} = Regex.compile(pattern)
+    %BannedPhrase{banned_phrase | loaded_phrase: compiled_phrase}
+  end
+
+  def load_phrase(%BannedPhrase{phrase: phrase} = banned_phrase) do
+    %BannedPhrase{banned_phrase | loaded_phrase: phrase}
+  end
+
+  @doc """
+  Performs a more extended set of automated checks to see if the message
+  being posted is
+  """
+  @spec message_severity(String.t(), severity(), [search_type()]) :: nil | BannedPhrase.severity()
+  def message_severity(message, min_severity \\ :high, types \\ [:raw, :fuzzy, :regex]) do
+    # First get the list of phrases we want to check against
+    banned_phrases =
+      Moderation.list_banned_phrases_cache()
+      |> Enum.filter(fn %BannedPhrase{type: type, severity: severity} ->
+        severity_is_at_least(severity, min_severity) and Enum.member?(types, type)
+      end)
+
+    banned_phrases
+    |> Enum.reduce(nil, fn
+      phrase, nil ->
+        if phrase_match?(phrase, message) do
+          phrase.severity
+        else
+          nil
+        end
+
+      _phrase, found_severity ->
+        found_severity
+    end)
+  end
+
+  defp severity_is_at_least(severity_tested, severity_minimum) do
+    case severity_minimum do
+      :low -> true
+      :medium -> Enum.member?([:medium, :high], severity_tested)
+      :high -> :high == severity_tested
+    end
+  end
+
+  @doc """
+  Given a banned phrase and a message, does the message
+  trigger the banned phrase?
+  """
+  @spec phrase_match?(BannedPhrase.t(), String.t()) :: boolean()
+  def phrase_match?(%BannedPhrase{type: :raw, loaded_phrase: loaded_phrase}, message) do
+    String.contains?(message, loaded_phrase)
+  end
+
+  def phrase_match?(%BannedPhrase{type: :fuzzy, loaded_phrase: loaded_phrase}, message) do
+    Regex.match?(loaded_phrase, message)
+  end
+
+  def phrase_match?(%BannedPhrase{type: :regex, loaded_phrase: loaded_phrase}, message) do
+    Regex.match?(loaded_phrase, message)
   end
 end
--- a/lib/teiserver/moderation/schemas/banned_phrase.ex
+++ b/lib/teiserver/moderation/schemas/banned_phrase.ex
@@ -58,30 +58,21 @@ defmodule Teiserver.Moderation.BannedPhrase do
   defp validate_fuzzy_phrase(%Changeset{} = changeset) do
     phrase = get_field(changeset, :phrase) || ""
 
-    {regex_result, possible_regex_error} =
-      phrase
-      |> Regex.escape()
-      |> String.replace("\\*", ".*")
-      |> Regex.compile()
+    pattern = phrase |> Regex.escape() |> String.replace("\\*", ".*")
 
-    cond do
-      not String.contains?(phrase, "*") ->
-        add_error(changeset, :phrase, "Fuzzy phrase needs to contain at least one wildcard (*)")
-
-      regex_result == :error ->
-        {error_chars, pos} = possible_regex_error
-
-        error_str = List.to_string(error_chars)
-        add_error(changeset, :phrase, "Invalid regex - #{error_str}, pos: #{pos}")
-
-      true ->
-        changeset
+    if String.contains?(phrase, "*") do
+      validate_pattern(changeset, pattern)
+    else
+      add_error(changeset, :phrase, "Fuzzy phrase needs to contain at least one wildcard (*)")
     end
   end
 
   defp validate_regex_phrase(%Changeset{} = changeset) do
     pattern = get_field(changeset, :phrase)
+    validate_pattern(changeset, pattern)
+  end
 
+  defp validate_pattern(%Changeset{} = changeset, pattern) do
     case Regex.compile(pattern) do
       {:ok, _regex} ->
         changeset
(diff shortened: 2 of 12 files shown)


config/dev.exs
```diff
@@ -66,10 +66,15 @@ config :teiserver, Teiserver,
...
+    notify: [
+      live_view: [
+        ~r"lib/teiserver_web/(live|live_components|components)/.*(ex|heex)$"
+      ]
+    ],
     patterns: [
       ~r"priv/static/.*(js|css|png|jpeg|jpg|gif|svg)$",
       ~r"priv/gettext/.*(po)$",
-      ~r"lib/teiserver_web/(controllers|live|components|live_components|views|templates)/.*(ex|heex)$"
+      ~r"lib/teiserver_web/(controllers|views|templates)/.*(ex|heex)$"
```
geekingfrog, line 77: this change seems to be coming from https://github.com/beyond-all-reason/teiserver/pull/1200#pullrequestreview-4370532409 and should be excluded from this PR (or you stack them)

lib/teiserver/moderation/queries/banned_phrase_queries.ex
```diff
@@ -0,0 +1,23 @@
...
+    from(banned_phrases in BannedPhrase, as: :banned_phrases)
+  end
+
+  @spec order_by_severity(t(), :asc | :desc) :: t()
+  def order_by_severity(query, direction \\ :asc) do
+    if direction == :asc do
+      from(banned_phrases in query, order_by: [asc: banned_phrases.severity])
+    else
+      from(banned_phrases in query, order_by: [desc: banned_phrases.severity])
+    end
```
geekingfrog, line 21: you don't need the if for the entire query:

```elixir
from(banned_phrases in query, order_by: [{direction, banned_phrases.severity}])
```
should work the same

lib/teiserver/moderation/queries/banned_phrase_queries.ex
geekingfrog, line 1: It's a bit small and early to tell, but I think I like this approach of queries.
I also did some variation on the existing pattern in the oauth queries (like [there](https://github.com/beyond-all-reason/teiserver/blob/2356a29e95e1870a36bfffffa0fc86542470647c/lib/teiserver/o_auth/queries/application_query.ex#L4) for example), which seems similar to what you have there.

I think having named binding is nice, because it allows `has_named_binding?` to do composable joins. This should allow filtering on different parts of the joins through simple function composition.

lib/teiserver/moderation/schemas/banned_phrase.ex
```diff
@@ -1,9 +1,27 @@
...
+  Score threshold is not currently used but is intended to be used
+  in the future for partial match types.
   """
+  alias Ecto.Changeset
+  alias Teiserver.Moderation
+  alias Teiserver.Moderation.BannedPhrase
+
   use TeiserverWeb, :schema
 
+  @type severity :: :low | :medium | :high
```
geekingfrog, line 22: I missed that in a previous PR I think, but I'm not a fan of storing that in the db itself. I think a number (from 0 to 1, or 0 to 100?) would be better, and we can give it meaning in application code.
The advantage of a number is that we can modify these categories without having to change the underlying data, and also query "everything medium or above" with a numeric comparison (making `severity_is_at_least` redundant)

lib/teiserver/moderation/schemas/banned_phrase.ex
```diff
@@ -26,5 +44,127 @@ defmodule Teiserver.Moderation.BannedPhrase do
     |> cast(attrs, [:phrase, :score_threshold, :type, :severity])
     |> validate_required([:phrase, :score_threshold, :type, :severity])
     |> unique_constraint([:phrase])
+    |> validate_phrase()
```
geekingfrog, line 47: before the `validate_phrase` you should also have a [validate_subset](https://hexdocs.pm/ecto/Ecto.Changeset.html#validate_subset/4) to make sure you're getting the correct keywords for search type.

lib/teiserver/moderation/schemas/banned_phrase.ex
```diff
@@ -26,5 +44,127 @@ defmodule Teiserver.Moderation.BannedPhrase do
...
+
+    {regex_result, possible_regex_error} =
+      phrase
+      |> Regex.escape()
+      |> String.replace("\\*", ".*")
+      |> Regex.compile()
+
+    cond do
+      not String.contains?(phrase, "*") ->
+        add_error(changeset, :phrase, "Fuzzy phrase needs to contain at least one wildcard (*)")
```
geekingfrog: I think this is just annoying. If there is no wildcard, why not change the type to `:raw` and allow it?

lib/teiserver/moderation/schemas/banned_phrase.ex
```diff
@@ -26,5 +44,127 @@ defmodule Teiserver.Moderation.BannedPhrase do
...
+  @doc """
+  Given a banned phrase and a message, does the message
+  trigger the banned phrase?
+  """
+  @spec phrase_match?(BannedPhrase.t(), String.t()) :: boolean()
+  def phrase_match?(%BannedPhrase{type: :raw, loaded_phrase: loaded_phrase}, message) do
+    String.contains?(message, loaded_phrase)
+  end
+
+  def phrase_match?(%BannedPhrase{type: :fuzzy, loaded_phrase: loaded_phrase}, message) do
```
geekingfrog: you can also do
```elixir
def foo(%Foo{type: type}) when type in [:regex, :fuzzy] do
```

lib/teiserver/moderation/schemas/banned_phrase.ex
```diff
@@ -26,5 +44,127 @@ defmodule Teiserver.Moderation.BannedPhrase do
...
+  defp validate_regex_phrase(%Changeset{} = changeset) do
+    pattern = get_field(changeset, :phrase)
+
+    case Regex.compile(pattern) do
+      {:ok, _regex} ->
+        changeset
+
+      {:error, {error_chars, pos}} ->
+        error_str = List.to_string(error_chars)
+        add_error(changeset, :phrase, "Invalid regex - #{error_str}, pos: #{pos}")
```
geekingfrog, line 86: looking at this duplication, I would extract a function

```elixir
defp validate_regex(changeset, pattern) do
  case Regex.compile(pattern) do
  ...
  end
end
```
that can be called from both the `validate_regex_phrase` and the `validate_fuzzy_phrase` once they extracted their respective regex.

lib/teiserver/moderation/schemas/banned_phrase.ex
```diff
@@ -26,5 +44,127 @@ defmodule Teiserver.Moderation.BannedPhrase do
...
+    %BannedPhrase{banned_phrase | loaded_phrase: compiled_phrase}
+  end
+
+  def load_phrase(%BannedPhrase{phrase: phrase} = banned_phrase) do
+    %BannedPhrase{banned_phrase | loaded_phrase: phrase}
+  end
+
+  @doc """
+  Performs a more extended set of automated checks to see if the message
+  being posted is
```
geekingfrog: the end of the sentence is missing

lib/teiserver/moderation/schemas/banned_phrase.ex
```diff
@@ -26,5 +44,127 @@ defmodule Teiserver.Moderation.BannedPhrase do
...
+      phrase, nil ->
+        if phrase_match?(phrase, message) do
+          phrase.severity
+        else
+          nil
+        end
+
+      _phrase, found_severity ->
+        found_severity
+    end)
```
geekingfrog: This is [Enum.find/3](https://hexdocs.pm/elixir/Enum.html#find/3), and you can merge the two operations together even:
```elixir
Moderation.list_banned_phrases_cache()
|> Enum.find(fn %BannedPhrase{type: type, severity: severity} ->
        severity_is_at_least(severity, min_severity) and Enum.member?(types, type) and phrase_match?(phrase, message)
      end)
```

test/teiserver/moderation/banned_phrase_test.exs
```diff
@@ -0,0 +1,272 @@
...
+    test "delete_banned_phrase/1 deletes the banned_phrase" do
+      banned_phrase = banned_phrase_fixture()
+      assert {:ok, %BannedPhrase{}} = Moderation.delete_banned_phrase(banned_phrase)
+      assert_raise Ecto.NoResultsError, fn -> Moderation.get_banned_phrase!(banned_phrase.id) end
+    end
+
+    test "change_banned_phrase/1 returns a banned_phrase changeset" do
+      banned_phrase = banned_phrase_fixture()
+      assert %Ecto.Changeset{} = Moderation.change_banned_phrase(banned_phrase)
+    end
```
geekingfrog: not convinced by the usefulness of this test, considering the implementation is 1 line

test/teiserver/moderation/banned_phrase_test.exs
```diff
@@ -0,0 +1,272 @@
...
+      {:error, changeset} =
+        Moderation.create_banned_phrase(%{
+          phrase: "abc",
+          type: :fuzzy,
+          severity: :low,
+          score_threshold: 0
+        })
+
+      assert changeset.errors == [
+               {:phrase, {"Fuzzy phrase needs to contain at least one wildcard (*)", []}}
```
geekingfrog: I am not a fan of asserting against specific error messages. When fixing a typo for example, it would make the test fails. There is no real logic being asserted with this.
What I'd suggest instead is to check the correct attribute has an associated error, which should be enough:
`assert Keyword.has_key?(changeset.errors, :phrase)`

test/teiserver/moderation/banned_phrase_test.exs
```diff
@@ -0,0 +1,272 @@
...
+    test "regex banned phrase - bad regex" do
+      {:error, changeset} =
+        Moderation.create_banned_phrase(%{
+          phrase: "(abc",
+          type: :regex,
+          severity: :low,
+          score_threshold: 0
+        })
+
+      assert changeset.errors == [{:phrase, {"Invalid regex - missing ), pos: 4", []}}]
```
geekingfrog: similar wrt error message

lib/teiserver/moderation/schemas/banned_phrase.ex
```diff
@@ -26,5 +44,127 @@ defmodule Teiserver.Moderation.BannedPhrase do
...
+  def load_phrase(%BannedPhrase{phrase: phrase} = banned_phrase) do
+    %BannedPhrase{banned_phrase | loaded_phrase: phrase}
+  end
+
+  @doc """
+  Performs a more extended set of automated checks to see if the message
+  being posted is
+  """
+  @spec message_severity(String.t(), severity(), [search_type()]) :: nil | BannedPhrase.severity()
+  def message_severity(message, min_severity \\ :high, types \\ [:raw, :fuzzy, :regex]) do
```
geekingfrog, line 119: what's the intent behind `min_severity`?

config/dev.exs
Teifion: I rebased it and am not seeing a change in the files changed now.

lib/teiserver/moderation/queries/banned_phrase_queries.ex
Teifion: Yeah, I got it from a place I worked at recently and it was far superior compared to what I started Teiserver with. The composable queries are much nicer, more expressive, more extensible, allow jumping to functions and the join part you mentioned :)

lib/teiserver/moderation/schemas/banned_phrase.ex
Teifion: I don't completely agree or disagree, let me address everything else and I'll return to this point having thought about it some more.

lib/teiserver/moderation/queries/banned_phrase_queries.ex
Teifion: Sadly it generates a compile time error:
```elixir
  def order_by_severity(query, direction \\ :asc) do
    from(banned_phrases in query, order_by: [{direction, banned_phrases.severity}])
  end
```

```
== Compilation error in file lib/teiserver/moderation/queries/banned_phrase_queries.ex ==
** (Ecto.Query.CompileError) expected :asc, :asc_nulls_last, :asc_nulls_first, :desc, :desc_nulls_last, :desc_nulls_first or interpolated value in `order_by`, got: `{:direction, [line: 17, column: 47], nil}`
```

lib/teiserver/moderation/schemas/banned_phrase.ex
Teifion: That's handled by Ecto automatically because it's an Enum field:

```ex
Moderation.create_banned_phrase(%{
        phrase: "xyz",
        type: :notype,
        severity: :off,
        score_threshold: 0
      })

# ... snip ...

errors: [
     type: {"is invalid",
      [
        type: {:parameterized,
         {Ecto.Enum,
          %{
            on_load: %{"fuzzy" => :fuzzy, "raw" => :raw, "regex" => :regex},
            type: :string,
            embed_as: :self,
            on_cast: %{"fuzzy" => :fuzzy, "raw" => :raw, "regex" => :regex},
            mappings: [raw: "raw", fuzzy: "fuzzy", regex: "regex"],
            on_dump: %{raw: "raw", regex: "regex", fuzzy: "fuzzy"}
          }}},
        validation: :inclusion,
        enum: ["fuzzy", "raw", "regex"]
      ]},
     severity: {"is invalid",
      [
        type: {:parameterized,
         {Ecto.Enum,
          %{
            on_load: %{"high" => :high, "low" => :low, "medium" => :medium},
            type: :string,
            embed_as: :self,
            on_cast: %{"high" => :high, "low" => :low, "medium" => :medium},
            mappings: [low: "low", medium: "medium", high: "high"],
            on_dump: %{high: "high", low: "low", medium: "medium"}
          }}},
        validation: :inclusion,
        enum: ["high", "low", "medium"]
      ]}
   ],
```

lib/teiserver/moderation/schemas/banned_phrase.ex
Teifion: Because someone intended it to be fuzzy and forgot the wildcard. Changing it to raw means you're masking/hiding their error and can allow the mistake to propagate. How would you feel about an error message suggesting they use raw if they don't actually need the wildcard?

lib/teiserver/moderation/schemas/banned_phrase.ex
Teifion: I'm not sure I follow on this one, I had a go at doing that and it just made the code longer because the only duplication is the creation of the error_str and adding the error. Could you provide a larger example of how you envisaged it being done please?

test/teiserver/moderation/banned_phrase_test.exs
Teifion: A generator created one, now I look at it I agree with you. I will remove it.

lib/teiserver/moderation/schemas/banned_phrase.ex
Teifion: High severity would be used for established users but low would be used for anybody un-trusted (e.g. brand new accounts).

lib/teiserver/moderation/queries/banned_phrase_queries.ex
geekingfrog: grmbl, it's probably because all of these are macro :/ oh well

lib/teiserver/moderation/schemas/banned_phrase.ex
geekingfrog: This is what I had in mind

```elixir
  defp validate_fuzzy_phrase(%Changeset{} = changeset) do
    phrase = get_field(changeset, :phrase) || ""

    pattern = phrase |> Regex.escape() |> String.replace("\\*", ".*")

    cond do
      not String.contains?(phrase, "*") ->
        add_error(changeset, :phrase, "Fuzzy phrase needs to contain at least one wildcard (*)")

      true ->
        validate_pattern(changeset, pattern)
    end
  end

  defp validate_regex_phrase(%Changeset{} = changeset) do
    pattern = get_field(changeset, :phrase)
    validate_pattern(changeset, pattern)
  end

  defp validate_pattern(%Changeset{} = changeset, pattern) do
    case Regex.compile(pattern) do
      {:ok, _regex} ->
        changeset

      {:error, {error_chars, pos}} ->
        error_str = List.to_string(error_chars)
        add_error(changeset, :phrase, "Invalid regex - #{error_str}, pos: #{pos}")
    end
  end
```

lib/teiserver/moderation/schemas/banned_phrase.ex
Teifion: Addressed in https://github.com/beyond-all-reason/teiserver/pull/1199/commits/2af12e29e0f886119287b3dee0baa562bf2019b4

lib/teiserver/moderation/schemas/banned_phrase.ex
Teifion: Entire doc was off! Addressed in https://github.com/beyond-all-reason/teiserver/pull/1199/commits/2af12e29e0f886119287b3dee0baa562bf2019b4

lib/teiserver/moderation/schemas/banned_phrase.ex
Teifion: Good spot, addressed in https://github.com/beyond-all-reason/teiserver/pull/1199/commits/2af12e29e0f886119287b3dee0baa562bf2019b4

test/teiserver/moderation/banned_phrase_test.exs
Teifion: Addressed in https://github.com/beyond-all-reason/teiserver/pull/1199/commits/2af12e29e0f886119287b3dee0baa562bf2019b4

test/teiserver/moderation/banned_phrase_test.exs
Teifion: Addressed in https://github.com/beyond-all-reason/teiserver/pull/1199/commits/2af12e29e0f886119287b3dee0baa562bf2019b4

lib/teiserver/moderation/schemas/banned_phrase.ex
Teifion: Thank you, implemented in https://github.com/beyond-all-reason/teiserver/pull/1199/commits/5c19d56fefdbc9bced5e021ab295948f44482f00

L-e-x-o-n (approved):

lib/teiserver/moderation/schemas/banned_phrase.ex
Teifion: I have thought more about it and think an enum is still appropriate. I'm happy to discuss it further if you feel strongly on it.

lib/teiserver/moderation/schemas/banned_phrase.ex
Teifion: I have updated the error message in https://github.com/beyond-all-reason/teiserver/pull/1199/commits/4c0f82c74edb02d08914bb58a18ea79bc967178f

PR #1213, opened 2026-05-28
Moderation related fixes and improvements

Teifion:
- User search by data improvements
- New account function
- Stub function check for VPN and similar users to have to manually verify
- Site configs in preparation for external IP lookup
- Cluster manager lookup tweaked
- Added option for moderators to shadowban manually (and revert it)
- Expanded smurf search listing

### DIFF
--- a/lib/teiserver/account/libs/user_lib.ex
+++ b/lib/teiserver/account/libs/user_lib.ex
@@ -73,10 +73,17 @@ defmodule Teiserver.Account.UserLib do
 
   @spec list_users_by_data(%{String.t() => String.t()}) :: [User.t()]
   def list_users_by_data(data_search_params) do
-    data_search_params
-    |> UserQueries.user_search_by_data()
-    |> QueryHelpers.limit_query(200)
-    |> Repo.all()
+    # If nil is returned then the arguments passed in were invalid or
+    # problematic in some way and we will not be running the query
+    case UserQueries.user_search_by_data(data_search_params) do
+      {:ok, query} ->
+        query
+        |> QueryHelpers.limit_query(200)
+        |> Repo.all()
+
+      _any_other ->
+        []
+    end
   end
 
   @spec count_users() :: integer
--- a/lib/teiserver/account/libs/user_lib.ex
+++ b/lib/teiserver/account/libs/user_lib.ex
@@ -7,6 +7,7 @@ defmodule Teiserver.Account.UserLib do
   alias Teiserver.Account.User
   alias Teiserver.Account.UserQueries
   alias Teiserver.CacheUser
+  alias Teiserver.Config
   alias Teiserver.Helper.StylingHelper
   alias Teiserver.Logging
   use TeiserverWeb, :library_newform
@@ -68,6 +69,28 @@ defmodule Teiserver.Account.UserLib do
     args
     |> UserQueries.query_users()
     |> Repo.all()
+  end
+
+  @spec list_users_by_data(%{String.t() => String.t()}) :: [User.t()]
+  def list_users_by_data(data_search_params) do
+    id_list =
+      Account.list_user_stats(
+        search: [
+          data_equal: {"hardware:gpuinfo", data_search_params["gpu"]},
+          data_equal: {"hardware:cpuinfo", data_search_params["cpu"]},
+          data_equal: {"hardware:osinfo", data_search_params["os"]},
+          data_equal: {"hardware:raminfo", data_search_params["ram"]},
+          data_equal: {"hardware:displaymax", data_search_params["screen"]},
+          data_string_starts_with: {"last_ip", data_search_params["ip"]},
+          data_equal: {data_search_params["custom_field"], data_search_params["custom_value"]}
+        ],
+        select: [:user_id],
+        limit: :infinity
+      )
+      |> Stream.map(fn stats -> stats.user_id end)
+      |> Enum.to_list()
+
+    list_users(search: [id_in: id_list])
   end
 
   @spec count_users() :: integer
@@ -430,5 +453,13 @@ defmodule Teiserver.Account.UserLib do
       true ->
         :ok
     end
+  end
+
+  @spec new_account?(User.id()) :: boolean()
+  def new_account?(user_id) do
+    user_stats = Account.get_user_stat_data(user_id)
+    play_hours = (user_stats["player_minutes"] || 0) / 60
+    play_time_cutoff = Config.get_site_config_cache("teiserver.New player cutoff")
+    play_hours < play_time_cutoff
   end
 end
--- a/lib/teiserver/account/libs/user_lib.ex
+++ b/lib/teiserver/account/libs/user_lib.ex
@@ -73,24 +73,9 @@ defmodule Teiserver.Account.UserLib do
 
   @spec list_users_by_data(%{String.t() => String.t()}) :: [User.t()]
   def list_users_by_data(data_search_params) do
-    id_list =
-      Account.list_user_stats(
-        search: [
-          data_equal: {"hardware:gpuinfo", data_search_params["gpu"]},
-          data_equal: {"hardware:cpuinfo", data_search_params["cpu"]},
-          data_equal: {"hardware:osinfo", data_search_params["os"]},
-          data_equal: {"hardware:raminfo", data_search_params["ram"]},
-          data_equal: {"hardware:displaymax", data_search_params["screen"]},
-          data_string_starts_with: {"last_ip", data_search_params["ip"]},
-          data_equal: {data_search_params["custom_field"], data_search_params["custom_value"]}
-        ],
-        select: [:user_id],
-        limit: :infinity
-      )
-      |> Stream.map(fn stats -> stats.user_id end)
-      |> Enum.to_list()
-
-    list_users(search: [id_in: id_list])
+    data_search_params
+    |> UserQueries.user_search_by_data()
+    |> Repo.all()
   end
 
   @spec count_users() :: integer
--- a/lib/teiserver/account/libs/user_lib.ex
+++ b/lib/teiserver/account/libs/user_lib.ex
@@ -75,6 +75,7 @@ defmodule Teiserver.Account.UserLib do
   def list_users_by_data(data_search_params) do
     data_search_params
     |> UserQueries.user_search_by_data()
+    |> QueryHelpers.limit_query(200)
     |> Repo.all()
   end
 
--- a/lib/teiserver/account/queries/user_queries.ex
+++ b/lib/teiserver/account/queries/user_queries.ex
@@ -481,7 +481,7 @@ defmodule Teiserver.Account.UserQueries do
       preload: [reports_responded: reports_responded]
   end
 
-  @spec user_search_by_data(%{String.t() => String.t()}) :: t()
+  @spec user_search_by_data(%{String.t() => String.t()}) :: {:ok, t()} | {:error, String.t()}
   def user_search_by_data(values) do
     # When submitting from the web, a blank value is an empty string so
     # we need to filter that out as if it was a nil value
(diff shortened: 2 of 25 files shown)


lib/teiserver/account/libs/user_lib.ex
```diff
@@ -70,6 +71,28 @@ defmodule Teiserver.Account.UserLib do
...
+          data_string_starts_with: {"last_ip", data_search_params["ip"]},
+          data_equal: {data_search_params["custom_field"], data_search_params["custom_value"]}
+        ],
+        select: [:user_id],
+        limit: :infinity
+      )
+      |> Stream.map(fn stats -> stats.user_id end)
+      |> Enum.to_list()
+
+    list_users(search: [id_in: id_list])
```
geekingfrog: This is not required. The `list_user_stats` could preload the users since there's a foreign key, and then you can `Enum.map` to get these. Potentially going through a MapSet to eliminate duplicates.

lib/teiserver/account/libs/user_lib.ex
```diff
@@ -70,6 +71,28 @@ defmodule Teiserver.Account.UserLib do
     |> Repo.all()
   end
 
+  @spec list_users_by_data(%{String.t() => String.t()}) :: [User.t()]
+  def list_users_by_data(data_search_params) do
```
geekingfrog, line 75: There should be a test for this function. Also, if there is no `ip` field, this will crash. Are all the parameters required? If yes, maybe having a struct, or a map with specified keys would be better for the signature?

Also, if the params don't contain `gpu`, it will search for `gpu=nil`, which I don't think is what we want.

What do you think of

```elixir
    params = [
      {:data_equal, "hardware:gpuinfo", data_search_params["gpu"]},
      {:data_equal, "hardware:cpuinfo", data_search_params["cpu"]},
      {:data_equal, "hardware:osinfo", data_search_params["os"]},
      {:data_equal, "hardware:raminfo", data_search_params["ram"]},
      {:data_equal, "hardware:displaymax", data_search_params["screen"]},
      {:data_string_starts_with, "last_ip", "ip"}
    ]

    params =
      if Map.has_key?(data_search_params, "custom_field") do
        [{:data_equal, "custom_field", "custom_value"} | params]
      else
        params
      end

    search =
      Enum.reduce(params, [], fn {search_key, source, param_key}, search ->
        if Map.has_key?(data_search_params, param_key) do
          [{search_key, {source, data_search_params[param_key]}} | search]
        else
          search
        end
      end)

    id_list =
      Account.list_user_stats(
        search: search,
        select: [:user_id],
        limit: :infinity
      )
      |> Stream.map(fn stats -> stats.user_id end)
      |> Enum.to_list()
```

The custom field makes it a little bit awkward but reasonable I think.

lib/teiserver/battle/servers/match_monitor_server.ex
```diff
@@ -380,6 +382,27 @@ defmodule Teiserver.Battle.MatchMonitorServer do
     {:noreply, state}
   end
 
+  def handle_griefing(username, event_type_name, game_time, match_id) do
+    userid = Account.get_userid_from_name(username)
```
geekingfrog: this should be inside the branch where the account is actually required (after the check on event type name)

lib/teiserver/moderation/tasks/external_ip_check_task.ex
```diff
@@ -0,0 +1,45 @@
...
+      URI.encode_query(%{
+        "q" => ip,
+        "key" => key
+      })
+
+    url = "#{endpoint}?#{query}"
+
+    case Req.get(url) do
+      {:ok, %Response{body: body}} ->
+        data = body |> Jason.decode!()
```
geekingfrog: the decode shouldn't be needed. `req` by defaults decode the body based on its content type.

lib/teiserver/moderation/tasks/external_ip_check_task.ex
```diff
@@ -0,0 +1,45 @@
...
+          {:is_crawler, "teiserver.external_ip_ban_is_crawler"},
+          {:is_datacenter, "teiserver.external_ip_ban_is_datacenter"},
+          {:is_proxy, "teiserver.external_ip_ban_is_proxy"},
+          {:is_tor, "teiserver.external_ip_ban_is_tor"},
+          {:is_vpn, "teiserver.external_ip_ban_is_vpn"}
+        ]
+        |> Enum.filter(fn {k, v} ->
+          Config.get_site_config_cache(v) and data[to_string(k)]
+        end)
+        |> Enum.map(fn {k, _v} -> k end)
```
geekingfrog, line 28: `Map.keys()`

lib/teiserver/moderation/tasks/external_ip_check_task.ex
```diff
@@ -0,0 +1,45 @@
...
+    key = Config.get_site_config_cache("teiserver.External IP check key")
+    endpoint = Config.get_site_config_cache("teiserver.External IP check endpoint")
+
+    query =
+      URI.encode_query(%{
+        "q" => ip,
+        "key" => key
+      })
+
+    url = "#{endpoint}?#{query}"
```
geekingfrog: you don't need to manually encode the query.
```elixir
Req.get(endpoint, params: [q: ip, key: key])
```

lib/teiserver/protocols/spring/spring_in.ex
```diff
@@ -348,18 +355,39 @@ defmodule Teiserver.Protocols.SpringIn do
...
+        correct_code = Map.get(stat_data, "verification_code")
+
+        require_manual_email? =
+          Enum.any?([
+            Map.get(stat_data, "vpn?", false),
+            Map.get(stat_data, "ip_check_bad?", false)
+          ])
+
+        cond do
+          code != to_string(correct_code) ->
```
geekingfrog: why is the `to_string` required there? shouldn't that be a string already?

geekingfrog (commented):
I think the more meaty functions could do with some tests.

And also, this PR was a bit all over the place. Some new features, most of them completely unrelated to one another.
Some renaming of functions, some refactoring of existing things (like the hardware key stuff). All of these could have been done in separate commits, making reviewing a lot easier to avoid context switching between features.

lib/teiserver/account/libs/user_lib.ex
Teifion: I've gone a slightly different route because of default values sent through the html form, addressed in https://github.com/beyond-all-reason/teiserver/pull/1213/commits/3d9200e95a42c40314fdb30eaf5418674ca33397

lib/teiserver/account/libs/user_lib.ex
Teifion: Tests and new query added in https://github.com/beyond-all-reason/teiserver/pull/1213/commits/3d9200e95a42c40314fdb30eaf5418674ca33397

lib/teiserver/moderation/tasks/external_ip_check_task.ex
Teifion: It was `Map.keys` in the version supplied elsewhere but that was generating Dialyzer errors. I swapped it back to check and the errors are persisting so I've put it as this for now. Happy to go back to `Map.keys` and put in an ignore if you'd rather.

lib/teiserver/protocols/spring/spring_in.ex
Teifion: Leftover from a copy-paste mistake, addressed in https://github.com/beyond-all-reason/teiserver/pull/1213/commits/14cccd6a3a36b5cf8859f509b6aa9a4fa119aab6

lib/teiserver/moderation/tasks/external_ip_check_task.ex
Teifion: Thank you, addressed in https://github.com/beyond-all-reason/teiserver/pull/1213/commits/14cccd6a3a36b5cf8859f509b6aa9a4fa119aab6

lib/teiserver/moderation/tasks/external_ip_check_task.ex
Teifion: Thank you, addressed in https://github.com/beyond-all-reason/teiserver/pull/1213/commits/14cccd6a3a36b5cf8859f509b6aa9a4fa119aab6

lib/teiserver/battle/servers/match_monitor_server.ex
Teifion: Removed entirely in https://github.com/beyond-all-reason/teiserver/pull/1213/commits/14cccd6a3a36b5cf8859f509b6aa9a4fa119aab6

lib/teiserver/moderation/tasks/external_ip_check_task.ex
geekingfrog: ah yeah, I know why, because the result of `filter` is a list of pairs, not a map.

test/teiserver/account/user_lib_test.exs
```diff
@@ -0,0 +1,96 @@
...
+               "ip" => "127."
+             }) == [user3]
+
+      # Custom value
+      assert user_ids_by_data(%{
+               "custom_field" => "some_key",
+               "custom_value" => "some_value"
+             }) == [user1, user3]
+    end
+  end
```
geekingfrog, line 94: bit of a nitpick but I'm not the biggest fan of a massive test for everything. I prefer having more granular tests under a `describe`, and use setup functions (or copy/paste stuff, it's fine in test imo).

test/teiserver/account/user_lib_test.exs
```diff
@@ -0,0 +1,96 @@
+defmodule Teiserver.Account.UserLibTest do
+  @moduledoc false
+  alias Teiserver.Account
+  alias Teiserver.Account.UserLib
+  alias Teiserver.AccountFixtures
+
+  use Teiserver.DataCase, async: false
+
+  describe "list_users_by_data" do
```
geekingfrog, line 9: There should be a test for empty and nil values as well

geekingfrog (commented):
some small changes for tests and that should be good to go.

I also realised we should think about a way to run the IP check locally without having to provision a real key. Probably something along the lines of the module is config dependent/use dependency injection and in test/dev it is stubbed out.
That's out of scope for this PR though

Teifion:
> I also realised we should think about a way to run the IP check locally without having to provision a real key

Something I planned to do as part of making use of the data. Have been weighing up using the existing `mock` vs using `mox`. If you have thoughts on the matter I'd very much welcome them.

geekingfrog:
aaaaaaah, also forgot something important, either put a limit on the query, or ensure that if the passed map is empty or only has `""` or nil values, then we don't return **all** users. We had similar problems in the past with other queries

geekingfrog:
> > I also realised we should think about a way to run the IP check locally without having to provision a real key
> 
> Something I planned to do as part of making use of the data. Have been weighing up using the existing `mock` vs using `mox`. If you have thoughts on the matter I'd very much welcome them.

In general, I dislike mocking a lot. Mocking usually lead to tests that are very tightly coupled to implementation, and often, changing irrelevant things in code, will require changes in tests as well.
It also often break abstraction boundaries.

When it comes to this kind of problem, I prefer dependency injection in a way that you can control in test if required.

So for the ip query stuff for example:

```elixir
#### in the module where you need to query some data about an IP
def get_ip_info(ip, client \\ nil) do
  client = client || get_client()
  client.query_ip(client, ip)
end

defp get_client() do
  module = Application.get_env(:teiserver, :ip_client) # this is a module name
  apply(module, :new_client, [])
  # or perhaps, more simply, hardcode the default module
  IPQuery.new_client()
end

#### behaviour definition
defmodule IPQuery do
  @callback new_client() :: term()
  @callback query_ip(client :: term(), ip :: String.t()) :: [term()]
end

#### the "production" client
defmodule IPClient do
  @behaviour IPQuery

  defstruct [:endpoint, :key]

  @impl IPQuery
  def new_client() do
    %__MODULE__{
      endpoint: Config.get_site_config_cache("teiserver.External IP check endpoint"),
      key: Config.get_site_config_cache("teiserver.External IP check key")
    }
  end

  @impl IPQuery
  def query_ip(%__MODULE__{} = client, ip) do
    resp = Req.get(client.endpoint, params: [q: ip, key: client.key])
    # snip
  end
end

#### in test, you can use this module where you can pass the module with the response you expect or
# whatever is needed for the test.
defmodule IPClientTest do
  @impl IPQuery
  def new_client() do
    %{mocked_responses: %{"127.0.0.1" => [:is_vpn]}}
  end

  @impl IPQuery
  def query_ip(client, ip) do
    Map.get(client.mocked_responses, ip, [])
  end
end
```

one advantage to have the module in the configuration is that we can provide something that works out of the box for local development, and not only tests with mocks. The application environment can also be modified in a test if you need the mocked client in a path where there is no direct access to the functions requiring it.

test/teiserver/account/user_lib_test.exs
Teifion: Given the addition of nil tests I now agree breaking them up makes sense, addressed in https://github.com/beyond-all-reason/teiserver/pull/1213/commits/123c8f04a93b67c9eb61be8ddc5620cd2c03b6dd

test/teiserver/account/user_lib_test.exs
Teifion: Good shout, added tests in https://github.com/beyond-all-reason/teiserver/pull/1213/commits/123c8f04a93b67c9eb61be8ddc5620cd2c03b6dd

Teifion:
@geekingfrog I've addressed the outstanding test comment and started the thread on discord to ensure the testing of the IP task isn't forgotten about.

lib/teiserver/account/queries/user_queries.ex
```diff
@@ -477,4 +480,46 @@ defmodule Teiserver.Account.UserQueries do
       left_join: reports_responded in assoc(users, :reports_responded),
       preload: [reports_responded: reports_responded]
   end
+
+  @spec user_search_by_data(%{String.t() => String.t()}) :: t()
```
geekingfrog: Isn't that returning a `[t()]` ?

lib/teiserver/account/queries/user_queries.ex
```diff
@@ -477,4 +480,46 @@ defmodule Teiserver.Account.UserQueries do
...
+      end)
+
+    # We want to treat IP slightly differently so we do that here
+    if values["ip"] != "" and values["ip"] != nil do
+      from [user_stats: user_stats] in query,
+        where:
+          like(fragment("(? ->> ?)::text", user_stats.data, "last_ip"), ^(values["ip"] <> "%"))
+    else
+      query
+    end
```
geekingfrog, line 534: There should be a case (+ test) handling if the search params is empty and/or a limit on the query so that passing invalid/empty map doesn't return every single user

lib/teiserver/account/queries/user_queries.ex
Teifion: Returns a query which is `t()` as part of the changes to this module.

lib/teiserver/account/queries/user_queries.ex
Teifion: Good shout, the limit would have prevented it from borking the server but it's still not great. I have added a handler for it in https://github.com/beyond-all-reason/teiserver/pull/1213/commits/d8ac7c979db99a8516eba1ee5c1ad9621c82d20f, would have been so much easier if IP wasn't a partial search :D

geekingfrog (approved):

Issue #1214, opened 2026-05-28
Fix datetime issues after Timex removal


L-e-x-o-n:
Removal of Timex in https://github.com/beyond-all-reason/teiserver/pull/1157 resulted in some pages no longer working.

- /logging/server
- /teiserver/reports/show/records
- /teiserver/reports/show/moderation_activity
- /teiserver/reports/show/time_spent
- /logging/match/day_metrics
```
** (MatchError) no match of right hand side value:

    ~D[2026-05-27]

        (elixir 1.19.4) lib/calendar/datetime.ex:748: DateTime.shift_zone/3
        (teiserver 0.1.0) lib/teiserver/helpers/date_helper.ex:5: Teiserver.Helper.DateHelper.convert/2
        (teiserver 0.1.0) lib/teiserver/helpers/date_helper.ex:64: Teiserver.Helper.DateHelper.date_to_str/2
        (teiserver 0.1.0) lib/teiserver_web/templates/logging/match_log/day_metrics_list.html.heex:76: anonymous fn/6 in TeiserverWeb.Logging.MatchLogView."day_metrics_list.html"/1
```

- /teiserver/reports/show/retention
- /teiserver/reports/show/user_age

```
** (FunctionClauseError) no function clause matching in DateTime.diff/3
        (elixir 1.19.4) lib/calendar/datetime.ex:1578: DateTime.diff(~U[2026-05-28 21:32:53.461659Z], ~N[2024-08-10 00:05:59], :second)
        (elixir 1.19.4) lib/calendar/datetime.ex:1579: DateTime.diff/3
        (teiserver 0.1.0) lib/teiserver/account/reports/user_age_report.ex:163: Teiserver.Account.UserAgeReport.get_registration_age/1
```

There may be more that haven't been found yet.

PR #1215, opened 2026-05-28
Fix datehelper formatting issue, removed some un-used functions

Teifion:
Addressing https://github.com/beyond-all-reason/teiserver/issues/1214

### DIFF
--- a/lib/teiserver/helpers/date_helper.ex
+++ b/lib/teiserver/helpers/date_helper.ex
@@ -1,11 +1,35 @@
 defmodule Teiserver.Helper.DateHelper do
   @moduledoc false
 
-  defp convert(timestamp, tz) do
+  defp convert(%Date{} = date, tz) do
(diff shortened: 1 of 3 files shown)


lib/teiserver/helpers/date_helper.ex
```diff
@@ -1,13 +1,37 @@
...
+  @doc """
+  Wraps the Calendar.strftime function to give a different
+  output if it is the same day (time) or a different day (date)
+  """
+  @spec hms_or_ymd(DateTime.t()) :: String.t()
+  def hms_or_ymd(%DateTime{} = the_time) do
+    time_date = DateTime.to_date(the_time)
+    today_date = DateTime.new!(today(), ~T[00:00:00.0])
+
+    if Date.compare(time_date, today_date) == :eq do
```
geekingfrog, line 24: I believe you don't need a datetime for `today_date`, `today()` should be enough.

PR #1216, opened 2026-05-28
feat: validate lobby name length

taekwombo:
Adds lobby name length validation to consul commands and spring protocol. https://github.com/beyond-all-reason/teiserver/issues/1001

#### c.battle.update_lobby_title
Decided to go with NO response for c.battle.update_lobby_title due to [_OnNo](https://github.com/beyond-all-reason/BYAR-Chobby/blob/544addacf844202a0b522df5cf9aba53e80d4841/libs/liblobby/lobby/interface.lua#L2369) handler in lobby - at least error will be visible (if used in the future).
It appears this command is not being issued (at least when [searching in bar organization](https://github.com/search?q=org%3Abeyond-all-reason%20update_lobby_title&type=code)).

### DIFF
--- a/lib/teiserver/lobby/libs/lobby_lib.ex
+++ b/lib/teiserver/lobby/libs/lobby_lib.ex
@@ -616,23 +616,18 @@ defmodule Teiserver.Lobby.LobbyLib do
     String.length(name) <= max_name_length()
   end
 
-  @err_name_empty "Lobby name must not be empty"
-  @err_name_too_long "Lobby name too long"
-  @err_name_chars_invalid "Lobby name contains invalid characters"
-  @err_name_not_allowed "Lobby name rejected"
-
   @spec validate_name(String.t()) :: :ok | {:error, String.t()}
   @spec validate_name(String.t(), boolean) :: :ok | {:error, String.t()}
   def validate_name(name, add_hints \\ false) do
     cond do
       name == "" ->
-        {:error, @err_name_empty}
+        {:error, "Lobby name must not be empty"}
 
       not name_length_valid?(name) ->
         message =
           validation_error(
             add_hints,
-            @err_name_too_long,
+            "Lobby name too long",
             "Maximum #{max_name_length()} characters allowed."
           )
 
@@ -642,7 +637,7 @@ defmodule Teiserver.Lobby.LobbyLib do
         message =
           validation_error(
             add_hints,
-            @err_name_chars_invalid,
+            "Lobby name contains invalid characters",
             "Allowed characters: alphanumeric, spaces, [, ], <, >, -, +, |, :."
           )
 
@@ -652,7 +647,7 @@ defmodule Teiserver.Lobby.LobbyLib do
         message =
           validation_error(
             add_hints,
-            @err_name_not_allowed,
+            "Lobby name rejected",
             "Please be aware that misuse of the lobby naming system can cause your chat privileges to be revoked."
           )
 
--- a/lib/teiserver/lobby/libs/lobby_lib.ex
+++ b/lib/teiserver/lobby/libs/lobby_lib.ex
@@ -4,6 +4,7 @@ defmodule Teiserver.Lobby.LobbyLib do
   alias Phoenix.PubSub
   alias Teiserver.Account
   alias Teiserver.Battle.LobbyServer
+  alias Teiserver.Chat.WordLib
   alias Teiserver.Config
   alias Teiserver.Coordinator
   alias Teiserver.Data.Types, as: T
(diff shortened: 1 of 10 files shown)


lib/teiserver/protocols/spring/spring_in.ex
```diff
@@ -845,6 +846,9 @@ defmodule Teiserver.Protocols.SpringIn do
             not Auth.is_bot?(state.userid) ->
               {:failure, "Not a bot"}
 
+            not LobbyLib.name_length_valid?(name) ->
+              {:failure, "Lobby name too long"}
```
taekwombo: Is just length validation enough? Consul additionally checks for invalid characters and banned words.
Same question applies to `c.battle.update_lobby_title` and tachyon.

lib/teiserver/protocols/spring/spring_battle_in.ex
```diff
@@ -3,18 +3,24 @@ defmodule Teiserver.Protocols.Spring.BattleIn do
...
-    if Lobby.allow?(state.userid, :update_lobby_title, state.lobby_id) do
-      Battle.rename_lobby(state.lobby_id, new_name, nil)
-      reply(:spring, :okay, "c.battle.update_lobby_title", msg_id, state)
-    else
-      state
+    cond do
+      not LobbyLib.name_length_valid?(new_name) ->
+        reply(:spring, :no, {"c.battle.update_lobby_title", "Lobby name too long"}, msg_id, state)
+
+      Lobby.allow?(state.userid, :add_bot, state.lobby_id) ->
```
L-e-x-o-n: You replaced `:update_lobby_title` with `:add_bot`.

L-e-x-o-n (commented):
Name length configuration should be read from site configs `Teiserver.TeiserverConfigs` (lib/teiserver/libs/teiserver_configs.ex), lobby section.

lib/teiserver/protocols/spring/spring_in.ex
L-e-x-o-n: It's not an issue in this case (opening a Spring lobby) because only bots can open lobbies and all bots are owned by BAR and configured to set valid names.
It is necessary for `c.battle.update_lobby_title` and Tachyon.

L-e-x-o-n:
`c.battle.update_lobby_title` might be unused by Chobby, afaik it currently uses a $rename command when editing the lobby name through chobby

lib/teiserver/protocols/spring/spring_in.ex
taekwombo: Right, so basically checking name for OPENBATTLE is not really needed. But rather full verification is needed for user facing handlers (tachyon, c.battle), right?

And it makes sense to create single method in `LobbyLib` to verify lobby name: length, chars, banned words at once. Can I freely use `Chat.WordLib` in `Lobby.LobbyLib`? I'm wondering if there's some good practices around modules called Libs that this approach would break.

L-e-x-o-n: > Right, so basically checking name for OPENBATTLE is not really needed. But rather full verification is needed for user facing handlers (tachyon, c.battle), right?

Yes, but you might as well use the same logic that verifies everything for consistency. (You mention Tachyon, which current changes don't touch, I suggest keeping that for a separate PR.)

> And it makes sense to create single method in LobbyLib to verify lobby name: length, chars, banned words at once. Can I freely use Chat.WordLib in Lobby.LobbyLib? I'm wondering if there's some good practices around modules called Libs that this approach would break.

Yes

taekwombo: I added new method to LobbyLib, it will also check for flagged words using WordLib in https://github.com/beyond-all-reason/teiserver/pull/1216/commits/0d5bc7ad699d3e3d15e66190393073b4b9cc5b90 .

It will change messages sent out by consul a bit - I tried to avoid it but I couldn't figure out nice way to achieve compatibility.

lib/teiserver/lobby/libs/lobby_lib.ex
```diff
@@ -595,4 +596,78 @@ defmodule Teiserver.Lobby.LobbyLib do
...
+
+  @spec name_length_valid?(String.t()) :: boolean
+  defp name_length_valid?(name) do
+    String.length(name) <= max_name_length()
+  end
+
+  @err_name_empty "Lobby name must not be empty"
+  @err_name_too_long "Lobby name too long"
+  @err_name_chars_invalid "Lobby name contains invalid characters"
+  @err_name_not_allowed "Lobby name rejected"
```
L-e-x-o-n: Minor thing, but I would remove these constants here and use the string values directly, I don't think they help much with readability in this case and it's a bit inconsistent when next to hints.

L-e-x-o-n (approved):

L-e-x-o-n:
Thanks

PR #1222, opened 2026-05-30
Casesensitive and match whole word for admin/chat

### LINKED ISSUE #1131: Add match case and match whole word options for chat filter
The chat filter at `/admin/chat` currently allows filtering by a term but it is always case insensitive and matching any part of the word. Additional options for case sensitive and whole word filtering would be useful.

BElluu:
Closed #1131 

I changed `core_components.ex` because the shared checkbox rendered with `value={@value`}, so its submit value tracked the field's boolean. HEEx turns that into a boolean attribute - `false` drops the value (checked box submits "on"), `true` makes it bare (submits "") - so it never submits "true" and Ecto's boolean cast rejects it. Result: the new whole_word/case_sensitive checkboxes couldn't be enabled from the form. Fixed by using the standard Phoenix `value="true"`. The only other caller already passes `value="true"` explicitly, so it's unaffected.

AI usage: This PR was developed with AI assistance (Claude Code), used for codebase exploration and tests implementation


L-e-x-o-n:
Unfortunately it looks like there was some misunderstanding with the linked issue.
It was created while banned phrases weren't a thing and it was intended for the lobby chat filter at `/admin/chat` specifically, I will update it now to clarify.

Having the case sensitive logic for banned phrases seems unnecessary IMO,  I can't think of a phrase that would or wouldn't be banned depending on the capitalisation,  Unsure about whole word, I am interested in what @Teifion thinks.

Teifion:
I think case-insensitivity would be an appropriate solution and changing BannedPhrase to be case insensitive would make sense.

BElluu:
Now, should be better. 

Without any option:
image

Case sensitive:
image

Whole word:
image

Both:
image

Additionaly, I made changes to case-insensitivity for BannedPhrases.

lib/teiserver_web/components/core_components.ex
```diff
@@ -327,8 +327,9 @@ defmodule TeiserverWeb.CoreComponents do
         id={@id}
         class="form-check-input"
         type="checkbox"
-        value={@value}
+        value="true"
```
Teifion: I believe this component is used in a lot of places and changing it here could have knock-on effects elsewhere. I would rather it wasn't changed as part of this PR

lib/teiserver/chat/libs/term_search.ex
```diff
@@ -0,0 +1,61 @@
...
+  @spec content_filter(String.t() | nil, keyword()) :: Ecto.Query.dynamic_expr() | nil
+  def content_filter(term, opts \\ [])
+  def content_filter(term, _opts) when term in [nil, ""], do: nil
+
+  def content_filter(term, opts) do
+    case_sensitive = Keyword.get(opts, :case_sensitive, false)
+    whole_word = Keyword.get(opts, :whole_word, false)
+
+    if whole_word do
+      pattern = "\\y" <> regex_term(term) <> "\\y"
```
Teifion, line 37: I've done some searching and can't find what `\\y` would do.

lib/teiserver/chat/libs/term_search.ex
```diff
@@ -0,0 +1,61 @@
...
+        dynamic([m], ilike(m.content, ^like_term))
+      end
+    end
+  end
+
+  # Escapes POSIX regex metacharacters in the term, then restores the `*`
+  # wildcard as `.*` so it keeps matching any number of characters.
+  defp regex_term(term) do
+    term
+    |> String.replace(~r/[.^$*+?()\[\]{}|\\]/, "\\\\\\0")
```
Teifion: Why not use [Regex.escape/1](https://elixir.hexdocs.pm/Regex.html#escape/1)?

test/teiserver/chat/libs/term_search_test.exs
```diff
@@ -0,0 +1,78 @@
...
+      })
+
+    msg
+  end
+
+  defp contents(search) do
+    Chat.list_lobby_messages(search: search, limit: 100)
+    |> Enum.map(& &1.content)
+    |> Enum.sort()
+  end
```
Teifion: I would prefer the setup functions be at the bottom of the file, the tests themselves are always the most interesting part of the test file.

Optionally for bonus points create a ChatFixtures module for creating a message fixture.

Teifion (commented):
A few questions/comments but I think this shows promise :)

lib/teiserver/chat/libs/term_search.ex
BElluu: In PostgreSQL (which this query is likely sent to), `\y` represents a **word boundary**. It works exactly like `\b` in standard regex libraries (like PCRE or JavaScript).

Because `\` is an escape character in Elixir, `"\\y"` compiles down to a literal `\y` in the final regex string.

### Example:
If `term` is `"cat"`, the resulting regex will be `\ycat\y`. 
* It **will match**: `"a cat sits here"` or `"the cat."`
* It **will NOT match**: `"catapult"` or `"bobcat"`, because "cat" is part of a larger word.
https://www.postgresql.org/docs/current/functions-matching.html

Teifion: Thanks, that's helpful and answers my question.

Why do you say "likely" sent to though?

BElluu: I used AI to format my answer, which was a rough draft, to format it legibly for markdown, and the AI genius cracked the case that it was most likely Postgres :D That's where the "likely" comes from. Sorry I didn't notice and introduced some senseless noise.

Teifion: I appreciate the honesty. I would much rather converse with someone than their LLM, otherwise what separates you from a different person? Also, if everything I say is just going into the LLM then nothing is being learned and internalised.

I'm happy to work with you and be patient if English isn't your first language if it means I'm speaking to an actual person.

BElluu: Sure thank you. Yea English isn't my first language. I learn English years ago but I don't use it - just when I'm on vacation or reading docs, so you know, its not perfect :) But, I have 121 days series on Duolingo, so will be better :)

Teifion: Your English is better than I can speak any other language so rest assured there is no judgement from me on that. Looking forwards to the rest of the comments :)

lib/teiserver_web/components/core_components.ex
BElluu: Ok I reverted that and adapted my code but I think, we should create a issue to globally fix it for the future :)

lib/teiserver/chat/libs/term_search.ex
BElluu: Coz I'm Elixir dumb. Fixed :)

test/teiserver/chat/libs/term_search_test.exs
BElluu: Created ChatFixture and moved funtions on the bottom of file.

lib/teiserver/chat/libs/term_search.ex
```diff
@@ -51,11 +52,9 @@ defmodule Teiserver.Chat.TermSearch do
     end
   end
 
-  # Escapes POSIX regex metacharacters in the term, then restores the `*`
-  # wildcard as `.*` so it keeps matching any number of characters.
   defp regex_term(term) do
     term
-    |> String.replace(~r/[.^$*+?()\[\]{}|\\]/, "\\\\\\0")
+    |> Regex.escape()
     |> String.replace("\\*", ".*")
```
Teifion, line 58: Why are we needing to do this replacement after the `Regex.escape`?

lib/teiserver_web/components/core_components.ex
Teifion: Agreed, feel free to create the issue on that :)

Teifion:
The changes you have made address my comments and have created only one more. Once that is done I think we can probably merge it.

lib/teiserver/chat/libs/term_search.ex
BElluu: Because `Regex.escape()` add `\` before every special character. `String.replace("\\*", ".*")` will revert this operation for `*` character which is wildcard.

For example:
When we want find any word starting with `cat` (like cat, catapult, catastrophe). So we type `cat*`
`Regex.escape()` will transform to `cat\\*` so our wildcard is broken.
`String.replace("\\*", ".*")` fix our word and we have `cat.*`

lib/teiserver_web/components/core_components.ex
BElluu: #1239

Teifion (approved):
I'm happy with the changes made, thank you. Will give the others a chance for a quick look but if no response I will merge it later today.

lib/teiserver/chat/libs/term_search.ex
```diff
@@ -0,0 +1,60 @@
...
+
+  @doc """
+  Builds a dynamic WHERE expression matching a message's `content` against the
+  given term, honouring the `:case_sensitive` and `:whole_word` options.
+
+  Returns `nil` when the term is blank, meaning no filtering should be applied.
+  The dynamic references the first query binding positionally, so it works with
+  any of the message schemas.
+  """
+  @spec content_filter(String.t() | nil, keyword()) :: Ecto.Query.dynamic_expr() | nil
```
geekingfrog: I think the API isn't great there. returning either a query or nil lead to a ton of check on the caller side.
A better design would be to take a query as the first argument, and return another query. If no filtering is required, you return the original query, otherwise you modify the object.

This way you can create the queries with a "fluent builder" like pattern:

```elixir
query = base_query()
|> content_filter(filter)
|> limit(20)
|> # whatever
```

lib/teiserver_web/live/admin/chat/index.html.heex
```diff
@@ -80,16 +110,17 @@
...
-            <th :if={@filters["mode"] == "Direct"}>Sender</th>
-            <th :if={@filters["mode"] == "Direct"}>Receiver</th>
-            <th :if={@filters["mode"] != "Direct"}>Sender</th>
+            <th :if={@filters["mode"] != "Direct"} style="width: 15%;">Location</th>
+            <th :if={@filters["mode"] == "Direct"} style="width: 15%;">Sender</th>
+            <th :if={@filters["mode"] == "Direct"} style="width: 15%;">Receiver</th>
+            <th :if={@filters["mode"] != "Direct"} style="width: 15%;">Sender</th>
             <th>Message</th>
-            <th>When</th>
+            <th style="width: 15%;">When</th>
```
geekingfrog: have you checked that on a narrow screen (like mobile)? hardcoding the width like that doesn't feel right

geekingfrog (approved):

BElluu:
@Teifion our green frog accepted, so I think we can do it :D

PR #1223, opened 2026-05-30
Fix mix assets.deploy under MIX_ENV=prod

### LINKED ISSUE #238: mix assets.deploy doesn't work with MIX_ENV=prod
Because that doesn't work with `MIX_ENV=prod`, it means that all dependencies and source code have to be compiled *twice* once in dev mode and once in prod mode to create a complete release.

BElluu:
Closed #238 

AI Used
Claude Code for research about compiling elixir

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -47,19 +47,13 @@ RUN mix deps.get
 RUN mkdir config
 COPY config/config.exs config/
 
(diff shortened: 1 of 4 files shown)


geekingfrog:
spot tested that on the dev server, seems to work fine.

PR #1226, opened 2026-06-01
Oauth authorization fixes

geekingfrog:
Two fixes there:

* when a user requests an authorization code, they can now specify which scopes they want and that will be honored. Instead of always putting the scopes of the application itself.

* when the application is already authorized, any new request for an authorization code will be automatically granted (when request is valid), skipping the authorization screen.


lib/teiserver_web/controllers/o_auth/authorize_controller.ex
```diff
@@ -9,41 +9,50 @@ defmodule TeiserverWeb.OAuth.AuthorizeController do
...
+         {:ok, redir_uri} <- OAuth.get_redirect_uri(app, Map.get(conn.params, "redirect_uri")) do
+      authorized_apps = OAuth.list_authorized_applications(conn.assigns.current_user.id)
+
+      reject_uri =
+        error_redirect_uri(
+          conn,
+          redir_uri,
+          "access_denied",
+          "user refused to authorize application"
+        )
```
L-e-x-o-n: Used only in the else block so it can be moved there.

PR #1229, opened 2026-06-01
Documented and tested external IP check task

Teifion:
In preparation for making use of the external IP check task I wanted to ensure it was correctly tested as well as allowing for different response formats via the plugin system.

### DIFF
--- a/lib/teiserver/moderation/tasks/external_ip_check_task.ex
+++ b/lib/teiserver/moderation/tasks/external_ip_check_task.ex
@@ -1,35 +1,67 @@
 defmodule Teiserver.Moderation.Tasks.ExternalIPCheckTask do
   @moduledoc """
-  Downloads the IP API data
+  Downloads the IP API data.
+
+  Data is expected to arrive in JSON format with keys in the body
+  for things like "is_proxy". Any missing will just be ignored.
+
+  If the check is not enabled/configured you will receive a list with
+  one atom saying why the check did not take place. This is intended
+  to help with debugging on a production system if the checks
+  are not coming back with valid data.
   """
   alias Req.Response
   alias Teiserver.Config
+  alias Teiserver.Plugins
+
+  use Plugins
 
   require Logger
 
+  @spec query_ip(String.t()) :: [atom()]
   def query_ip(ip) do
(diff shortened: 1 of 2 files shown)


lib/teiserver/moderation/tasks/external_ip_check_task.ex
```diff
@@ -1,35 +1,67 @@
...
+        {:error, error} ->
+          Logger.error("Failed to lookup ip #{ip} - #{inspect(error)}")
+          []
+      end
+    else
+      [:not_configured]
     end
   end
+
+  @decorate Plugins.plugin(:parse_external_ip_response)
```
geekingfrog, line 50: Under what circumstance you expect to have different parsing of the response?

lib/teiserver/moderation/tasks/external_ip_check_task.ex
```diff
@@ -1,35 +1,67 @@
...
+      case Req.get(endpoint, params: [q: ip, key: key]) do
+        {:ok, %Response{body: body}} ->
+          parse_body(body)
+
+        {:error, error} ->
+          Logger.error("Failed to lookup ip #{ip} - #{inspect(error)}")
+          []
+      end
+    else
+      [:not_configured]
```
geekingfrog, line 46: That seems odd. When there is an error you return an empty list, but when disabled it's a special atom? (same remark for the `:not_enabled` bit.
Why not always return `[]`, or, instead of magic values, `{:error, :not_enabled | :not_configured | req_error}`, or even exceptions?

geekingfrog (commented):
I am not a fan of the plugin logic and exposing the parsing of body.
To me, this module should produce an interface which looks like `ip in string format -> array of properties`, and anything that goes into this should be an implementation detail of the module itself.

It is conceivable that this check is done against an in-memory dump/cache, or a database, of a flat file imported somehow. For the caller of this module these change are irrelevant, but if you expose the bit about "parsing a body" then you are tied to http.

Also, I'm not convinced about having a dynamically configurable endpoint, do we have different API that still return the same body? If not, I'd say the endpoint is part of the ip check module itself and shouldn't be configurable.

geekingfrog:
Here's an alternative approach: https://github.com/beyond-all-reason/teiserver/pull/1230

PR #1230, opened 2026-06-02
Rework ban reasons based on IP address

geekingfrog:
* Split the logic about getting info about a given IP address from the moderation related logic.

* Run a stub client in dev and test. This allow the server to run out of the box and still give some useful results locally.

The `IpApi` module isn't tested there. It's probably a good place to use a mock. An alternative would be to have a real http server as stub, but that feels way too much.


Teifion (approved):
Looks nice, a better solution than my idea.

PR #1231, opened 2026-06-02
Structify lobby data

geekingfrog:
This adds proper structs for almost all types defined for tachyon lobbies. Almost all `@type ...` have been moved to structs.
I've also done a smoke test with the new client and it still works.
There is no business logic change on this PR.

### DIFF
--- a/lib/teiserver/tachyon_lobby/types/bot.ex
+++ b/lib/teiserver/tachyon_lobby/types/bot.ex
@@ -1,0 +1,22 @@
+defmodule Teiserver.TachyonLobby.Types.Bot do
+  @moduledoc """
+  A bot in a lobby. It always has a team (a bot cannot be a spectator)
+  and is also tied to a player: their machine is used to run the AI
+  """
+
+  alias Teiserver.TachyonLobby.Types.Types, as: LT
+
+  @enforce_keys [:id, :host_user_id, :team, :name]
+  defstruct [:id, :host_user_id, :team, :name, :short_name, :version, :options]
+
+  @type id() :: String.t()
+  @type t() :: %__MODULE__{
+          id: id(),
+          host_user_id: Teiserver.Account.User.id(),
+          team: LT.team(),
+          name: String.t(),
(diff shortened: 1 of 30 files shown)


lib/teiserver/tachyon_lobby/types/bot.ex
```diff
@@ -0,0 +1,23 @@
+defmodule Teiserver.TachyonLobby.Types.Bot do
+  @moduledoc """
+  A bot in a lobby. It always has a team (a bot cannot be a spectator)
+  and is also tied to a player: their machine is used to run the AI
+  """
+
+  alias Teiserver.Data.Types, as: DataTypes
```
Teifion: I want to step away from the Types module I made, instead for userid we should use `User.id()`

lib/teiserver/tachyon_lobby/types/types.ex
```diff
@@ -0,0 +1,18 @@
+defmodule Teiserver.TachyonLobby.Types.Types do
+  @moduledoc """
+  central location for lobby related types that aren't really suited to be structs
+  """
+
+  @typedoc """
+  These represent the indices respectively into
+  {ally team index, team index, player index}
+  since we don't really support "archon mode" though, the player index
```
Teifion, line 14: I am confident we will be asked to support it; I would also expect a different name as that is SC2 specific.

Teifion (commented):
I've done a first pass as requested. Broadly I think it's good and the direction I was envisaging going. I'd probably add stucts to the function header where possible to better take advantage of the type system too.

More importantly though, how do you feel about it?

lib/teiserver/tachyon_lobby/types/types.ex
geekingfrog: the current code almost support it, adding it wouldn't really be much of a problem.

L-e-x-o-n (commented):
Nice

L-e-x-o-n (approved):

PR #1234, opened 2026-06-04
feat(tachyon): add lobby name validation for create and update

### LINKED ISSUE #1001: [Feature]: Limit the length of lobby names

taekwombo:


lib/teiserver/tachyon_lobby.ex
```diff
@@ -45,7 +46,8 @@ defmodule Teiserver.TachyonLobby do
   end
 
   def create(start_params) do
-    with {:ok, %{pid: pid, id: id}} <- TachyonLobby.Supervisor.start_lobby(start_params),
+    with :ok <- LobbyLib.validate_name(start_params.name),
+         {:ok, %{pid: pid, id: id}} <- TachyonLobby.Supervisor.start_lobby(start_params),
```
taekwombo: Updated `LobbyLib.validate_name` with keyword list so that depending on the context more appropriate message is being sent:
- `lobby/update` returns in details `"Cannot update lobby: name must not be empty"`
- `lobby/create` returns in details `"Lobby name must not be empty"` 
- spring and consul returns `"Lobby name must not be empty"` with hints for consul

My question is: should `lobby/create` align with `lobby/update` error format - `"Cannot create lobby: #{error}"`?

taekwombo:
And next question: should I add tests for all changes?
I have uncommited tests for `lobby/create`, `lobby/update` and consul commands but it feels like it's the same test all over again.

geekingfrog:
Definitely needs test.
tachyon and spring are completely separate, so yes, it seems that it's the same test, because it largely is, but for very different systems.

L-e-x-o-n (commented):
Why the need for subject, it seems unnecessary, I would always use either one or the other.

taekwombo:
> Why the need for subject, it seems unnecessary, I would always use either one or the other.

Well, I wanted it to fit in nicely for both consul and api error formats. But you're right - it's easier to keep them simple and change the usage later if needed.

test/support/fixtures/lobby.ex
```diff
@@ -0,0 +1,33 @@
...
+  end
+
+  def invalid_names do
+    [
+      invalid_name(:empty),
+      invalid_name(:too_long),
+      invalid_name(:invalid_char),
+      invalid_name(:flagged)
+    ]
+  end
```
geekingfrog: bleh, that's kind of convoluted. I'd rather have functions that directly return the strings, like
```elixir
def name_too_long(), do: ...
def illegal_char(), do: ...
```
the empty one also doesn't need a function for that

But also, I don't think this file makes sense at all. It barely saves any time and creates another layer of indirection for things that are, overall, rather trivial.
These could instead be test on their own when checking only name validation logic, in pure functions.

Like
```elixir
test "reject empty name" do
  {:error, _msg} = LobbyLib.validate_name("")
end

test "reject name too long" do
  name = ...
  {:error, _msg} = LobbyLib.validate_name(name)
end
# and so on
```

test/teiserver/coordinator/consul_commands_test.exs
```diff
@@ -0,0 +1,58 @@
...
+        Coordinator.send_consul(lobby_id, %{command: "rename", senderid: host.id, remaining: name})
+      end)
+
+      Enum.each(invalid_names, fn name ->
+        {:error, message} = LobbyLib.validate_name(name, hints: true)
+        assert_receive(%{channel: ^channel, event: :announce, message: ^message}, 2000)
+      end)
+
+      PubSub.unsubscribe(Teiserver.PubSub, channel)
+    end
```
geekingfrog, line 72: This test kind of conflate 2 things: name validation and sending a message when invalid.
I think these 2 things should be tested separately. You don't care about every edge case of name validation when you just want to check that a message was sent.
And conversely, you also don't care about lobbies and processes and pubsub when checking the names.

test/teiserver/lobby/libs/lobby_lib_test.exs
```diff
@@ -3,20 +3,23 @@ defmodule Teiserver.Lobby.LobbyLibTest do
...
+      assert {:error, _msg} = invalid_name(:too_long) |> LobbyLib.validate_name()
     end
 
     test "rejects name with invalid characters" do
-      assert {:error, _msg} = LobbyLib.validate_name("invalid char: =")
+      assert {:error, _msg} = invalid_name(:invalid_char) |> LobbyLib.validate_name()
+    end
+
+    test "rejects name with flagged words" do
+      assert {:error, _msg} = invalid_name(:flagged) |> LobbyLib.validate_name()
```
geekingfrog: see my comment about fixtures, you should just inline the names there, it's much clearer what's going on.

test/teiserver/protocols/spring/spring_battle_host_test.exs
```diff
@@ -452,10 +460,16 @@ defmodule Teiserver.SpringBattleHostTest do
     ignore_events(socket)
     assert reply =~ "BATTLEOPENED"
 
-    _send_raw(socket, "c.battle.update_lobby_title #{invalid_name}\n")
-    reply = _recv_until(socket)
+    # Updating lobby with c.battle.update_lobby_title returns NO response for each invalid name
```
geekingfrog: Same, you don't care about each possible failure, just test that you get a `NO` for one invalid name and call it a day

test/teiserver_web/tachyon/lobby_test.exs
```diff
@@ -50,6 +51,20 @@ defmodule TeiserverWeb.Tachyon.LobbyTest do
...
+      Enum.each(LobbyFixtures.invalid_names(), fn name ->
+        data = %{
+          name: name,
+          map_name: "test-map",
+          ally_team_config: Tachyon.mk_ally_team_config(2, 1)
+        }
+
+        response = Tachyon.create_lobby!(client, data)
+
+        assert %{"status" => "failed", "details" => "Cannot create lobby:" <> _message} = response
```
geekingfrog: You also should check `"reason" => "invalid_request"`

test/teiserver_web/tachyon/lobby_test.exs
```diff
@@ -50,6 +51,20 @@ defmodule TeiserverWeb.Tachyon.LobbyTest do
       %{"status" => "failed", "reason" => "invalid_request", "details" => "already_in_lobby"} =
         Tachyon.create_lobby!(client, lobby_data2)
     end
+
+    test "cannot create lobby with invalid name", %{client: client} do
+      Enum.each(LobbyFixtures.invalid_names(), fn name ->
```
geekingfrog: same comment, one bad name is enough

test/teiserver_web/tachyon/lobby_test.exs
```diff
@@ -510,6 +525,14 @@ defmodule TeiserverWeb.Tachyon.LobbyTest do
       assert last_vote["outcome"] == "passed"
       assert last_vote["vote"] == %{"type" => "changeMap", "newMapName" => "new map name"}
     end
+
+    test "cannot update lobby with invalid name", %{client: client} do
+      Enum.each(LobbyFixtures.invalid_names(), fn name ->
```
geekingfrog: you can ditch this test completely.
Instead of doing that through the websocket which is rather heavy, you can instead test that through the internal API, see [this file](https://github.com/beyond-all-reason/teiserver/blob/main/test/teiserver/tachyon_lobby/lobby_test.exs)

test/teiserver/coordinator/consul_commands_test.exs
taekwombo: Should be a bit better now for consul commands. There is now a test to see if rename command works, a test for lobby chat message in case name is incorrect.

But for the case we want to check that name does not change I couldn't find a way to do so without `:timer.sleep` so I didn't include it - I wasn't sure if it's ok to rely on timer only.

geekingfrog (commented):
looks fine, just need to resolve the merge conflicts

geekingfrog (approved):

PR #1235, opened 2026-06-04
Smtp dev setup

geekingfrog:
Setup a container for smtp test.

Also fixed a small problem when sending an email fails and that would cause an error with ecto.

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -1,0 +1,15 @@
+services:
+
+  # dev version of an smtp server so we can easily test sending emails
+  smtp:
+    image: "axllent/mailpit"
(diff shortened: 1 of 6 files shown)


Teifion:
@L-e-x-o-n is there a way we can test a specific branch on Integration before merge? If so I would like to see this working on Integration as part of the review :)

Aside from that it looks very sensible and I would approve it.

geekingfrog:
The repo is also cloned on the integration server. You can checkout the branch there, then build and deploy that

geekingfrog:
Just deployed that and tested a password reset on integration and it worked.

L-e-x-o-n (approved):

PR #1236, opened 2026-06-04
Added some uses of moderation automation

Teifion:
- WordLib blacklisted phrases now uses BannedPhrases so can be configured at runtime
- User registration makes use of external IP check functionality
- Fixed noisy log with external IP check test

### DIFF
--- a/test/teiserver/moderation/tasks/external_ip_check_task_test.exs
+++ b/test/teiserver/moderation/tasks/external_ip_check_task_test.exs
@@ -7,6 +7,8 @@ defmodule Teiserver.Moderation.Tasks.ExternalIpCheckTaskTest do
 
   use Teiserver.DataCase, async: false
 
+  import ExUnit.CaptureLog
+
   describe "return ban reason when config enabled" do
     test "abuser" do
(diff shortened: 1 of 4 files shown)


test/teiserver/moderation/tasks/external_ip_check_task_test.exs
```diff
@@ -50,7 +52,12 @@ defmodule Teiserver.Moderation.Tasks.ExternalIpCheckTaskTest do
   end
 
   test "doesn't raise on error" do
-    assert ExternalIPCheckTask.get_ban_reasons(Stub.error_ip()) == []
+    {result, _log} =
+      with_log(fn ->
+        ExternalIPCheckTask.get_ban_reasons(Stub.error_ip())
+      end)
```
L-e-x-o-n, line 58: What's the purpose of `with_log` here? The log isn't used.

Teifion: To prevent the log being spat out into the console during tests. We know the log is output and we want to ignore it for the test.

PR #1240, opened 2026-06-05
Add telemetry span around message_severity function

Teifion:
This is a function potentially being called a lot of times with an expensive cost, this would give us an idea of how much and at what processing cost we are looking at.

I thought about wrapping the `phrase_match?` function to get a per phrase metric too but wanted to keep it light and limited for now.

### DIFF
--- a/lib/teiserver/monitoring/internal.ex
+++ b/lib/teiserver/monitoring/internal.ex
@@ -1,0 +1,51 @@
+defmodule Teiserver.Monitoring.Internal do
+  @moduledoc false
+  use PromEx.Plugin
+
+  @impl PromEx.Plugin
+  def event_metrics(_opts) do
+    Event.build(
+      :teiserver_internal_event_metrics,
+      [
+        distribution(
+          [:teiserver, :moderation, :message_severity, :stop],
+          event_name: [:moderation, :message_severity],
+          measurement: :duration,
+          reporter_options: [
+            buckets: [
+              1,
+              2,
+              3,
+              5,
+              8,
+              13,
+              21,
+              34,
+              55,
+              89,
+              144,
+              233,
+              377,
+              610,
+              987,
+              1597,
+              2584,
+              4181
+            ]
+          ],
+          tags: [:matched]
+        ),
+        counter(
+          [:teiserver, :moderation, :message_severity, :stop],
+          event_name: [:moderation, :message_severity],
+          measurement: :count
+        )
+      ]
+    )
+  end
+
+  @impl PromEx.Plugin
+  def polling_metrics(_opts) do
+    []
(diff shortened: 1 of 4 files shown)


lib/teiserver/monitoring/internal.ex
```diff
@@ -0,0 +1,51 @@
...
+              233,
+              377,
+              610,
+              987,
+              1597,
+              2584,
+              4181
+            ]
+          ],
+          tags: [:matched]
```
geekingfrog, line 36: what's this `tags` thingy?

lib/teiserver_web/telemetry.ex
```diff
@@ -48,7 +48,18 @@ defmodule TeiserverWeb.Telemetry do
...
+      # Moderation
+      distribution(
+        [:moderation, :message_severity],
+        event_name: [:teiserver, :moderation, :message_severity, :stop],
+        measurement: :duration,
+        reporter_options: [
+          buckets: [1, 2, 3, 5, 8, 13, 21, 34, 55, 89, 144, 233, 377, 610, 987, 1597, 2584, 4181]
+        ],
+        tags: [:matched]
+      )
```
geekingfrog, line 62: I don't think you need this block there since you added it as a different PromEx plugin already.
It should be one or the other, and since this is not spring specific, I think it's better to leave it in the other plugin

lib/teiserver/monitoring/internal.ex
Teifion: From [temelemetry_metrics](<https://telemetry-metrics.hexdocs.pm/Telemetry.Metrics.html#module-metrics>)

> :tags - a subset of metadata keys by which aggregations will be broken down. Defaults to an empty list.9

lib/teiserver_web/telemetry.ex
Teifion: This makes it appear in the liveview dashboard too, it will help verify the events are being emitted.

geekingfrog: hum, then the dashboard should be amended, or the other distribution should be removed. you effectively have duplicate definition.
The dashboard bit though, is it still useful now we have grafana?

Teifion: I'm happy to remove it from the liveview dashboard later, at this stage I'm suggesting it purely so we can see if the result is being output to assist in debugging it maybe not appearing in Grafana.

geekingfrog (approved):
let's ship that, but there should be more work done to avoid the duplication of metrics, and also probably rethink the custom dashboards, since some can be moved to grafana (where data retention of 3 months doesn't matter)

PR #1241, opened 2026-06-06
Updading banned item now re-caches them

Teifion:
Currently creating, updating or deleting a banned item (IP, Domain or Phrase) would leave you with a stale cache. This solves that and will update the cache on a successful change.

### DIFF
--- a/lib/teiserver/moderation.ex
+++ b/lib/teiserver/moderation.ex
@@ -925,7 +925,7 @@ defmodule Teiserver.Moderation do
     %BannedDomain{}
     |> BannedDomain.changeset(attrs)
     |> Repo.insert()
-    |> LoadBannedDomainsTask.maybe_recache()
+    |> LoadBannedDomainsTask.cache_if_ok()
   end
 
   @doc """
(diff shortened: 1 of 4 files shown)


lib/teiserver/moderation.ex
geekingfrog, line 1: this should be a single function used in all call sites, like `|> update_cache()` or something like that.

Teifion: Addressed in https://github.com/beyond-all-reason/teiserver/pull/1241/commits/f6621ebcc4d91057bcf721a37e901345d1d25f69

geekingfrog (commented):
maybe nitpick, but I am not a fan of names like `maybe_xxx`. I find that quite confusing. Generally I prefer `ensure_<final_state>`, so like `ensure_cache_updated`.

With a `maybe_xxx` the caller doesn't really know what's going to happen. But with `ensure_xxx` it's clearer about what calling this function will achieve.

Teifion:
> I am not a fan of names like `maybe_xxx`.

In this case I'm liking it because it may or may not update the cache. I disagree with `ensure_xxx` in this case because is not ensuring anything, if there is a failed state nothing at all should happen. I'm open to discuss other names though.

geekingfrog (approved):

PR #1243, opened 2026-06-06
Migrate Bot page to ListView

BElluu:
I saw `<%!-- TODO: add a modal confirmation for deleting an bot --%>` so I wanted add this, but Bot page wasn't LiveView, so I migrated it too :)

AI used to rewrite tests from View to LiveView :)


lib/teiserver_web/live/admin/bot/form_component.ex
```diff
@@ -0,0 +1,60 @@
...
+      :edit -> save(socket, Bot.update_bot(socket.assigns.bot, params))
+    end
+  end
+
+  defp save(socket, {:ok, bot}) do
+    send(self(), {__MODULE__, {:saved, bot}})
+
+    {:noreply,
+     socket
+     |> put_flash(:info, "Saved!")
```
Teifion: No need for an exclamation here, a simple "Bot saved correctly" would be appropriate.

lib/teiserver_web/live/admin/bot/form_component.ex
```diff
@@ -0,0 +1,60 @@
...
+     |> push_patch(to: socket.assigns.patch)}
+  end
+
+  defp save(socket, {:error, changeset}) do
```
Teifion: Please typematch on the Changeset struct

lib/teiserver_web/live/admin/bot/show.ex
```diff
@@ -0,0 +1,121 @@
...
+
+  @impl LiveView
+  def mount(_params, _session, socket) do
+    socket =
+      socket
+      |> add_breadcrumb(name: "Admin", url: "/teiserver/admin")
+      |> add_breadcrumb(name: "Bots", url: "/teiserver/admin/bot")
+      |> assign(:client_secret, nil)
+      |> assign(:confirm_delete, false)
+      |> assign(:credential_to_delete, nil)
```
Teifion: Please use the existing confirmation for deletion method rather than passing around state unless there is a specific need for it.

lib/teiserver_web/live/admin/bot/show.ex
```diff
@@ -0,0 +1,121 @@
...
+     socket
+     |> put_flash(:info, "Bot deleted")
+     |> push_navigate(to: ~p"/teiserver/admin/bot")}
+  end
+
+  @impl LiveView
+  def handle_event("create_credential", %{"application" => app_id}, socket) do
+    app = ApplicationQueries.get_application_by_id(app_id)
+    client_id = UUID.uuid4()
+    secret = 32 |> :crypto.strong_rand_bytes() |> Base.hex_encode32()
```
Teifion: I see this is used in a few other places in the application, could you create a helper function for it in an appropriate place and then an issue flagging the code duplication and refactor to use the helper function.

No need to edit the rest of the code, we want to prevent PR-scope creep.

lib/teiserver_web/live/admin/bot/show.ex
```diff
@@ -0,0 +1,121 @@
...
+
+    case OAuth.create_credentials(app, socket.assigns.bot, client_id, secret) do
+      {:ok, _cred} ->
+        {:noreply,
+         socket
+         |> put_flash(:info, "Credential created")
+         |> assign(:client_secret, secret)
+         |> assign(:credentials, CredentialQueries.for_bot(socket.assigns.bot))}
+
+      {:error, err} ->
```
Teifion: Please match on the Changeset struct

Teifion (commented):
Thank you for taking the time to convert this from a deadview to a liveview, that is really helpful. I initially had concerns about the way the form component is organised compared to the generated one used in BannedPhrases but I think this is nicer way to do it and I will be re-factoring the others to use this!

As per the discord discussion we would like to have the editing take place on the `show` page rather than the `index` page. I have not yet reviewed the tests as I suspect some of them will need to be tweaked as part of that change.

Overall promising; thank you.

lib/teiserver_web/live/admin/bot/show.ex
BElluu: I added helper. I will create new issue to reuse it in other places in code, after merge :)

```
@spec generate_client_credentials() :: {client_id :: String.t(), secret :: String.t()}
 def generate_client_credentials do
   client_id = UUID.uuid4()
   secret = 32 |> :crypto.strong_rand_bytes() |> Base.hex_encode32()
   {client_id, secret}
 end
```

Teifion:
Apologies for taking so long to get to this, I have only just had time to test it. The manual test has gone well and I will do a review of the code after dinner; I will hopefully merge it shortly after that.

BElluu:
> Apologies for taking so long to get to this, I have only just had time to test it. The manual test has gone well and I will do a review of the code after dinner; I will hopefully merge it shortly after that.

No problem mate. Take your time :)

test/teiserver_web/live/admin/bot_live_test.exs
```diff
@@ -0,0 +1,211 @@
...
+  defp create_bot(_context) do
+    bot = BotFixtures.create_bot("testing bot")
+    %{bot: bot}
+  end
+
+  defp create_app(context) do
+    owner_id = context[:user].id
+    app = OAuthFixtures.app_attrs(owner_id) |> OAuthFixtures.create_app()
+    %{app: app}
+  end
```
Teifion: Why not use the `create_app` from the Oauth fixtures?

test/teiserver_web/live/admin/bot_live_test.exs
```diff
@@ -0,0 +1,211 @@
...
+             |> render_change() =~ "should be at least"
+    end
+
+    test "rejects name too long", %{conn: conn} do
+      {:ok, live, _html} = live(conn, ~p"/teiserver/admin/bot/new")
+
+      assert live
+             |> form("#bot-form", bot: %{name: String.duplicate("a", 40)})
+             |> render_change() =~ "should be at most"
+    end
```
Teifion: These tests should be against the changeset itself, we don't need to do them via the LiveView

test/teiserver_web/live/admin/bot_live_test.exs
```diff
@@ -0,0 +1,211 @@
...
+      live
+      |> form("form[phx-submit=create_credential]", application: app.id)
+      |> render_submit()
+
+      html = render(live)
+      assert html =~ "Credential created"
+      assert html =~ "Secret only shown once!"
+
+      [_full, secret] = Regex.run(~r/Secret only shown once! <pre>([^<]+)<\/pre>/, html)
+      secret = String.trim(secret)
```
Teifion: We never want to be using a Regex with HTML if we can help it, put it inside an element and pull that element out to get the secret.

Teifion (commented):
Minor tweaks to make to the tests and then good to approve.

Teifion (approved):
Nice work.

PR #1248, opened 2026-06-07
feat(devcontainer): set up development environment devcontainer

WybrenKoelmans:


.devcontainer/startup.sh
```diff
@@ -0,0 +1,78 @@
...
+  Mail:    http://localhost:8025
+  
+  Login credentials:
+    Email:    root@localhost
+    Password: password
+
+  GitHub:
+  =======================
+  This container comes with the GitHub CLI installed. Use $ gh auth login to authenticate.
+  Make sure the remotes are configured to HTTPS (and not SSH) to avoid authentication issues.
```
geekingfrog, line 33: it does? I can't see anything in the dockerfile about that tool

.devcontainer/startup.sh
```diff
@@ -0,0 +1,78 @@
...
+  After authentication, you can use git as normal.
+
+MOTD
+
+exec > >(tee -a "$LOG") 2>&1
+
+echo "Waiting for database..."
+until pg_isready -h database -U teiserver_dev -q; do
```
geekingfrog, line 7: this should not be hardcoded but come from env vars. And also, `set -o nounset` so bash errors on unbound vars.

compose.yaml
```diff
@@ -1,15 +1,102 @@
...
+      # Discord bridge
+      #- TEI_ENABLE_DISCORD_BRIDGE=false
+      #- TEI_DISCORD_BOT_TOKEN=
+      #- TEI_DISCORD_GUILD_ID=
+      #- TEI_DISCORD_BOT_NAME=
+      # Metrics server
+      #- TEI_METRICS_SERVER_PORT=4001
+
+  database:
+    image: postgres:latest
```
geekingfrog: Don't use latest, we are on 17, and upgrade to this should be deliberate, not whatever tag this points to changes.

compose.yaml
```diff
@@ -1,15 +1,102 @@
...
 
-  # dev version of an smtp server so we can easily test sending emails
-  smtp:
+  teiserver:
+    build:
+      context: .
+      dockerfile: Dockerfile.dev
+    container_name: teiserver
+    ports:
+      - "4000:4000"
```
geekingfrog: you also want to map 8000 (spring tcp, used by chobby) and 8001 (spring tls, though this one is unused by default iirc). These are the default values.
There is also 4001 aka `TEI_METRICS_SERVER_PORT` to expose metrics.

geekingfrog (commented):
I don't use devcontainer, so can't really test that, if someone could give it a go that'd be grand.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,42 @@
...
+	// connected. This is typically a file mount in .devcontainer/docker-compose.yml
+	"workspaceFolder": "/workspaces/${localWorkspaceFolderBasename}",
+
+	// Features to add to the dev container. More info: https://containers.dev/features.
+	"features": {
+		"ghcr.io/devcontainers/features/github-cli:1": {}
+	},
+
+	// Use 'forwardPorts' to make a list of ports inside the container available locally.
+	"forwardPorts": [4000, 8025],
```
geekingfrog: also need to update this list (see follow up comment about the other ports exposed by teiserver)

.devcontainer/startup.sh
```diff
@@ -0,0 +1,78 @@
...
+mix ecto.migrate
+
+SETUP_MARKER=".devcontainer/.setup_complete"
+if [ ! -f "$SETUP_MARKER" ]; then
+  echo "Loading test data..."
+  mix teiserver.fakedata
+  touch "$SETUP_MARKER"
+fi
+
+cat > /etc/motd << 'MOTD'
```
geekingfrog, line 18: why is this MOTD section repeated?

.devcontainer/startup.sh
```diff
@@ -0,0 +1,78 @@
...
+    
+  Teiserver Dev Container
+  =======================
+
+  Server:  http://localhost:4000
+  Mail:    http://localhost:8025
+
+  Login credentials:
+    Email:    root@localhost
+    Password: password
```
geekingfrog: by default there is nothing, you need to run https://github.com/beyond-all-reason/teiserver/blob/main/documents/guides/development.md#fake-data to populate it with some defaults.
That should be displayed instead of that. Though I'm cool if you want to tackle that later

geekingfrog:
There is an [AI policy](https://github.com/beyond-all-reason/teiserver/blob/main/AI_POLICY.md) which is the same as for the main bar repo, I'm pretty sure some form of slop machine was used there (the repeated and very verbose motd as well as the github cli comment looks fishy)

.devcontainer/startup.sh
WybrenKoelmans: At this point the initial server setup is completed, and this MOTD is without the message "Server is setting up in the background."

.devcontainer/startup.sh
WybrenKoelmans: see .devcontainer/devcontainer.json:22

compose.yaml
WybrenKoelmans: ok

compose.yaml
WybrenKoelmans: yep

.devcontainer/startup.sh
WybrenKoelmans: it's fine to be hardcoded

.devcontainer/startup.sh
WybrenKoelmans: it runs as part of the initial setup script (see line 53)

.devcontainer/devcontainer.json
WybrenKoelmans: yep

WybrenKoelmans:
>There is an [AI policy](https://github.com/beyond-all-reason/teiserver/blob/main/AI_POLICY.md) which is the same as for the main bar repo, I'm pretty sure some form of slop machine was used there (the repeated and very verbose motd as well as the github cli comment looks fishy)

Thank you for your concern about AI usage. I've addressed your comments. Feel free to close.

.devcontainer/startup.sh
ChocoboCoding: Why are we using github CLI instead of cloning via SSH?
The latter would allow us to clone with a configured ssh key, instead of requiring a login. Cloning via ssh using a token would allow for automating things better, instead of logging in manually

WybrenKoelmans: It's for developers that are not experienced with that, running that command and clicking a link in the terminal to authenticate with GitHub is a lot less friction that setting up SSH keys.

...and this doesn't exclude it, people can still setup their SSH key (a new one, as the container does not have access to the host ssh keys!) or forward their SSH agent if they want.

.devcontainer/startup.sh
geekingfrog: It's also fine to have env var, you are already in a context where these are set anyway.

.devcontainer/setup.sh
```diff
@@ -0,0 +1,9 @@
+echo "Installing dependencies..."
+mix deps.get
+
+echo "Setting up database..."
+mix ecto.create
+mix ecto.migrate
```
geekingfrog: you can run `mix setup` which is an alias for these 3 steps.

.devcontainer/startup.sh
geekingfrog: ah indeed, missed that file

geekingfrog:
> Thank you for your concern about AI usage. I've addressed your comments. Feel free to close.

You did not answer my question.
It's not about closing anything with AI on sight, if I do QA for a slopmachine I'd like to know.

geekingfrog:
I think you should also add a small section in the [development guide](https://github.com/beyond-all-reason/teiserver/blob/main/documents/guides/development.md) about this devcontainer

.devcontainer/setup.sh
WybrenKoelmans: done

WybrenKoelmans:
> You did not answer my question.
It's not about closing anything with AI on sight, if I do QA for a slopmachine I'd like to know.

Well.. if you insist I will quote the AI policy
> The above requirements apply to cases where AI is used to write production code.

Since this doesn't touch a single line of production code I don't need to disclose anything lol. Your "AI suspicions" were a little weak imo as they seemed to be oversights by you (in fact they were all handwritten). 

For my last year of contributing I stand behind the code of each line of code contributed. Frankly these [AI accusations](https://github.com/beyond-all-reason/Beyond-All-Reason/pull/7777#issuecomment-4563023668) are a little annoying and that's why I say feel free to close anything you don't trust. And even if you don't trust a PR, a simple "Would you mind disclosing any AI usage?" is enough, a little courtesy goes a long way.

> I think you should also add a small section in the [development guide](https://github.com/beyond-all-reason/teiserver/blob/main/documents/guides/development.md) about this devcontainer

Good idea. I've added a section. Note that I kept it simple on purpose, as more advanced users should scroll down and do a more advanced setup as described there.

geekingfrog:
In the future, I would appreciate if you could avoid the rule lawyering or snark, that attitude is not ok.
The AI stuff is part about trust, but also because reviewing ai code is different from reviewing human authored stuff.

PR #1250, opened 2026-06-08
Replace bamboo with swoosh

geekingfrog:
The repository bamboo_smtp has been archived. It depends on a version of `gen_smtp` that doesn't compile on otp 28.
So it's simpler to move to swoosh, which has a very similar API, and a more maintained ecosystem. Including swoosh smtp adapter that depends on a newer version of gen_smtp which works with otp 28.

probably needs a smoke test on integration server, locally it does work:
2026-06-08-1361x153-scrot


lib/teiserver/common/email_helper.ex
```diff
@@ -35,10 +35,10 @@ defmodule Teiserver.EmailHelper do
 
   def send_password_reset(user, code \\ nil) do
     {code, email} = password_reset(user, code)
-    response = Mailer.deliver_now(email, response: true)
+    response = Mailer.deliver(email, response: true)
 
     case response do
-      {:ok, _email, _response} ->
```
L-e-x-o-n, line 41: What is this `_x` actually? Doing it this way just defeats the purpose of the Credo check we have for `UnusedVariableNames`.

geekingfrog: Here's an example of response: `"2.0.0 Ok: queued as 04r3Urxr90dDscJTrnGlsh\r\n"`
I amended to `_resp`

lib/teiserver/common/email_helper.ex
```diff
@@ -12,7 +12,7 @@ defmodule Teiserver.EmailHelper do
     case Config.get_site_config_cache("teiserver.Require email verification") do
       true ->
         case do_new_user(user) do
-          {:ok, _email, _response} ->
+          {:ok, _x} ->
```
L-e-x-o-n: Here as well
```suggestion
          {:ok, _resp} ->
```

PR #1252, opened 2026-06-08
Kick and ban support from lobby

### LINKED ISSUE #1141: add support for kick&ban from lobby
protocol: https://github.com/beyond-all-reason/tachyon/pull/135

BElluu:
Closed #1141

### DIFF
--- a/lib/teiserver/player/tachyon_handler.ex
+++ b/lib/teiserver/player/tachyon_handler.ex
@@ -1026,11 +1026,11 @@ defmodule Teiserver.Player.TachyonHandler do
     target_id = data["userId"]
     ban_until = parse_ban_until(data["banUntil"])
 
-      case Session.lobby_kickban(state.user.id, target_id, ban_until) do
-        :ok -> {:response, state}
-        {:error, :unauthorized} -> {:error_response, :unauthorized, state}
-        {:error, reason} -> {:error_response, :invalid_request, to_string(reason), state}
-      end
+    case Session.lobby_kickban(state.user.id, target_id, ban_until) do
+      :ok -> {:response, state}
+      {:error, :unauthorized} -> {:error_response, :unauthorized, state}
+      {:error, reason} -> {:error_response, :invalid_request, to_string(reason), state}
+    end
   end
 
   def handle_command(_command_id, _message_type, _message_id, _message, state) do
@@ -1458,10 +1458,11 @@ defmodule Teiserver.Player.TachyonHandler do
   end
 
   defp parse_ban_until(nil), do: nil
+
   defp parse_ban_until(ts) do
     case DateTime.from_unix(ts, :microsecond) do
       {:ok, dt} -> if DateTime.compare(dt, DateTime.utc_now()) == :gt, do: dt, else: nil
-      _ -> nil
+      _error -> nil
     end
   end
 end
--- a/lib/teiserver/player/tachyon_handler.ex
+++ b/lib/teiserver/player/tachyon_handler.ex
@@ -246,6 +246,16 @@ defmodule Teiserver.Player.TachyonHandler do
     {:event, "lobby/left", %{id: lobby_id, reason: reason}, state}
   end
 
+  def handle_info({:lobby, lobby_id, {:left, reason, ban_until}}, state) do
+    data =
+      case ban_until do
+        nil -> %{id: lobby_id, reason: reason}
+        dt -> %{id: lobby_id, reason: reason, bannedUntil: DateTime.to_unix(dt, :microsecond)}
+      end
+
+    {:event, "lobby/left", data, state}
+  end
+
   def handle_info({:lobby_list, {:add_lobby, lobby_id, %LT.ListOverview{} = overview}}, state) do
     data = %{lobbies: %{lobby_id => lobby_overview_to_tachyon(lobby_id, overview)}}
     {:event, "lobby/listUpdated", data, state}
@@ -1012,6 +1022,17 @@ defmodule Teiserver.Player.TachyonHandler do
     {:response, state}
   end
 
+  def handle_command("lobby/kickban", "request", _msg_id, %{"data" => data}, state) do
+    target_id = data["userId"]
+    ban_until = parse_ban_until(data["banUntil"])
+
+      case Session.lobby_kickban(state.user.id, target_id, ban_until) do
+        :ok -> {:response, state}
+        {:error, :unauthorized} -> {:error_response, :unauthorized, state}
+        {:error, reason} -> {:error_response, :invalid_request, to_string(reason), state}
+      end
+  end
+
   def handle_command(_command_id, _message_type, _message_id, _message, state) do
     {:error_response, :command_unimplemented, state}
   end
@@ -1433,6 +1454,14 @@ defmodule Teiserver.Player.TachyonHandler do
       end
     else
       %{}
+    end
+  end
+
+  defp parse_ban_until(nil), do: nil
+  defp parse_ban_until(ts) do
+    case DateTime.from_unix(ts, :microsecond) do
+      {:ok, dt} -> if DateTime.compare(dt, DateTime.utc_now()) == :gt, do: dt, else: nil
+      _ -> nil
     end
   end
 end
--- a/lib/teiserver/tachyon_lobby/lobby.ex
+++ b/lib/teiserver/tachyon_lobby/lobby.ex
@@ -468,7 +468,7 @@ defmodule Teiserver.TachyonLobby.Lobby do
 
   def handle_event(
         {:call, from},
-        {:join, %LT.PlayerJoinData{id: user_id} = _join_data, _pid},
+        {:join, %LT.PlayerJoinData{id: user_id} = join_data, pid},
         _state,
         %LT.Data{} = data
       )
@@ -477,7 +477,7 @@ defmodule Teiserver.TachyonLobby.Lobby do
 
     if ban_expired?(ban_until) do
       data = update_in(data.banned_users, &Map.delete(&1, user_id))
-      {:keep_state, data, [{:postpone, true}]}
+      {:keep_state, data, [{:next_event, {:call, from}, {:join, join_data, pid}}]}
     else
       {:keep_state, data, [{:reply, from, {:error, :banned}}]}
     end
--- a/lib/teiserver/tachyon_lobby/lobby.ex
+++ b/lib/teiserver/tachyon_lobby/lobby.ex
@@ -982,9 +982,6 @@ defmodule Teiserver.TachyonLobby.Lobby do
       end
 
     cond do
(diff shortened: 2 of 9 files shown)


lib/teiserver/tachyon_lobby/lobby.ex
```diff
@@ -928,6 +951,64 @@ defmodule Teiserver.TachyonLobby.Lobby do
...
+      ) do
+    effective_ban_until =
+      case ban_until do
+        nil -> nil
+        dt -> if DateTime.compare(dt, DateTime.utc_now()) == :gt, do: dt, else: nil
+      end
+
+    cond do
+      not data.boss_enabled? ->
+        {:keep_state, data, [{:reply, from, {:error, :no_boss_allowed}}]}
```
L-e-x-o-n: What's the purpose of this? Why require bosses but also support voting if there aren't any?
IMO kickbans should still be possible democratically like they are now when boss mode is disabled.

lib/teiserver/tachyon_lobby/lobby.ex
```diff
@@ -1513,8 +1594,13 @@ defmodule Teiserver.TachyonLobby.Lobby do
 
           {:passed, {:appoint_boss, boss_id}} ->
             process_event({:update_boss, :add, boss_id}, new_aggregate)
-            # just let the thing crash if a new vote action shows up. It'll be easy
-            # to spot and fix/add support. :start isn't yet supported
+
+          # just let the thing crash if a new vote action shows up. It'll be easy
+          # to spot and fix/add support. :start isn't yet supported
```
L-e-x-o-n: Minor thing, but the comment should go below your changes.

L-e-x-o-n:
It would be nice if it didn't allow kickbanning moderators and allowed moderators to kickban without vote or being a boss but it isn't a requirement.

BElluu:
> It would be nice if it didn't allow kickbanning moderators and allowed moderators to kickban without vote or being a boss but it isn't a requirement.

This is good idea but I prefer do it in separate PR. The problem is, lobby doesn't know anything about user roles. I will create new issue, after merge.

L-e-x-o-n:
> > It would be nice if it didn't allow kickbanning moderators and allowed moderators to kickban without vote or being a boss but it isn't a requirement.
> 
> This is good idea but I prefer do it in separate PR. The problem is, lobby doesn't know anything about user roles. I will create new issue, after merge.

The client doesn't necessarily need to know roles, the server does. Sounds good.

L-e-x-o-n (approved):

lib/teiserver/player/tachyon_handler.ex
```diff
@@ -1012,6 +1022,17 @@ defmodule Teiserver.Player.TachyonHandler do
     {:response, state}
   end
 
+  def handle_command("lobby/kickban", "request", _msg_id, %{"data" => data}, state) do
+    target_id = data["userId"]
```
geekingfrog, line 1026: this doesn't work. User ids are strings at the protocol level, but in teiserver these are the primary keys and are integer. You are missing one "happy path" test in teiserver_web to make sure the tachyon handler works.

lib/teiserver/tachyon_lobby/lobby.ex
```diff
@@ -928,6 +951,61 @@ defmodule Teiserver.TachyonLobby.Lobby do
...
+      ) do
+    effective_ban_until =
+      case ban_until do
+        nil -> nil
+        dt -> if DateTime.compare(dt, DateTime.utc_now()) == :gt, do: dt, else: nil
+      end
+
+    cond do
+      not Enum.empty?(data.bosses) and not MapSet.member?(data.bosses, user_id) ->
+        {:keep_state, data, [{:reply, from, {:error, :unauthorized}}]}
```
geekingfrog, line 986: This is not the meaning of unauthorized. According to https://github.com/beyond-all-reason/tachyon/blob/master/docs/commands.md that code should be reserved for role/permission check. I would argue that in this case, it should be a generic bad request error, not unauthorized.

lib/teiserver/tachyon_lobby/lobby.ex
```diff
@@ -1948,6 +2031,39 @@ defmodule Teiserver.TachyonLobby.Lobby do
...
+      case {data.players[target_id], data.spectators[target_id]} do
+        {%{pid: pid}, _spec} -> pid
+        {_player, %{pid: pid}} -> pid
+        _not_found -> nil
+      end
+
+    data =
+      case ban_until do
+        nil -> data
+        dt -> put_in(data.banned_users[target_id], dt)
```
geekingfrog, line 2049: the datetime check against the current time should be done at that point, otherwise it may be valid when the vote starts, and in the past when the vote ends.

lib/teiserver/tachyon_lobby/lobby.ex
```diff
@@ -1513,6 +1591,11 @@ defmodule Teiserver.TachyonLobby.Lobby do
 
           {:passed, {:appoint_boss, boss_id}} ->
             process_event({:update_boss, :add, boss_id}, new_aggregate)
+
+          {:passed, {:kickban, target_id, ban_until}} ->
+            new_data = do_kickban(target_id, ban_until, new_aggregate.data)
```
geekingfrog, line 1596: you need to check, either there or in `do_kickban` that the target is still in the lobby at that point.

lib/teiserver/tachyon_lobby/lobby.ex
```diff
@@ -1948,6 +2031,39 @@ defmodule Teiserver.TachyonLobby.Lobby do
...
+      if is_map_key(data.players, target_id) do
+        remove_player_from_lobby(target_id, data)
+      else
+        remove_spectator_from_lobby(target_id, data)
+      end
+
+    if target_pid do
+      reason = if ban_until != nil, do: "banned", else: "kicked"
+      send(target_pid, {:lobby, data.id, {:left, reason, ban_until}})
+    end
```
geekingfrog, line 2062: There is a slight problem with this logic: the target client will receive 2 updates. One with its own id being removed from the lobby, followed by another event `lobby/left`. I think it's rather confusing, there should only be one event sent to the target player.

lib/teiserver/tachyon_lobby/lobby.ex
geekingfrog, line 1: If the target leaves the lobby and the vote is for a kick only (not ban), then there should also be logic to cancel the vote.

The `banned_users` map is also only cleaned up when the user re-attempt to join the lobby. There should be some sort of cleanup process/message for this. Probably with a `:timer.send_after` message.

test/teiserver/tachyon_lobby/lobby_test.exs
```diff
@@ -1700,6 +1700,155 @@ defmodule Teiserver.TachyonLobby.LobbyTest do
...
+        |> Map.put(:boss_enabled?, true)
+        |> Lobby.create()
+
+      {:ok, sink_pid} = Task.start_link(:timer, :sleep, [:infinity])
+      {:ok, _lobby_pid, _details} = Lobby.join(id, mk_player("user2"), sink_pid)
+      assert_receive {:lobby, ^id, {:updated, _}}
+      ban_until = DateTime.add(DateTime.utc_now(), 3600, :second)
+      :ok = Lobby.kickban(id, @default_user_id, "user2", ban_until)
+
+      {:error, :banned} = Lobby.join(id, mk_player("user2"), sink_pid)
```
geekingfrog, line 1716: When banned, there should be a way to indicate to the caller the time until when they are banned.

test/teiserver/tachyon_lobby/lobby_test.exs
```diff
@@ -1700,6 +1700,155 @@ defmodule Teiserver.TachyonLobby.LobbyTest do
...
+      :ok = Lobby.kickban(id, @default_user_id, "user2")
+      assert_receive {:lobby, ^id, {:left, "kicked", nil}}
+
+      {:ok, details} = LobbyProcess.get_details(id)
+      refute is_map_key(details.players, "user2")
+
+      {:ok, _lobby_pid, _details} = Lobby.join(id, mk_player("user2"), self())
+      {:ok, _details} = Lobby.join_ally_team(id, "user2", 0)
+      {:ok, details} = LobbyProcess.get_details(id)
+      assert is_map_key(details.players, "user2")
```
geekingfrog, line 1757: I don't think this is useful. You are just testing that kickban works on both player and spectators, no need to re-test the logic to join an ally team.

test/teiserver/tachyon_lobby/lobby_test.exs
```diff
@@ -1700,6 +1700,155 @@ defmodule Teiserver.TachyonLobby.LobbyTest do
...
+      {:ok, sink_pid} = Task.start_link(:timer, :sleep, [:infinity])
+      {:ok, _lobby_pid, _details} = Lobby.join(id, mk_player("user2"), sink_pid)
+      assert_receive {:lobby, ^id, {:updated, _}}
+
+      ban_until = DateTime.add(DateTime.utc_now(), 1, :second)
+      :ok = Lobby.kickban(id, @default_user_id, "user2", ban_until)
+
+      {:error, :banned} = Lobby.join(id, mk_player("user2"), sink_pid)
+
+      Process.sleep(1100)
```
geekingfrog, line 1846: :vomiting_face: especially for such a long time. Put something much shorter, like 30ms perhaps.

test/teiserver/tachyon_lobby/lobby_test.exs
```diff
@@ -1700,6 +1700,155 @@ defmodule Teiserver.TachyonLobby.LobbyTest do
...
+      drain_msg_queue()
+
+      :ok = Lobby.kickban(id, @default_user_id, "user2")
+
+      assert_receive {:lobby, ^id, {:updated, %{current_vote: vote}}}
+      assert vote.action == {:kickban, "user2", nil}
+      assert vote.voters[@default_user_id] == :yes
+      assert vote.voters["user2"] == :pending
+      assert vote.voters["user3"] == :pending
+    end
```
geekingfrog, line 1829: I would extend this test to make the vote pass, so as to test the logic around voting + kickban.

geekingfrog (commented):
You need to add some tests under teiserver_web, as it stands, the tachyon handler, as well as the interaction between handler <-> session <-> lobby are untested

lib/teiserver/tachyon_lobby/lobby.ex
Hectate: Currently, the client reasonably expects to get a ``lobby/left`` event rather than a ``lobby/updated`` event if they are the removed user, FWIW. The other client's should get the ``lobby/updated`` of course.

BElluu:
> You need to add some tests under teiserver_web, as it stands, the tachyon handler, as well as the interaction between handler <-> session <-> lobby are untested

Sure! I will look at this tomorrow

PR #1256, opened 2026-06-08
Disable $website command for MFA users

Teifion:
Users with MFA or any roles requiring MFA will no longer be able to get a one time link using the `$website` command as it would allow them to circumvent the MFA.


ChocoboCoding:
The plan was to disable the command if someone has MFA disabled, not to completely remove it

Teifion:
Yes, Lexon messaged me about it and I'm in the process of changing that over :)

PR #1259, opened 2026-06-09
Fix test SAYPRIVATE with special characters

BElluu:
Fix `_recv_raw` and `_recv_until` to correctly handle non-ASCII data.

### DIFF
--- a/test/support/teiserver_test_lib.ex
+++ b/test/support/teiserver_test_lib.ex
@@ -160,7 +160,7 @@ defmodule Teiserver.TeiserverTestLib do
 
   def _recv_raw({:sslsocket, _tcp, _tls} = socket) do
     case :ssl.recv(socket, 0, 500) do
-      {:ok, reply} -> reply |> to_string()
+      {:ok, reply} -> IO.iodata_to_binary(reply)
       {:error, :timeout} -> :timeout
       {:error, :closed} -> :closed
     end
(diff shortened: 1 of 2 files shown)


L-e-x-o-n:
Looking at the docs  [`IO.data_to_binary`](https://elixir.hexdocs.pm/1.12/IO.html#iodata_to_binary/1) it looks like [`IO.chardata_to_string`](https://elixir.hexdocs.pm/1.12/IO.html#chardata_to_string/1) might be more appropriate here.

BElluu:
All three functions tested against the same input:

```elixir
mix run -e '
bytes = :binary.bin_to_list("тест!")
IO.inspect(bytes, label: "raw bytes from socket")

result1 = to_string(bytes)
IO.inspect(result1 == "тест!", label: "to_string correct?")

result2 = IO.chardata_to_string(bytes)
IO.inspect(result2 == "тест!", label: "IO.chardata_to_string correct?")

result3 = IO.iodata_to_binary(bytes)
IO.inspect(result3 == "тест!", label: "IO.iodata_to_binary correct?")
'
```

Output:

```
raw bytes from socket: [209, 130, 208, 181, 209, 129, 209, 130, 33]
to_string correct?: false
IO.chardata_to_string correct?: false
IO.iodata_to_binary correct?: true
```

The test client socket connects without `mode: :binary`, so `:gen_tcp.recv` returns a charlist of raw bytes. Both `to_string/1` and `IO.chardata_to_string/1` treat list elements as Unicode code points (not bytes), which corrupts multi-byte UTF-8 sequences. `IO.iodata_to_binary/1` treats them as raw bytes, which is what we need here.

L-e-x-o-n:
Thanks!

PR #1260, opened 2026-06-09
fetch stats and ratings only for newly joined players

BElluu:
Instead of reload whole data for all of players, we check diff who joined, who left and do query just for new players.


Teifion:
This is adding a moderate amount of complexity but I can't see the big advantage. Could you elaborate on why you see this as an issue that needs addressing and the value in accepting the added complexity please?

BElluu:
This code had a little bug. When a player changes battle status (switch teams), the diff doesn't refetch their client data since they already existsted in lobby. After fix that, I used AI to do some benchmarks. The full reload of `get_user_and_clients` takes ~0.285ms for 32 players. The lobby throttle fires at most every 500ms, so this uses <0.1% of the available time. The diff approach saves some per update but adds complexity.

image

So I'll leave it up to you to decide. If we don't want to do that, let's just close this PR and while we're working on another one, we'll simply delete the TODO from here :)

Teifion:
I think in this case it's not important enough to add the complexity. Long term I think this page is going to be removed completely. I'll close the PR in this case.

PR #1263, opened 2026-06-09
Credo - WeakAssertions

### LINKED ISSUE #1103: Enable WeakAssertion credo check
As part of https://github.com/beyond-all-reason/teiserver/pull/1099 a credo check for tests with weak assertions was added but commented out. A weak assertion in this case is:

```elixir
assert some_value != nil
```

Ideally in every case we would match against a struct but given our lack of use of structs this will not always be possible.

BElluu:
Closed #1103 

After removed Timex time for WeakAssertions:)


geekingfrog:
Before we even consider that, I'd like some discussion in the linked issue. basically, I'm not keen on this check being introduced

Teifion:
I have created a [discord thread](https://discord.com/channels/549281623154229250/1514012854100033648) for said discussion.

test/teiserver/o_auth/token_test.exs
```diff
@@ -131,7 +131,7 @@ defmodule Teiserver.OAuth.TokenTest do
     {:error, changeset} =
       OAuth.refresh_token(refresh_token, scopes: ["tachyon.lobby", "lolscope"])
 
-    assert Keyword.get(changeset.errors, :scopes) != nil
+    assert Keyword.has_key?(changeset.errors, :scopes)
```
Teifion: This change is semantically not a change, please make it assert something about the value.

test/teiserver_web/controllers/admin/bot_controller_test.exs
```diff
@@ -162,8 +162,7 @@ defmodule TeiserverWeb.Admin.BotControllerTest do
       assert conn.status == 400
 
       # cred is still here
-      assert {:ok, cred} = OAuth.get_valid_credentials("client_id", "verysecret")
-      refute is_nil(cred)
+      assert {:ok, _cred} = OAuth.get_valid_credentials("client_id", "verysecret")
```
Teifion: We're not asserting the value of the credential (which should be a `Teiserver.OAuth.Credential` struct).

test/teiserver_web/controllers/admin/o_auth_application_controller_test.exs
```diff
@@ -79,7 +79,6 @@ defmodule TeiserverWeb.Admin.OAuthApplicationControllerTest do
       resp = assert html_response(conn, 200)
       assert resp =~ "generic name"
       db_app = OAuth.get_application_by_uid(data["application"]["uid"])
-      assert is_binary(db_app.secret)
```
Teifion, line 82: Why are we removing this assertion?

Teifion (commented):
A few minor issues and then should be good to merge.

BElluu:
Approved on discord, so please look at this code and I will fix or explain everything at once :)

test/teiserver_web/controllers/admin/o_auth_application_controller_test.exs
BElluu: Initially I removed `is_binary` thinking that `Argon2.verify_pass` below would implicitly catch a non-binary secret. After checking implementation of `verift_pass` it turns out it calls `:binary.bin_to_list` which accepts any binary so it wouldn't catch e.g empty string. I added assertion which verifies the secret is actually a valid Argon2 hash, not just any binary.

test/teiserver/o_auth/token_test.exs
BElluu: Fixed. I found other places with same thing so fixed too :)

test/teiserver_web/controllers/admin/bot_controller_test.exs
BElluu: This file is deleted, but I found other place with that code, so I fixed it :)

Teifion (approved):
Nice work, thank you :)

PR #1266, opened 2026-06-10
First stage of a UserPicker component

Teifion:
As part of an intent to be able to create a smurf link to any user I realised we needed a new UserPicker component, this PR adds that and the non-functional design for the page. It looks bad because the CoreComponent code used is designed for Tailwind but that is for a later discussion/PR.

Page currently shows the in-progress warning
image

Allows searching of users
image

If an exact match happens that user will always appear first, no more than 10 users are ever shown
image

When the user is selected (by clicking their row in the table) the input is updated with something holding the ID of the user (for the backend to read/use) and the name of the user selected (for the person using the page)
image

### DIFF
--- a/lib/teiserver/account/queries/user_queries.ex
+++ b/lib/teiserver/account/queries/user_queries.ex
@@ -533,4 +533,39 @@ defmodule Teiserver.Account.UserQueries do
       {:ok, query}
     end
   end
+
+  # New query style
+  @spec users() :: t()
+  def users do
+    from(users in User, as: :users)
+  end
+
+  @spec where_name(t(), String.t()) :: t()
+  def where_name(query, search_term) do
+    from users in query,
+      where: users.name == ^search_term
+  end
+
+  @spec where_name_like(t(), String.t()) :: t()
+  def where_name_like(query, search_term) do
+    uname = "%" <> search_term <> "%"
+
+    from users in query,
+      where: ilike(users.name, ^uname)
+  end
+
+  @spec where_id_not_in(t(), [User.id()]) :: t()
+  def where_id_not_in(query, ids) do
+    from users in query,
+      where: users.id not in ^ids
+  end
+
+  @spec order_by_name(t(), :asc | :desc) :: t()
+  def order_by_name(query, direction \\ :asc) do
+    if direction == :asc do
+      from(users in query, order_by: [asc: users.name])
+    else
+      from(users in query, order_by: [desc: users.name])
+    end
+  end
 end
--- a/lib/teiserver/account/queries/user_queries.ex
+++ b/lib/teiserver/account/queries/user_queries.ex
@@ -540,6 +540,12 @@ defmodule Teiserver.Account.UserQueries do
     from(users in User, as: :users)
   end
 
+  @spec where_id(t(), pos_integer() | String.t()) :: t()
+  def where_id(query, id) do
+    from users in query,
+      where: users.id == ^id
+  end
+
   @spec where_name(t(), String.t()) :: t()
   def where_name(query, search_term) do
     from users in query,
--- a/lib/teiserver_web/live/moderation/action_live/smurf_link.ex
+++ b/lib/teiserver_web/live/moderation/action_live/smurf_link.ex
@@ -1,0 +1,30 @@
+defmodule TeiserverWeb.Moderation.ActionLive.SmurfLink do
+  @moduledoc false
+  alias Teiserver.Account
+  alias Teiserver.Account.UserLib
+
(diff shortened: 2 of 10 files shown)


L-e-x-o-n (commented):
Nice. code seems good but I have a few feature suggsetions.

Can you make it accept user ID as well, it would be very convenient for cases where you have the user ID and don't remember the name (or they renamed).

It would be nice if it showed the last~3 names of the user as well, our current rename limits are quite generous and some users like to rename frequently and sometimes to names of other users which could lead to the wrong user being selected. There is no need for previous names to be searchable, just shown.

Will this eventually replace the old user picker?

Teifion:
> Can you make it accept user ID as well, it would be very convenient for cases where you have the user ID and don't remember the name (or they renamed).

Agreed, will add searching for if you pass in an ID.

> It would be nice if it showed the last~3 names of the user as well

We could add that in future though it strays from the purpose of the component. The idea of this component is if you know who you are trying to pick, you just need an interface to pick them. Something more complex with things like past names or other data points deserves a different interface.

> Will this eventually replace the old user picker?

Yes, that user picker won't work on liveviews for a start.

lib/teiserver_web/live_components/user_picker.ex
```diff
@@ -1,11 +1,201 @@
...
+      UserQueries.users()
+      |> UserQueries.where_name_like(search_term)
+      |> UserQueries.where_id_not_in(exact_ids)
+      |> UserQueries.order_by_name()
+      |> limit(^query_limit)
+      |> Repo.all()
+
+    socket
+    |> assign(users: exact_matches ++ users)
+  end
```
BElluu: Shouldn't you check if `search_term != ""` and only then run the query?

lib/teiserver_web/live/moderation/action_live/smurf_link.ex
```diff
@@ -0,0 +1,30 @@
...
+  use TeiserverWeb, :live_view
+
+  @impl LiveView
+  def mount(_params, _session, socket) do
+    {:ok, socket}
+  end
+
+  @impl LiveView
+  def handle_params(%{"user_id" => user_id}, _url, socket) do
+    user = Account.get_user!(user_id)
```
BElluu: if user_id do not exists, I think we will get 500. Maybe you should check it and redirect, same like below when _no_access?

lib/teiserver/account/queries/user_queries.ex
```diff
@@ -533,4 +533,39 @@ defmodule Teiserver.Account.UserQueries do
...
+  end
+
+  @spec order_by_name(t(), :asc | :desc) :: t()
+  def order_by_name(query, direction \\ :asc) do
+    if direction == :asc do
+      from(users in query, order_by: [asc: users.name])
+    else
+      from(users in query, order_by: [desc: users.name])
+    end
+  end
```
BElluu, line 576: What do you think about pattern matching in this case instead of if else?
```
defp order_by_name(query, :asc), do: from(users in query, order_by: [asc: users.name])
defp order_by_name(query, :desc), do: from(users in query, order_by: [desc: users.name])
```

Teifion: I am in two minds on that, I know with the moderation updates @geekingfrog had an opinion so I'm curious as to his thoughts. I am happy either way.

lib/teiserver_web/live/moderation/action_live/smurf_link.ex
Teifion: I will test it and if it does then I will make changes, thank you :)

lib/teiserver_web/live_components/user_picker.ex
Teifion: Good shout, addressed in https://github.com/beyond-all-reason/teiserver/pull/1266/commits/16d307f4ff89084d0b32b867d2433a23fc6bd69d

lib/teiserver_web/live/moderation/action_live/smurf_link.ex
Teifion: You were correct, addressed in https://github.com/beyond-all-reason/teiserver/pull/1266/commits/16d307f4ff89084d0b32b867d2433a23fc6bd69d

L-e-x-o-n (approved):

lib/teiserver/account/queries/user_queries.ex
L-e-x-o-n: I prefer the pattern matching as well but don't mind your approach either.

PR #1268, opened 2026-06-11
Simpler tachyon envelope validation

geekingfrog:
Instead of using json schema to parse the outer layer of a tachyon message, manually verify it. The schema is super simple and well known, and it is much faster without json schema.

before:
2026-06-11-996x393-scrot

after:
2026-06-11-996x360-scrot


L-e-x-o-n (approved):

Teifion (approved):
Looks good, might be worth tightening up the tests one day to verify the error message returned but I don't think we need to do that right now.

PR #1269, opened 2026-06-11
Rework lobby list

geekingfrog:
There is currently a problem with the way (tachyon) lobbies listing work. Every overview is gathered in a central process, which then broadcast the updates. This scales poorly, increasing the number of core/scheduler doesn't help at all.
I tried to shard the process, but ultimately it was rather complicated code, and it turns out we can do this whole list lobby logic differently.

This PR moves the overviews to a registry. This is conceptually the same as a central process, but has the benefit of supporting partitions (not enabled in this PR), and is in the standard lib + battle tested.
This requires an additional process to monitor lobbies and broadcast an event when they close/die. That new process: `ListMonitor` could also be sharded, but it's also only looking at `DOWN` events, and so I am not expecting a lot of load.

A few additional considerations:

* The `ListMonitor` process is introduced to avoid a `n * m` problem where every session needs to monitor every lobby to know when to remove them from the list. This would scale poorly. A single process handling the monitors work better.
* The `ListMonitor` can potentially hold tens of thousands of monitor. We can easily partition it, but even then, it should be fine. The login queue already monitors all session blocked in the queue, and I had 50k+ monitors without much problem.
* If the `ListMonitor` process dies, and a lobby exits while it is restarting, then subscribe clients will never receive the update about the lobby dying. I believe this is an acceptable tradeoff. This should happen rather rarely, and we can also send an update to remove the lobby from the list when a lobby operation fails with an invalid lobby code (though this may send bogus updates if the id is just plain wrong).
* There is no batching with this method. I'll address that in a future PR. After some quick load testing, we absolutely need batching. Each session can batch the updates. This will cost a bit more memory, but drastically cut down the number of message to send to clients.


lib/teiserver/tachyon_lobby/lobby.ex
```diff
@@ -1237,6 +1236,22 @@ defmodule Teiserver.TachyonLobby.Lobby do
...
+
+    message = %{
+      event: :add_lobby,
+      counter: state.counter,
+      lobby_id: state.id,
+      overview: overview
+    }
+
+    PubSubHelper.broadcast(list_topic(), message)
+    :ok
```
Teifion: Returning `:ok` is not needed as this is never matched elsewhere.

Teifion (approved):

PR #1272, opened 2026-06-11
Moderation actions start expiring after acknowledgement

### LINKED ISSUE #624: [Feature]: Moderation actions start expiring after acknowledgement
### Describe the Feature

Currently moderation actions have a set expiration date and so it is possible for users to miss moderation actions entirely if they don't login while the action is in effect.

### Expected Behavior

Moderation actions should start expiring only after the user logs in once and acknowledges the action.

### Screenshots

_No response_

### Additional context

Not an easy task, will require significant rework to how moderation actions function. User actioning form expiration date field should be replaced with action duration. Expiration date should then be set automatically (based on set duration) when the user logs in and acknowledges the action.

BElluu:
Closed #624 

Preview showing the operation of the mechanism with a user login simulation: https://streamable.com/qitpvu


lib/teiserver/moderation/schemas/action.ex
L-e-x-o-n, line 1: It seems that duration can currently be `nil` but that doesn't do anything?

Not a requirement, but I would make use of the new duration field and make is so that when duration is `nil` it's considered a permanent ban, this seems cleaner and nicer than having 1000 year long bans.

lib/teiserver_web/controllers/moderation/action_controller.ex
```diff
@@ -246,15 +246,12 @@ defmodule TeiserverWeb.Moderation.ActionController do
   def create(conn, %{"action" => action_params}) do
     user = Account.get_user(action_params["target_id"])
 
-    expires_string = action_params["expires"] || ""
+    now = DateTime.utc_now()
```
L-e-x-o-n: This is unnecessary and a default handled by `human_input_to_datetime`.

lib/teiserver_web/controllers/moderation/action_controller.ex
```diff
@@ -246,15 +246,12 @@ defmodule TeiserverWeb.Moderation.ActionController do
...
-      case DateHelper.human_input_to_datetime(expires_string) do
-        {:ok, expires_datetime} ->
-          expires_datetime
-
-        nil ->
-          action_params["expires"]
+    duration_seconds =
+      case DateHelper.human_input_to_datetime(action_params["duration"] || "", now) do
+        {:ok, future_datetime} -> DateTime.diff(future_datetime, now)
+        nil -> nil
```
L-e-x-o-n: What happens when duration is `nil`?

lib/teiserver_web/controllers/moderation/action_controller.ex
```diff
@@ -350,14 +348,31 @@ defmodule TeiserverWeb.Moderation.ActionController do
...
+      end
+
     restrictions =
       action_params["restrictions"]
       |> Map.values()
       |> Enum.reject(fn v -> v == "false" end)
 
+    expires =
+      if duration_seconds != nil and action.expires != nil do
+        NaiveDateTime.add(action.inserted_at, duration_seconds, :second)
```
L-e-x-o-n: Why add to `inserted_at`? I would add it to `utc_now` instead.

Teifion:
I've checked with the moderation team and this issue may be out of date, please hold fire on it for now.

I will try to make time to remove any no-longer-relevant issues as I can see you are making steady progress on them :)

lib/teiserver_web/controllers/moderation/action_controller.ex
```diff
@@ -350,14 +348,31 @@ defmodule TeiserverWeb.Moderation.ActionController do
   def update(conn, %{"id" => id, "action" => action_params}) do
     action = Moderation.get_action!(id)
 
+    now = DateTime.utc_now()
+
+    duration_seconds =
+      case DateHelper.human_input_to_datetime(action_params["duration"] || "", now) do
```
L-e-x-o-n: Same

lib/teiserver/coordinator/coordinator_server.ex
L-e-x-o-n, line 1: Actions with `Warning reminder` restriction should start expiring only after the user directly acknowledges the restriction. Coordinator will message the user with active restrictions that need acknowledgement and ask them to acknowledge it in chat by sending the appropriate message, only after the user sends the acknowledgment message should those actions start to expire.

kroIya: > Actions with `Warning reminder` .....

That is correct. Whether we want to implement a similar system for other restrictions is an open question that will need to be resolved before merge.
In this PR specifically, what does acknowledging an action other than a `Warning reminder` look like, given that we don't currently have a mechanism for acknowledging other restriction types?

L-e-x-o-n: In this PR just logging in would be "acknowledging" the action/starting the expiration.

Teifion: > In this PR specifically, what does acknowledging an action other than a Warning reminder look like, given that we don't currently have a mechanism for acknowledging other restriction types?

Given this is focused around source code you can see what it looks like using the [Files changed](https://github.com/beyond-all-reason/teiserver/pull/1272/changes) link. Optionally reviewers could also request screenshots of interface changes if there are any.

BElluu:
I fix some things from review + nil (empty) duration is Permament.
image

Teifion:
I've chased the moderation team regarding a decision on this and they have promised to get back to me today or tomorrow. Apologies again for the delay on this.

BElluu:
No problem. We do not need to rush :)

PtaQQ:
Thanks for your patience, and sorry for the delay while we aligned internally.

The conclusion is to narrow this PR’s scope for now:

* Apply the acknowledgement-based expiry behaviour to **Warning reminders only**. A warning reminder should remain active until it has been acknowledged, using the existing acknowledgement mechanism.
* Do **not** change expiry behaviour for other moderation restrictions/actions in this PR. That needs to be decided alongside broader changes to notification flow, the public moderation log, and the future client UI.
* It would be good for the implementation to remain reasonably extensible so we can apply similar behaviour to other actions later, but that is not a requirement for this PR.

As a small follow-up, it would be nice to support warning reminders that expire immediately once acknowledged, rather than needing a token duration such as one second. That is non-blocking, though.

Thank you again for the work on this and let me know if there are any additional questions.

BElluu:
Sounds good. Let me handle it at the end of this week. If I will have any question, I will catch you on discord :)

Teifion:
Thanks for the response PtaQQ.

> As a small follow-up, it would be nice to support warning reminders that expire immediately once acknowledged, rather than needing a token duration such as one second. That is non-blocking, though.

I would suggest that goes into a separate PR.

PR #1274, opened 2026-06-12
party/create response return party state insead of party_id

### LINKED ISSUE #1267: Return full party state with `party/create` response
The `party/create` response [was changed](https://github.com/beyond-all-reason/tachyon/pull/142), instead of just returning the party ID, the full party state should be returned instead.

BElluu:
Closed #1267

### DIFF
--- a/lib/teiserver/player/session.ex
+++ b/lib/teiserver/player/session.ex
@@ -416,7 +416,7 @@ defmodule Teiserver.Player.Session do
   end
 
   @spec create_party(User.id()) ::
-          {:ok, Party.id()} | {:error, :already_in_party} | {:error, reason :: term()}
+          {:ok, Party.state()} | {:error, :already_in_party} | {:error, reason :: term()}
   def create_party(user_id) do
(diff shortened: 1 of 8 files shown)


BElluu:
One test not pass, but that's not my :) There is too https://github.com/beyond-all-reason/teiserver/actions/runs/27402832096/job/80984685798

Teifion (approved):
This looks good to me and I provisionally approve it but given @geekingfrog is more familiar with the changes I would welcome his input too.

lib/teiserver/player/session.ex
```diff
@@ -928,6 +928,8 @@ defmodule Teiserver.Player.Session do
   def handle_call({:party, :create}, _from, state) do
     case Party.create_party(state.user.id) do
       {:ok, party_id, pid} ->
+        party_state = Party.get_state(party_id)
```
geekingfrog: This is racy and wasteful, `create_party` should return the state directly instead of requiring another call. Instead of returning `party_id`, returns the state

geekingfrog (approved):

PR #1276, opened 2026-06-12
Handle rate limit login queue for Tachyon

### LINKED ISSUE #1208: Rate limit login queue for tachyon
Similar to how it is done in the [existing rate limiter](https://github.com/beyond-all-reason/teiserver/blob/d835f07e623d1e6742e25b20afbb4af9a75ac050/lib/teiserver/account/servers/login_throttle_server.ex), the tachyon rate limiter should have a configurable limit of user per minute to let in.
The default limit should be extremely high (like 10s of thousands per minute) so it doesn't interfere with any local testing or tests.

BElluu:
Closed #1208 

AI used for describe how BurstyRateLimiter works and find way with `should_fill` because I didn't understand why tests fails :D

### DIFF
--- a/lib/teiserver/player/login_queue.ex
+++ b/lib/teiserver/player/login_queue.ex
@@ -12,6 +12,7 @@ defmodule Teiserver.Player.LoginQueue do
 
   alias Teiserver.Account.User
   alias Teiserver.Config
+  alias Teiserver.Helpers.BurstyRateLimiter
   alias Teiserver.Player.SessionRegistry
   alias Teiserver.Player.TachyonHandler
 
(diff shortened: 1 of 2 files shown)


lib/teiserver/player/login_queue.ex
```diff
@@ -53,6 +55,16 @@ defmodule Teiserver.Player.LoginQueue do
...
 
+  @doc """
+  Set the login rate (users per minute).
+  If `should_fill?` is true, the new rate limiter starts with full permits and
+  admits users immediately up to the burst capacity.
+  """
+  @spec set_rate(non_neg_integer(), boolean()) :: :ok
+  def set_rate(rate, should_fill? \\ false) do
+    GenServer.call(__MODULE__, {:set_rate, rate, should_fill?})
+  end
```
geekingfrog, line 66: Instead of sending a message, I would do the same as the existing login throttle, where you have this API, but also one where you can directly pass a rate limiter object, and you pass this struct in the message to override the limiter in the process

geekingfrog (approved):

PR #1280, opened 2026-06-13
spring: allow longer `!mode` commands in battle chat

keithharvey:
## Problem

`SAYBATTLE`/`SAYBATTLEEX` slice normal battle-chat messages to 256 characters. The `!mode` autohost command (used by the sharing-mode preset system) carries a set of modoption `key=value` pairs and can exceed 256 chars. When it does, it's silently truncated mid-token into an invalid command, so the autohost rejects it (e.g. `Invalid parameter format "take_del"`) and no vote is started — with no feedback to the user.

## Change

Give `!mode`/`$mode` a 1024-character allowance in both `SAYBATTLE` and `SAYBATTLEEX`, the same tier already used for `welcome-message` commands (tweakdefs/tweakunits already get 16384). 1024 comfortably fits a mode name plus its full modoption set.

Part of: https://github.com/beyond-all-reason/BYAR-Chobby/issues/1040

### DIFF
--- a/lib/teiserver/protocols/spring/spring_in.ex
+++ b/lib/teiserver/protocols/spring/spring_in.ex
@@ -47,6 +47,10 @@ defmodule Teiserver.Protocols.SpringIn do
 
   @cluster_manager_regex ~r/^Host\[[A-Z]+\d+\]$/
 
+  # Battle chat commands that carry more data than a normal message (e.g. modoption
+  # key=value pairs) and so get a 1024-char allowance instead of the default 256.
+  @long_battle_command_prefixes ["$welcome-message", "!welcome-message", "!mode "]
+
   # Commands that don't require the user to be logged in
   @unauthenticated_commands ~w(
     PING STLS LOGIN REGISTER CONFIRMAGREEMENT RESETPASSWORDREQUEST EXIT CHANGEPASSWORD LISTCOMPFLAGS
@@ -1267,10 +1271,7 @@ defmodule Teiserver.Protocols.SpringIn do
               String.starts_with?(lowercase_msg, "!bset tweakunits") ->
             msg |> String.trim() |> String.slice(0..16_384)
 
-          String.starts_with?(lowercase_msg, ["$welcome-message", "!welcome-message"]) ->
-            msg |> String.trim() |> String.slice(0..1024)
-
-          String.starts_with?(lowercase_msg, "!mode ") ->
+          String.starts_with?(lowercase_msg, @long_battle_command_prefixes) ->
             msg |> String.trim() |> String.slice(0..1024)
 
           true ->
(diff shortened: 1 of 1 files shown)


keithharvey:
Think I found a Heisentest, can someone with privs kick the CI for me (passes locally, including with the same seed and in parallel)?

lib/teiserver/protocols/spring/spring_in.ex
```diff
@@ -1296,6 +1299,9 @@ defmodule Teiserver.Protocols.SpringIn do
           String.starts_with?(lowercase_msg, ["$welcome-message", "!welcome-message"]) ->
             msg |> String.trim() |> String.slice(0..1024)
 
+          String.starts_with?(lowercase_msg, ["!mode ", "$mode "]) ->
+            msg |> String.trim() |> String.slice(0..1024)
+
```
L-e-x-o-n: Why `$mode`? $ are teiserver commands and I am not aware of one for mode existing.

keithharvey: Yes, in the linked Chobby branch. It's not merged yet.

L-e-x-o-n: `$mode` implies it's a teiserver command except no such command exists nor is there a PR to add it to teiserver.
I think what you need is just `!mode`.

Teifion: Chobby doesn't implement Teiserver commands though?

keithharvey: I am pretty sure autohost passes through chat_lib.ex @Teifion Raising the limit for welcome_message is another example. So I need to extend the line length in order to get my untrunacted mode through autohost

@L-e-x-o-n you're totally right, dropped $mode, that's intercepted by Chobby.

lib/teiserver/protocols/spring/spring_in.ex
```diff
@@ -1270,6 +1270,9 @@ defmodule Teiserver.Protocols.SpringIn do
           String.starts_with?(lowercase_msg, ["$welcome-message", "!welcome-message"]) ->
             msg |> String.trim() |> String.slice(0..1024)
 
+          String.starts_with?(lowercase_msg, "!mode ") ->
+            msg |> String.trim() |> String.slice(0..1024)
+
```
Teifion: These should be combined into a single block, with `!mode` being added to the above list. Given the length of that list at this stage it should be swapped into a module attribute further up.

lib/teiserver/protocols/spring/spring_in.ex
L-e-x-o-n: > Chobby doesn't implement Teiserver commands though?

No it doesn't and this isn't a Teiserver command.

lib/teiserver/protocols/spring/spring_in.ex
keithharvey: Done

lib/teiserver/protocols/spring/spring_in.ex
```diff
@@ -47,6 +47,10 @@ defmodule Teiserver.Protocols.SpringIn do
 
   @cluster_manager_regex ~r/^Host\[[A-Z]+\d+\]$/
 
+  # Battle chat commands that carry more data than a normal message (e.g. modoption
+  # key=value pairs) and so get a 1024-char allowance instead of the default 256.
```
keithharvey, line 51: The space is intentional, since I could see other commands that start with "mode" in the future.

Teifion (approved):

PR #1286, opened 2026-06-15
Fixed mfa auth check for bots

Teifion:
Bots do not have MFA enabled (and cannot have it enabled) so would always fail the check. This addresses that.


L-e-x-o-n (commented):
Can you add a test please?

L-e-x-o-n (approved):

PR #1294, opened 2026-06-16
Preventing duplication of game rating logs

### LINKED ISSUE #433: [Bug]: Some games are ranked multiple times
### Describe the Bug

Run into it when I was performing data analysis. There are matches that contain multiple entries in `teiserver_game_rating_logs` for the same match, for example `match_id=2283667` from 2024-03-09. There are 120 games like this, one game is even rated 4 times.

### Reproduce the bug

Query to run e.g. on integration server to pull match ids:
```sql
WITH
dups AS (
    SELECT
        count(*) AS c,
        match_id,
        user_id
    FROM
        teiserver_game_rating_logs tgrl
    GROUP BY
        match_id,
        user_id
    HAVING
        count(*) > 1
),
dups_per_match AS (
    SELECT
        max(c) AS c,
        match_id
    FROM
        dups
    GROUP BY
        match_id
),
top_dup_matches AS (
    SELECT
        tbm.id,
        tbm.started,
        c AS duplicates
    FROM
        dups_per_match dpm
    JOIN teiserver_battle_matches tbm ON
        dpm.match_id = tbm.id
    ORDER BY
        duplicates DESC,
        tbm.started DESC
)
SELECT
    *
FROM
    top_dup_matches;
```

### Screenshots

_No response_

### Additional context

_No response_

BElluu:
Closed #433

### DIFF
--- a/priv/repo/migrations/20260616101633_unique_index_rating_logs.exs
+++ b/priv/repo/migrations/20260616101633_unique_index_rating_logs.exs
@@ -1,0 +1,22 @@
+defmodule Teiserver.Repo.Migrations.UniqueIndexRatingLogs do
+  use Ecto.Migration
+
+  def up do
+    # Remove duplicate entries keeping the earliest 
+    # (lowest id) per (match_id, user_id, rating_type_id)
+    execute("""
+    DELETE FROM teiserver_game_rating_logs
+    WHERE id NOT IN (
+      SELECT DISTINCT ON (match_id, user_id, rating_type_id) id
+      FROM teiserver_game_rating_logs
+      ORDER BY match_id, user_id, rating_type_id, id ASC
+    )
+    """)
+
+    create unique_index(:teiserver_game_rating_logs, [:match_id, :user_id, :rating_type_id])
+  end
+
+  def down do
+    drop unique_index(:teiserver_game_rating_logs, [:match_id, :user_id, :rating_type_id])
+  end
+end
--- a/priv/repo/migrations/20260616101633_unique_index_rating_logs.exs
+++ b/priv/repo/migrations/20260616101633_unique_index_rating_logs.exs
@@ -2,7 +2,7 @@ defmodule Teiserver.Repo.Migrations.UniqueIndexRatingLogs do
   use Ecto.Migration
 
   def up do
-    # Remove duplicate entries keeping the earliest 
+    # Remove duplicate entries keeping the earliest
     # (lowest id) per (match_id, user_id, rating_type_id)
     execute("""
     DELETE FROM teiserver_game_rating_logs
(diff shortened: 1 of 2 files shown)


L-e-x-o-n:
We should check how many logs are affected.

This is a good change, but it doesn't address the cause, why is the same match having duplicate logs in the first place?

priv/repo/migrations/20260616101633_unique_index_rating_logs.exs
```diff
@@ -0,0 +1,22 @@
...
+    execute("""
+    DELETE FROM teiserver_game_rating_logs
+    WHERE id NOT IN (
+      SELECT DISTINCT ON (match_id, user_id, rating_type_id) id
+      FROM teiserver_game_rating_logs
+      ORDER BY match_id, user_id, rating_type_id, id ASC
+    )
+    """)
+
+    create unique_index(:teiserver_game_rating_logs, [:match_id, :user_id, :rating_type_id])
```
geekingfrog, line 16: What is the SQL for this? I'm pretty sure this is going to acquire an exclusive lock on the entire table for the duration of the operation. This table has 13M rows at time of writing.
How long would the index creation take? If it's more than a couple of seconds, we need to change it and make it non blocking.

with postgres, you can create an index concurrently, and then create a unique constraint using the existing index.

BElluu:
```sql
CREATE UNIQUE INDEX teiserver_game_rating_logs_match_id_user_id_rating_type_id_index
ON teiserver_game_rating_logs (match_id, user_id, rating_type_id);
```
Creating index for that many rows, can take some minutes.

So maybe we should firstly manualy delete duplicates
```sql
DO $$
DECLARE deleted integer;
BEGIN
  LOOP
    WITH dupes AS (
      SELECT id FROM (
        SELECT id, ROW_NUMBER() OVER (
          PARTITION BY match_id, user_id, rating_type_id
          ORDER BY id ASC
        ) rn
        FROM teiserver_game_rating_logs
      ) s WHERE s.rn > 1
      LIMIT 10000
    )
    DELETE FROM teiserver_game_rating_logs t USING dupes WHERE t.id = dupes.id;
    GET DIAGNOSTICS deleted = ROW_COUNT;
    EXIT WHEN deleted = 0;
    PERFORM pg_sleep(0.1);
  END LOOP;
END $$;
```
and then create index concurently out of transaction without lock

```elixir
defmodule Teiserver.Repo.Migrations.UniqueIndexRatingLogs do
  use Ecto.Migration

  @disable_ddl_transaction true
  @disable_migration_lock true

  def up do
    create unique_index(
             :teiserver_game_rating_logs,
             [:match_id, :user_id, :rating_type_id],
             concurrently: true
           )
  end

  def down do
    drop unique_index(
           :teiserver_game_rating_logs,
           [:match_id, :user_id, :rating_type_id],
           concurrently: true
         )
  end
end
```
What do you think ?

PR #1297, opened 2026-06-17
Rework metadata cache

geekingfrog:
This rework the application metadata cache warmup logic.
The main issue is `startup.ex`, that file should not exist at all and everything should be under the main supervision tree.
This deals with the app metadata cache.

### DIFF
--- a/lib/teiserver/startup.ex
+++ b/lib/teiserver/startup.ex
@@ -6,10 +6,6 @@ defmodule Teiserver.Startup do
   alias Teiserver.Coordinator
   alias Teiserver.Coordinator.AutomodServer
   alias Teiserver.LobbyIdServer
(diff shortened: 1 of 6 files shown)


lib/teiserver/startup.ex
```diff
@@ -37,19 +33,6 @@ defmodule Teiserver.Startup do
       end)
     end
 
-    # If test mode we don't want to load this up for every test as we will generate
-    # a lot of errors related to the SQL sandbox. We have specific unit tests for
-    # these elsewhere in the codebase
```
geekingfrog, line 42: putting the warmup code under the supervision tree seems to fix the issue of sandboxing. I suspect it's because task tracks their caller, though I haven't really checked that.

PR #1299, opened 2026-06-17
Kickban - improvements

BElluu:
@geekingfrog I made some changes. In #1252 you writed target_user can recieve `lobby/left` and `lobby/updated` too.
This doesn't actually happen. The target_user only receives lobby/left, never lobby/updated.

`do_kickban` calls `remove_player_from_lobby/remove_spectator_from_lobby`, which internally run `process_events` + `broadcast_updates`. `broadcast_updates` delegates to `broadcast_to_members`, which iterates over `aggregate.data.players` and `aggregate.data.spectators` - the state, **after** event processing. At that point the target user is already removd from those maps, so the `for {p_id, p} <- state.players/state.spectators` loops skip them entirely. Only after that does `do_kickban` send `{:left, reason, ban_until}` directly to `target_pid`.

I added two tests to confirm that:
- `kickban kicked player receives only left event, no updated` - target join with `self()`, creator gets a separate `sink_pid`. After kickban assert and refute passes, so target user never sees the updated event.
- `kickban remaining members receive updated when player is kicked` - verifies the other side: the creator with `self()` does receive `{:updated, %{spectators: %{"user2" => nil}}}, confrming updated goes to remaining members, not the target.

### DIFF
--- a/lib/teiserver/player/tachyon_handler.ex
+++ b/lib/teiserver/player/tachyon_handler.ex
@@ -757,6 +757,10 @@ defmodule Teiserver.Player.TachyonHandler do
       {:error, :lobby_full} ->
         {:error_response, :lobby_full, state}
 
+      {:error, :banned, ban_until} ->
+        details = "Banned until #{DateTime.to_unix(ban_until, :microsecond)}"
+        {:error_response, :banned, details, state}
+
       {:error, reason} ->
         {:error_response, :invalid_request, to_string(reason), state}
     end
@@ -1005,13 +1009,17 @@ defmodule Teiserver.Player.TachyonHandler do
   end
 
   def handle_command("lobby/kickban", "request", _msg_id, %{"data" => data}, state) do
-    target_id = data["userId"]
-    ban_until = parse_ban_until(data["banUntil"])
+    case TachyonParser.parse_user_id(data["userId"]) do
+      {:ok, target_id} ->
+        ban_until = parse_ban_until(data["banUntil"])
 
-    case Session.lobby_kickban(state.user.id, target_id, ban_until) do
-      :ok -> {:response, state}
-      {:error, :unauthorized} -> {:error_response, :unauthorized, state}
-      {:error, reason} -> {:error_response, :invalid_request, to_string(reason), state}
+        case Session.lobby_kickban(state.user.id, target_id, ban_until) do
+          :ok -> {:response, state}
+          {:error, reason} -> {:error_response, :invalid_request, to_string(reason), state}
+        end
+
+      {:error, err} ->
+        {:error_response, :invalid_request, to_string(err), state}
     end
   end
 
--- a/lib/teiserver/player/tachyon_handler.ex
+++ b/lib/teiserver/player/tachyon_handler.ex
@@ -1015,6 +1015,7 @@ defmodule Teiserver.Player.TachyonHandler do
 
         case Session.lobby_kickban(state.user.id, target_id, ban_until) do
           :ok -> {:response, state}
+          {:error, :not_boss} -> {:error_response, :unauthorized, state}
           {:error, reason} -> {:error_response, :invalid_request, to_string(reason), state}
         end
 
@@ -1401,8 +1402,21 @@ defmodule Teiserver.Player.TachyonHandler do
 
   defp vote_action_to_tachyon(action) do
     case action do
-      {:change_map, name} -> %{type: :changeMap, newMapName: name}
-      {:appoint_boss, boss_id} -> %{type: :appointBoss, bossId: to_string(boss_id)}
+      {:change_map, name} ->
+        %{type: :changeMap, newMapName: name}
+
+      {:appoint_boss, boss_id} ->
+        %{type: :appointBoss, bossId: to_string(boss_id)}
+
+      {:kickban, user_id, nil} ->
+        %{type: :kickban, userId: to_string(user_id)}
+
+      {:kickban, user_id, ban_until} ->
+        %{
+          type: :kickban,
+          userId: to_string(user_id),
+          banUntil: DateTime.to_unix(ban_until, :microsecond)
+        }
     end
   end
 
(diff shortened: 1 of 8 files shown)


lib/teiserver/player/tachyon_handler.ex
```diff
@@ -757,6 +757,10 @@ defmodule Teiserver.Player.TachyonHandler do
       {:error, :lobby_full} ->
         {:error_response, :lobby_full, state}
 
+      {:error, :banned, ban_until} ->
+        details = "Banned until #{DateTime.to_unix(ban_until, :microsecond)}"
```
geekingfrog: put that as human readable time, not timestamp. It's in a string anyway so not machine readable.
We may want to revisit that at some point, but that'll come later (maybe)

lib/teiserver/player/tachyon_handler.ex
```diff
@@ -1005,13 +1009,18 @@ defmodule Teiserver.Player.TachyonHandler do
...
   def handle_command("lobby/kickban", "request", _msg_id, %{"data" => data}, state) do
-    target_id = data["userId"]
-    ban_until = parse_ban_until(data["banUntil"])
+    case TachyonParser.parse_user_id(data["userId"]) do
+      {:ok, target_id} ->
+        ban_until = parse_ban_until(data["banUntil"])
+
+        case Session.lobby_kickban(state.user.id, target_id, ban_until) do
+          :ok -> {:response, state}
+          {:error, :not_boss} -> {:error_response, :unauthorized, state}
```
geekingfrog: not `unauthorized`, `invalid_request` instead.

lib/teiserver/player/tachyon_handler.ex
```diff
@@ -1005,13 +1009,18 @@ defmodule Teiserver.Player.TachyonHandler do
...
+          {:error, reason} -> {:error_response, :invalid_request, to_string(reason), state}
+        end
 
-    case Session.lobby_kickban(state.user.id, target_id, ban_until) do
-      :ok -> {:response, state}
-      {:error, :unauthorized} -> {:error_response, :unauthorized, state}
-      {:error, reason} -> {:error_response, :invalid_request, to_string(reason), state}
+      {:error, err} ->
+        {:error_response, :invalid_request, to_string(err), state}
```
geekingfrog: don't use `to_string` but `inspect`, since `to_string` fails on most things.

lib/teiserver/tachyon_lobby/lobby.ex
```diff
@@ -1679,6 +1710,19 @@ defmodule Teiserver.TachyonLobby.Lobby do
     end
   end
 
+  defp maybe_cancel_kick_vote(leaving_user_id, aggregate) do
+    case aggregate.data.current_vote do
+      nil ->
+        aggregate
+
+      %{action: {:kickban, ^leaving_user_id, nil}} ->
+        process_event({:vote_ended, DateTime.utc_now(), :failed}, aggregate)
```
geekingfrog: there is `:cancelled` vote outcome

lib/teiserver/tachyon_lobby/lobby.ex
```diff
@@ -1679,6 +1710,19 @@ defmodule Teiserver.TachyonLobby.Lobby do
     end
   end
 
+  defp maybe_cancel_kick_vote(leaving_user_id, aggregate) do
+    case aggregate.data.current_vote do
+      nil ->
```
geekingfrog: you don't need 2 clauses `-> aggregate`, you can do

```elixir
    case aggregate.data.current_vote do
      %{action: {:kickban, ^leaving_user_id, nil}} ->
        process_event({:vote_ended, DateTime.utc_now(), :failed}, aggregate)

      _other ->
        aggregate
    end
```

test/teiserver_web/tachyon/lobby_test.exs
```diff
@@ -575,6 +575,42 @@ defmodule TeiserverWeb.Tachyon.LobbyTest do
...
+      {:ok, lobby_data} = setup_lobby(%{client: ctx[:client]}, %{boss_enabled?: true})
+
+      {:ok, ctx2} = Tachyon.setup_client()
+      %{"status" => "success"} = Tachyon.join_lobby!(ctx2[:client], lobby_data[:lobby_id])
+      %{"commandId" => "lobby/updated"} = Tachyon.recv_message!(ctx[:client])
+
+      %{"status" => "success"} = Tachyon.lobby_kickban(ctx[:client], ctx2[:user].id)
+
+      %{"commandId" => "lobby/left", "data" => data} = Tachyon.recv_message!(ctx2[:client])
+      assert data["reason"] == "kicked"
```
geekingfrog, line 594: let's add one more assert for good measure: the remaining player received the corresponding lobby/updated event

test/teiserver/tachyon_lobby/lobby_test.exs
```diff
@@ -1829,6 +1857,15 @@ defmodule Teiserver.TachyonLobby.LobbyTest do
       assert vote.voters[@default_user_id] == :yes
       assert vote.voters["user2"] == :pending
       assert vote.voters["user3"] == :pending
+
+      :ok = Lobby.vote_submit(id, "user3", {vote.id, :yes})
+
+      assert_receive {:lobby, ^id, {:updated, %{current_vote: nil}}}
+      assert_receive {:lobby, ^id, {:vote_ended, _vote_id, :passed}}
```
geekingfrog, line 1865: you should also check you got the corresponding `updated` event when the vote passed

test/teiserver/tachyon_lobby/lobby_test.exs
```diff
@@ -1737,6 +1737,37 @@ defmodule Teiserver.TachyonLobby.LobbyTest do
...
+        |> Map.put(:boss_enabled?, true)
+        |> Map.put(:creator_pid, sink_pid)
+        |> Lobby.create()
+
+      {:ok, _lobby_pid, _details} = Lobby.join(id, mk_player("user2"), self())
+
+      :ok = Lobby.kickban(id, @default_user_id, "user2")
+
+      assert_receive {:lobby, ^id, {:left, "kicked", nil}}
+      refute_receive {:lobby, ^id, {:updated, _}}, 100
```
geekingfrog: Is there a reason you increase the timeout? Ideally we shouldn't have too many of these since they will overtime make the tests run so much slower

PR #1301, opened 2026-06-17
Fixed banned domain check (again), optimise domain and IP checks by u…

L-e-x-o-n:
…sing MapSet

### DIFF
--- a/lib/teiserver/moderation.ex
+++ b/lib/teiserver/moderation.ex
@@ -880,20 +880,37 @@ defmodule Teiserver.Moderation do
     Repo.all(BannedDomain)
   end
 
-  @spec list_banned_domains_cache :: [String.t()]
+  @spec list_banned_domains_cache :: MapSet.t(String.t())
   def list_banned_domains_cache do
-    Teiserver.cache_get(:application_metadata_cache, "banned_domains", [])
+    Teiserver.cache_get(:application_metadata_cache, "banned_domains", MapSet.new())
   end
 
   @spec banned_domain?(String.t()) :: boolean()
   def banned_domain?(email) do
     case String.split(email, "@") do
       [_start, domain] ->
-        String.contains?(domain, list_banned_domains_cache())
+        banned_domains = list_banned_domains_cache()
+
+        domain
+        |> String.downcase()
+        |> domain_segments()
+        |> Enum.any?(&MapSet.member?(banned_domains, &1))
 
       _no_email ->
         false
     end
+  end
+
+  # Given a full email domain such as sub.domain.foo.bar
+  # will return a diminishing list of the dot-separated
+  # parts of the list, e.g.
+  # ["sub.domain.foo.bar", "domain.foo.bar", "foo.bar", "bar"]
+  defp domain_segments(domain) do
+    String.split(domain, ".")
+    |> Enum.reverse()
+    |> Enum.scan(fn elem, acc ->
+      elem <> "." <> acc
+    end)
   end
 
   @doc """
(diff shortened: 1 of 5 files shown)


lib/teiserver/moderation.ex
```diff
@@ -880,22 +880,38 @@ defmodule Teiserver.Moderation do
...
+  defp split_domain_segments(domain) do
+    parts = String.split(domain, ".")
+    count = length(parts)
+
+    Enum.map(0..(count - 1), fn i ->
+      parts
+      |> Enum.drop(i)
+      |> Enum.join(".")
+    end)
+  end
```
geekingfrog: This can be done with [Enum.scan/2](https://elixir.hexdocs.pm/Enum.html#scan/2)

```elixir
  def domain_segments(domain) do
    String.split(domain, ".")
    |> Enum.reverse()
    |> Enum.scan(fn elem, acc ->
      elem <> "." <> acc
    end)
  end
```

```
domain_segments("foo.bar.baz.co.uk")
["uk", "co.uk", "baz.co.uk", "bar.baz.co.uk", "foo.bar.baz.co.uk"]
```

lib/teiserver/moderation.ex
```diff
@@ -880,22 +880,38 @@ defmodule Teiserver.Moderation do
...
   def banned_domain?(email) do
     case String.split(email, "@") do
       [_start, domain] ->
-        String.contains?(domain, list_banned_domains_cache())
+        banned_domains = list_banned_domains_cache()
+
+        domain
+        |> String.downcase()
+        |> split_domain_segments()
+        |> Enum.any?(&MapSet.member?(banned_domains, &1))
```
geekingfrog, line 897: I don't think this is going to work. If you have the following banned domain: "foo.bar", you are going to also ban "foo.bar.legit.tld"
You need to search for suffix match
I think `domain |> domain_segments() |> Enum.any?(&String.ends_with?(&1, banned_domains))` may work?

lib/teiserver/moderation.ex
```diff
@@ -880,22 +880,38 @@ defmodule Teiserver.Moderation do
...
 
       _no_email ->
         false
     end
   end
 
+  defp split_domain_segments(domain) do
```
Teifion: Adding a comment here would aide in comprehension.

```suggestion
  # Given a full email such as abc.def@sub.domain.co.uk
  # will return a diminishing list of the dot-separated
  # parts of the list, e.g.
  #   ["abc.def@sub.domain.co.uk", "def@sub.domain.co.uk", "domain.co.uk", "co.uk", "uk"]
  defp split_domain_segments(domain) do
```

lib/teiserver/moderation.ex
L-e-x-o-n: I can try, I considered this but thought a banned subdomain of a legit one wouldn't be a common case to cause issues.

lib/teiserver/moderation.ex
L-e-x-o-n: Nice, I wasn't aware of `scan`.

Teifion (commented):
Great use of MapSet and something I should have thought of. Only thing I'd request is adding the comment block to make it easier to instantly understand what the `split_domain_function` is doing.

Teifion (approved):

lib/teiserver/moderation.ex
Teifion: If "com" is added as a domain then any .com address would be banned. I think it comes down to user training and ensuring they understand what will and will not be banned. Maybe some sort of live tester to see if a given domain would get banned? (future PR)

lib/teiserver/moderation.ex
```diff
@@ -880,22 +880,39 @@ defmodule Teiserver.Moderation do
...
 
       _no_email ->
         false
     end
   end
 
+  # Given a full email domain such as sub.domain.foo.bar
+  # will return a diminishing list of the dot-separated
+  # parts of the list, e.g.
+  # ["sub.domain.foo.bar", "domain.foo.bar", "foo.bar", "bar"]
```
Teifion: With the use of reverse/scan the order has now changed.
```suggestion
  #   ["bar", "foo.bar", "domain.foo.bar", "sub.domain.foo.bar"]
```

PR #1307, opened 2026-06-21
Add ability to show/hide moderator status

Teifion:
With this change, when a user with moderator powers logs in they will by default not show to other users as a moderator. They can enable showing as a moderator with `$modme` and hide it again with `$unmodme`.

This is a visual change only, SPADS and others bots will see the correct moderator status as will all server auth checks.

### DIFF
--- a/lib/teiserver/data/client.ex
+++ b/lib/teiserver/data/client.ex
@@ -23,6 +23,9 @@ defmodule Teiserver.Client do
         away: false,
         rank: 1,
         moderator: false,
+
+        # Used to control showing/hiding the moderator-ness of the client
+        show_moderator: false,
         bot: false,
 
         # Battle stuff
@@ -97,6 +100,7 @@ defmodule Teiserver.Client do
         tcp_pid: self(),
         rank: user.rank,
         moderator: Auth.moderator?(db_user) or Auth.is_event_organizer?(db_user),
+        show_moderator: false,
         bot: Auth.is_bot?(db_user),
         away: false,
         in_game: false,
@@ -197,6 +201,9 @@ defmodule Teiserver.Client do
   @spec update(map(), :silent | :client_updated_status | :client_updated_battlestatus) ::
           T.client()
   def update(client, reason), do: ClientLib.replace_update_client(client, reason)
+
+  @spec merge_update_client(User.id(), map()) :: nil | :ok
+  defdelegate merge_update_client(user_id, updates), to: ClientLib
 
   @spec get_client_pid(User.id()) :: pid() | nil
   defdelegate get_client_pid(userid), to: ClientLib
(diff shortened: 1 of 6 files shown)


StanczakDominik (approved):

lib/teiserver/data/client.ex
```diff
@@ -97,6 +100,7 @@ defmodule Teiserver.Client do
         tcp_pid: self(),
         rank: user.rank,
         moderator: Auth.moderator?(db_user) or Auth.is_event_organizer?(db_user),
+        show_moderator: false,
```
Teifion, line 103: Default to not showing it at login

lib/teiserver/protocols/spring/spring.ex
```diff
@@ -19,17 +19,24 @@ defmodule Teiserver.Protocols.Spring do
...
+  @spec create_client_status(map(), boolean()) :: Integer.t()
+  def create_client_status(client, true_mod_status \\ false) do
     [r1, r2, r3] = BitParse.parse_bits("#{client.rank || 1}", 3)
 
+    moderator =
+      if true_mod_status do
+        client.moderator
+      else
+        Map.get(client, :show_moderator, client.moderator)
+      end
```
Teifion, line 31: This is the part controlling what is shown to other clients., If `true_mod_status` is true we will always show the moderation status of the client. If not we will get the `show_moderator` value with a fallback to their true status.

test/teiserver/coordinator/coordinator_commands_sync_test.exs
```diff
@@ -50,5 +50,43 @@ defmodule Teiserver.Coordinator.CoordinatorCommandsSyncTest do
...
       Application.put_env(:teiserver, Teiserver, config)
     end
+
+    test "modme - not moderator", %{socket: socket, user: user} do
+      client = Account.get_client_by_id(user.id)
+      refute client.moderator
+      refute client.show_moderator
+
+      _send_raw(socket, "SAYPRIVATE coordinator $modme\n")
+      :timer.sleep(100)
```
geekingfrog, line 64: the sleeps aren't ideal, there are a few helper function under `Teiserver.Support.Polling` that you should use instead, it'll help reduce the duration of tests.

test/teiserver/coordinator/coordinator_commands_sync_test.exs
geekingfrog, line 1: Since you are testing at the socket level, shouldn't you check you receive the correct status message through a socket instead of reaching for the internals?

None of these test really make sure other clients will see the correct status for the client you are modifying.

Teifion:
Based on the way the tests are going this feature is not working, I'll pause work on it for now and may drop it entirely. If anybody wants to take over the PR (or replace it entirely) I would fully support that.

steeldesade:
The more I think about this, the more I like the idea. I would probably play more if it were not so obvious I was a moderator (who is not great at the game).

Teifion:
While there is still a desire for this to happen, this PR is not going to be the PR to do it.

PR #1311, opened 2026-06-24
UI improvements on moderation pages

Snipebusher:
Improve UI of moderation panels

# 1.
#### BEFORE
image
image

#### AFTER
image

### DIFF
--- a/assets/scss/custom/_styles.scss
+++ b/assets/scss/custom/_styles.scss
@@ -307,7 +307,7 @@ body {
 }
 
 .row-left {
-  width: 145px;
+  width: 160px;
 }
 
 body.modal-open {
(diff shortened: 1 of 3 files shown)


Snipebusher:
# 2.

#### BEFORE
image

#### AFTER
image

assets/scss/custom/_styles.scss
```diff
@@ -307,7 +307,7 @@ body {
 }
 
 .row-left {
-  width: 145px;
+  width: 160px;
```
Teifion, line 310: This class is used in a lot of files other than the moderation ones; are you intending to change it for them too as part of the PR?

Snipebusher: Could you point out an example please? It may be due to my inexperience with this, but I couldn't find any...

Teifion: This is a CSS class selector, row-left class is used in a few different files.

PR #1314, opened 2026-06-28
Struct for lobby events

geekingfrog:
This looks massive but is actually "just" a refactoring of how lobby processes events. Instead of having a big union of types, each event is a separate struct.
These events implement a protocol to apply it to an aggregate. The aggregate is a data structure that represent a change: the new state alongside updates and side effects.

The separate commits are more for review and development, all of that should be squashed when merged since it's "just" one big refactor.

### DIFF
--- a/test/teiserver/tachyon_lobby/lobby_test.exs
+++ b/test/teiserver/tachyon_lobby/lobby_test.exs
@@ -1127,8 +1127,6 @@ defmodule Teiserver.TachyonLobby.LobbyTest do
       assert result == :failed
     end
 
-    # re-enable when voting events are converted to structs
-    @tag :skip
     test "voter disconnect is the same as leaving (abstain)" do
       %{id: id, users: users} = setup_full_lobby([1, 1])
       :ok = Lobby.join_queue(id, "2")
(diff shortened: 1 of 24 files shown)


test/teiserver/tachyon_lobby/lobby_test.exs
```diff
@@ -1860,8 +1860,7 @@ defmodule Teiserver.TachyonLobby.LobbyTest do
 
       :ok = Lobby.vote_submit(id, "user3", {vote.id, :yes})
 
-      assert_receive {:lobby, ^id, {:updated, %{players: %{"user2" => nil}}}}
-      assert_receive {:lobby, ^id, {:updated, %{current_vote: nil}}}
+      assert_receive {:lobby, ^id, {:updated, %{players: %{"user2" => nil}, current_vote: nil}}}
```
geekingfrog, line 1863: This is the only change to tests, since the updates are correctly batched now, there is no more 2 messages.

Teifion (approved):
I've taken the time to review this as best I can but the size of it (broken up into commits was incredibly helpful, thank you) means I do not have the normal confidence when pushing the approve button. I think given this is not used in production at scale yet it is safe to merge even with such reservations though.

PR #1316, opened 2026-06-28
Adds rate limiting for user logins

Teifion:
* Uses Hammer library to perform rate limiting
* Affects both Web interface and Spring interaface
* Intended to be a good first step rather than a permanent approach


lib/teiserver/general/rate_limit.ex
geekingfrog, line 1: I am not a fan of this file. I think it should be in auth lib, or account related stuff, but not under a general rate_limter module. The whole logic is about login. This module should just be used to start the rate limiter process.
So basically, drop everything but `use Hammer`, and move the rest elsewhere together with the login logic.

lib/teiserver/account/libs/user_lib.ex
```diff
@@ -367,7 +369,12 @@ defmodule Teiserver.Account.UserLib do
         {:error, "Invalid credentials"}
 
       user ->
-        authenticate_user(conn, user, plain_text_password)
+        if RateLimit.can_login?(user.id) do
+          authenticate_user(conn, user, plain_text_password)
+        else
+          Telemetry.log_simple_server_event(user.id, "Login rate limit")
```
geekingfrog: I'd like to move away from these custom events, especially when there is a direct equivalent with open telemetry. The advantage of having these exposed through metrics is that we can use the powerful grafana UI and not care about anything else.

lib/teiserver/general/rate_limit.ex
```diff
@@ -0,0 +1,24 @@
...
+  @moduledoc """
+  Allows for rate limiting various functionality within the application.
+  """
+
+  alias Teiserver.Account.User
+  alias Teiserver.General.RateLimit
+
+  use Hammer, backend: :ets
+
+  defp failed_login_key(user_id), do: "failed-login:#{user_id}"
```
geekingfrog: might as well inline that function.

Teifion: Addressed in https://github.com/beyond-all-reason/teiserver/pull/1316/commits/55212c08b0804d0165735302541344dab24d2385

lib/teiserver/general/rate_limit.ex
Teifion: Addressed in https://github.com/beyond-all-reason/teiserver/pull/1316/commits/55212c08b0804d0165735302541344dab24d2385

lib/teiserver/account/libs/user_lib.ex
Teifion: Addressed in https://github.com/beyond-all-reason/teiserver/commit/55212c08b0804d0165735302541344dab24d2385

I have added it to the `tachyon login error` namespace as that seemed appropriate but if you feel it would be more appropriate to be in the internal module under a different namespace please let me know.

geekingfrog (approved):

PR #1321, opened 2026-06-30
Added al_check

Teifion:
To allow the partitioned tests I had to follow [this guide](https://oneuptime.com/blog/post/2026-01-25-manage-connection-limits-postgresql/view) to enable additional connections.

Timing check against precommit we can see a very nice improvement
```
time mix precommit
________________________________________________________
Executed in  249.07 secs    fish           external
   usr time  130.13 secs  370.00 micros  130.13 secs
   sys time   12.14 secs    0.00 micros   12.14 
```

```
time check
________________________________________________________
Executed in  132.55 secs      fish           external
   usr time  948.89 millis  328.00 micros  948.57 millis
   sys time  167.67 millis   66.00 micros  167.60 millis
```

The main downside for me is the test results don't get printed to the terminal but at this stage I think it is worth considering for inclusion in the project with later work being to:
- Break up slower tests into parts
- Include in CI


config/test.exs
```diff
@@ -1,14 +1,18 @@
...
+
 import Config
 
 # This makes anything in our tests involving user passwords (creating or logging in) much faster
 config :argon2_elixir, t_cost: 1, m_cost: 8
 
+partition = System.get_env("MIX_TEST_PARTITION", "0") |> String.to_integer()
+partition_port = partition * 100
+# <> partition
+database_url = "postgresql://teiserver_test:123456789@localhost:5432/teiserver_test"
```
LKlemens: please try to run partitions with separate dbs 
```
partition = System.get_env("MIX_TEST_PARTITION", "0")
partition_port = String.to_integer(partition) * 100
database_url = "postgresql://teiserver_test:123456789@localhost:5432/teiserver_test" <> partition
```
that change speeds up running test in partitions x2

.check.json
```diff
@@ -0,0 +1,72 @@
...
+    },
+    "modified_tests": {
+      "name": "Modified Tests",
+      "run": "builtin:modified_tests"
+    }
+  },
+  "coverage": {
+    "html": false,
+    "limit": 80,
+    "mod": "native"
```
LKlemens: calculating coverage takes some time in large project - please consider to delete that config to have fair comparison  with `mix precommit`

.check.json
```diff
@@ -0,0 +1,72 @@
...
+    {
+      "run": "mix format"
+    },
+    {
+      "files": ".check/credo*.txt",
+      "run": "mix recode"
+    }
+  ],
+  "max_concurrency": 10,
+  "partitions": 5,
```
LKlemens: try to start with 3 partitions - often more partitions doesn't mean faster suite

config/runtime.exs
```diff
@@ -252,5 +252,8 @@ if enable_discord_bridge do
     bot_name: Teiserver.ConfigHelpers.get_env("TEI_DISCORD_BOT_NAME")
```
LKlemens, line 252: separate problem is that we have a lot of test modules that last very long 
```
mix test --slowest-modules 5
Top 5 slowest (309.4s), 80.5% of total time:

Teiserver.Protocols.Spring.SpringPartyTest (150sec)
 [test/teiserver/protocols/spring/spring_party_test.exs]
Teiserver.SpringAuthTest (61sec)
 [test/teiserver/protocols/spring/spring_auth_test.exs]
Teiserver.SpringBattleHostTest (52sec)
```

Even if we run every test module in parallel, the total execution time is still limited by the slowest module in the test suite.

If you'd like to reduce the runtime even further, consider splitting the slowest test modules into smaller ones. That allows them to run in parallel, which can further reduce the overall execution time 🤞

Teifion:
@LKlemens thank you for taking the time to go through these, I've gotten the time down to 145s which is probably because of the party test so might break that up. This is significantly faster now and my opinion has done a complete 180.

Of course, I also have to accept the problem was 100% me ;)

.check.json
Teifion: That's a very good point. We don't use the coverage so I have removed that.

.check.json
Teifion: Thank you, after apply all the other fixes I swapped it back to 5 to check and you were completely correct, 3 is a good sweet spot for us right now.

config/test.exs
Teifion: An error from when I was setting it up and didn't correct. Can't believe I missed it!

config/runtime.exs
LKlemens, line 249: > The main downside for me is the test results don't get printed to the terminal 

have u tried `check --verbose` command?


or `check` in terminal1
and `check --watch` in terminal2

LKlemens: In CI you can split checks in two steps 

```
1. testing 
check --only test --verbose --partitions 3 --no-coverage || check --failed --quiet

2. rest
check --only format,credo,dialyzer

```

Teifion: I gave verbose a go but because of the concurrent nature of the runs the output is a bit of a mess. Is there a way to make it so the output gets posted after each section is run? Is it possible to make it so output is printed only when a section fails maybe?

Thank you for continuing to answer questions, it's most appreciated.

LKlemens: I see, I added new `--verbose-sections` flag - please try to update check `mix check.update` and let me know if it works for u 🤞 
and if u want to set that behavior as default u can update check.json with 
>   "test_output" => "sections"`

Teifion: The `--verbose-sections` is working very nicely and answers my request. The edit you made to control it as a default via the json is the final part to make it perfect. Thank you :)

Teifion:
@L-e-x-o-n and @geekingfrog what else would you like to see in this PR prior to merge? As mentioned inclusion in CI would be done in a subsequent PR.

documents/guides/local_setup.md
```diff
@@ -72,6 +72,14 @@ ALTER USER teiserver_dev WITH SUPERUSER;
...
 CREATE USER teiserver_test WITH PASSWORD '123456789';
 ALTER USER teiserver_test WITH SUPERUSER;
+
+-- If using partitioned databases
+CREATE USER teiserver_test0 WITH PASSWORD '123456789';
+ALTER USER teiserver_test0 WITH SUPERUSER;
+CREATE USER teiserver_test1 WITH PASSWORD '123456789';
+ALTER USER teiserver_test1 WITH SUPERUSER;
+CREATE USER teiserver_test2 WITH PASSWORD '123456789';
+ALTER USER teiserver_test2 WITH SUPERUSER;
```
LKlemens, line 80: we probably need users 1,2 and 3 - because if partition is 0 - we do not have MIX_TEST_PARTITION set 🤞

Teifion: We shouldn't need it as when not set it will default to 0 (and given the partitions start at 0 we only need to go to 2).

To be sure I will delete my test-users and recreate them to test it (I have extra users as I was using more partitions previously).

LKlemens: partitions starts from 1 😬

LKlemens: ` test_partition_2.txt test_partition_3.txt test_partition_1.txt`

Teifion: I added something to print out the partition and you are right, there was a test that was not set and that's what was throwing me off. I'll tweak that as part of the test.exs at the same time. Thank you.

Teifion:
Discussed with the team about this tool and while it is nice we didn't feel it added a huge amount of value when weighed against the fact the library was clearly written using an LLM.

Worse yet, [later commits](github.com/LKlemens/al_check/commit/fb534e2850d3e6c2466e3c7c3279e7697614a2cb) were made specifically to cover up that an LLM was used. This introduces, to our mind, a moral hazard and we won't be making use of this library.

Screenshot of commit for posterity:
image

LKlemens:
Hey,

I just wanted to say that I feel I was treated unfairly.
Yes, I use LLMs every day, both at work and on my side projects. They help me implement my ideas by offloading much of the coding, but the ideas, design, and decisions are still mine.

I've spent a lot of time over the past few months building al_check into what it is today. While I use an LLM as a tool to speed up implementation, I still spend my time designing features, reviewing the generated code, testing, debugging, refining the implementation, and making all of the engineering decisions. We use al_check in my team, and I thought it might be useful for others as well. That's the only reason I shared it. 

I also spent my free time helping you set it up, added new feature u were missing and I used my own time and expertise to try to help resolve the distributed nodes issue.
It was not - "build al_check, or fix horde issue" prompts.

It's completely fine if you decided not to use it - that's entirely your choice, but I feel like things went a step further, and in return I was treated as if I had tried to trick or scam you. I don't think that's fair.  


I added a note to the [README](https://github.com/LKlemens/al_check/commit/47e23faa984279c6ab1f6ae6680e55d74f9a0e1c) to avoid any confrontation like this in the future

Screenshot 2026-07-09 at 10 28 28

PR #1323, opened 2026-07-01
Added warning for actions on VIP users

Teifion:
Adds a block of text in an info window for any user with VIP status flagging that there might need to be additional communication around the action being applied to them.

image
image


PtaQQ:
Looks good

L-e-x-o-n (approved):

PR #1325, opened 2026-07-01
Added CSV output to server log list

Teifion:
As part of a request from PtaQ we need to be able to export this data in a repeatable manner. Long term we will want to use a proper ETL pipeline and completely refactor this so as a short term solution I have implemented this.

### DIFF
--- a/lib/teiserver_web/controllers/logging/server_log_controller.ex
+++ b/lib/teiserver_web/controllers/logging/server_log_controller.ex
@@ -341,6 +341,7 @@ defmodule TeiserverWeb.Logging.ServerLogController do
     |> assign(:logs, logs)
     |> assign(:filter, filter)
     |> assign(:unit, unit)
+    |> assign(:raw_data, data_as_raw_csv(logs))
     |> add_breadcrumb(name: "List #{unit}", url: conn.request_path)
     |> render("metric_list.html")
   end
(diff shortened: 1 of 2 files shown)


lib/teiserver_web/controllers/logging/server_log_controller.ex
```diff
@@ -533,4 +534,36 @@ defmodule TeiserverWeb.Logging.ServerLogController do
...
+            data["aggregates"]["minutes"]["player"],
+            data["aggregates"]["minutes"]["lobby"],
+            data["aggregates"]["minutes"]["spectator"]
+          ])
+
+        [
+          Calendar.strftime(row.date, "%Y-%m-%d"),
+          data["aggregates"]["stats"]["unique_users"],
+          data["aggregates"]["stats"]["peak_user_counts"]["total"],
+          round(total_minutes / 60),
```
L-e-x-o-n, line 561: It's under the header of play time but it includes both lobby and spectator time, is this intentional?

Teifion: Yes, PtaQ has specifically requested this be the case. I will tweak the heading to "Active time" to help avoid confusion.

L-e-x-o-n (approved):

PR #1328, opened 2026-07-02
Replace last uses of routes module

### LINKED ISSUE #1147: Replace all usage of `Routes.some_url_path` with  ~p"/some_url_path"
We currently have a lot of instances of using `Routes.some_url_path` that should instead use the ~p"/some_url_path". This can be done piecemeal making it perfect for first time contributors.

Suggested workflow:
- Search for an instance of `Routes.`
- Take the path name after it and put it into `mix phx.routes | grep <path goes here>`
- Use the appropriate URL spat out from that
- Test the URL change is correct

Teifion:
This PR finally removes the last of the usage of `Routes.` to close https://github.com/beyond-all-reason/teiserver/issues/1147

I found the un-used files because they had an instance of `Routes` in them but the route didn't actually exist. It appears the DeadView pages were replaced by LiveView ones and I never removed the DeadView pages.


Teifion:
Latest commits get us down to 17 results across 8 files and fixes a bug with some of the `*_with_user` routes. At this stage I will move over the remaining routes and close the issue as part of the PR.

L-e-x-o-n (approved):

PR #1338, opened 2026-07-06
More mm tests

geekingfrog:
According to Hectate, there are some issues with how `lost` events are processed, so adding more tests. So far, can't reproduce his error.


Hectate:
Apologies; there was a client-side IPC issue. See https://github.com/beyond-all-reason/bar-lobby/pull/628 for that fix.

geekingfrog:
I'm still going to merge these tests, they can be useful to build upon without having to setup a whole ws connection.

Issue #1146, opened 2026-05-11
[Task] Remove CacheUser


Teifion:
Currently we have both `User` and `CacheUser` which are meant to represent the same thing in different formats. Ideally we would remove it and instead just use `User`.

This would also be a good opportunity to start making use of structs in the code to make the usage more explicit and enabling more compile-time checks.

Suggested steps:
- Remove any extra fields from `CacheUser` we can
- Any remaining extra fields become columns in the user table
- Duplicate data from the extra fields into said columns
- Calls to retrieve a `CacheUser` now return a `User` instead
- Remove `CacheUser` modules

PR #1341, opened 2026-07-07
Bring User struct up to parity with CacheUser fields

Teifion:
This is the first stage of removing the CacheUser struct. It aims to allow User structs to be used in place of CacheUser structs without needing to change any other code.

For day to day running the data will be double-written, first to the existing column in the data field but then also to the new extra columns. For users not currently active there is a long running function which will allow the entire table to be updated.

This commit is expected to introduce no behavioural changes to the general system aside from the double-writing of the user data.


Teifion:
Makes progress on https://github.com/beyond-all-reason/teiserver/issues/1146

L-e-x-o-n (approved):

Teifion:
I will merge this after the successful deployment of the current changes to avoid changing too much at once.

PR #1348, opened 2026-07-08
Cowboy rollback and additional spads_controller tests

Teifion:
The recent phoenix version bump introduced an updated cowboy version and that broke forms sending `POST` data.


lib/teiserver/protocols/spring/spring.ex
```diff
@@ -212,6 +212,8 @@ defmodule Teiserver.Protocols.Spring do
   @doc """
   Takes zipped and base64'd data and tries to extract it
   """
+  def read_compressed_base64(nil), do: {:error, "base64 decode error"}
```
Teifion, line 215: If this is called with `nil` it will error. In theory this should never happen but I'd rather return the error tuple than raise as it _can_ happen if you send bad data on purpose.

lib/teiserver_web/controllers/api/spads_controller.ex
```diff
@@ -56,13 +59,15 @@ defmodule TeiserverWeb.API.SpadsController do
     server_balance_enabled = Config.get_site_config_cache("teiserver.Enable server balance")
 
     raw_player_data =
-      params["players"]
+      params
```
Teifion, line 62: If you submit no player data you will get an exception, this solves that. Same for bots below.

L-e-x-o-n (approved):

PR #1353, opened 2026-07-10
Mark many CacheUser functions as deprecated

Teifion:
The intent is to replace all CacheUser related functions, as such those responsible for returning a CacheUser rather than a User are being renamed to allow new functions to replace them where User objects will instead be returned.

### DIFF
--- a/lib/teiserver/account.ex
+++ b/lib/teiserver/account.ex
@@ -2150,29 +2150,26 @@ defmodule Teiserver.Account do
   @spec get_userid_from_name(String.t()) :: integer() | nil
   def get_userid_from_name(name), do: UserCacheLib.get_userid(name)
 
-  @spec get_user_by_name(String.t()) :: T.user() | nil
-  defdelegate get_user_by_name(username), to: UserCacheLib
+  @spec deprecated_get_user_by_name(String.t()) :: T.user() | nil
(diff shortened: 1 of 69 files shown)


geekingfrog:
> The next PR will git blame ignore these changes.

why not the next commit? doesn't that also work? But also, I'm not sure we want to ignore these changes. They are not just reformatting, there is intent and a plan behind, and that shouldn't be hidden.

geekingfrog:
That seems fine to me, though I'm not sure about the value it brings.
Wouldn't it be possible to modify these functions piecemeal to return User instead of CacheUser instead of first renaming them?
In a different project we were using [semgrep](https://github.com/semgrep/semgrep) in CI to prevent the introduction of pattern we deemed deprecated, while chipping away at replacing existing usage. Although, seeing there are 3 people working on this codebase, I don't even think this is that useful. If someone adds new call sites for these function, we can catch that in code review.

Teifion:
> Wouldn't it be possible to modify these functions piecemeal to return User instead of CacheUser instead of first renaming them?

The first attempt was to perform the chnages wholesale as you suggest but it caused so many breakages I am reverting to a "little by little" process.

geekingfrog (approved):

PR #1355, opened 2026-07-10
Multinode poc

geekingfrog:
This is some proof of concept for going multinode. It is around the lobbies since they already have an event sourcing architecture that makes it easy to replicate.

The first commit has some documentation about the design of state replication, it's probably a good place to start.
There are some funky things going on with `poll_module_function` because of how the `:peer` module works and interact with node monitoring from `:net_kernel`.

I haven't tried to make it generic. It can be a bit difficult, especially when I'm not sure how the pattern can be abstracted away from lobby specific stuff.

There are a few things out of scope for this PR:
* when things go wrong. If the cluster topology changes mid-way through a request, or a replica dies at the wrong time. These will need to be addressed but I think we can do that without significant changes.
* setting up the cluster and its topology. We'll probably use libcluster, but for now there is nothing automated.
* Not all lobby features are supported. Most notably, starting a battle currently isn't replicated. It's a bit complex and we can add that later.


ChocoboCoding:
Teifion pointed me here for some gotchas about Nostrum specific multi-node stuff.

Nostrum will need some minor changes to work properly across multiple nodes.
That being said, it might also be worth looking into v0.11.0 which is currently under development, as that introduces [seamless migration](https://kraigie.github.io/nostrum/multi_node.html#migrating-bots-between-nodes) of the bot between nodes without losing any state of events

PR #1358, opened 2026-07-11
Add start and stop server tasks for devcontainer

WybrenKoelmans:
Introduce tasks to start and stop the Teiserver Phoenix server within the development container. Use "Run Task/[Start/Stop/Restart]" or "CTRL+SHIFT+B" to restart the server. 

AI disclaimer: Claude Opus 5


geekingfrog:
I think you're the only one using dev containers, so if that works let's merge it

geekingfrog (approved):

PR #1360, opened 2026-07-11
Fix "Sprint" typo inside two files

PrivacyIsARight:


Teifion:
Provisionally a yes from me though we will need to check the value stored on the server and be prepared to re-set it once this change goes live as a possible side-effect.

geekingfrog:
The current value is the default value, so this is fine.
Do we have a way to purge old values from the db?

Teifion:
There is not a way to purge old values, they would need to be done manually.

L-e-x-o-n (approved):

PR #1361, opened 2026-07-12
Change verifier regex to match the entire string

PrivacyIsARight:


lib/teiserver/o_auth.ex
```diff
@@ -332,7 +332,8 @@ defmodule Teiserver.OAuth do
       s > 128 ->
         {:error, "verifier cannot be more than 128 chars"}
 
-      not String.match?(verifier, ~r/[A-Za-z0-9\-._~]/) ->
+      # Anchor with \A/\z so the whole string must be valid, not just one character
+      not String.match?(verifier, ~r/\A[A-Za-z0-9\-._~]+\z/) ->
```
geekingfrog, line 336: there is only one line, so `^` and `$` are fine. Also, the comment should be the commit message, it makes sense there, but not really in the code itself

geekingfrog:
You know commits can have a description as well right?

Usually you keep the commit title (first line) to a short description, then leave a blank line, and then you can write whatever you want.

geekingfrog (approved):

PR #1362, opened 2026-07-12
Fix renaming

ChocoboCoding:
This fixes various `user_lib` functions not having the naming restrictions in place for usernames.

edit: Since lots of tests use duplicate names (and other things that are forbidden with the naming rules) I have to update a lot of other tests as well

### DIFF
--- a/lib/teiserver/account/schemas/user.ex
+++ b/lib/teiserver/account/schemas/user.ex
@@ -88,9 +88,9 @@ defmodule Teiserver.Account.User do
 
     user
     |> cast(
-         attrs,
-         ~w(name email password icon colour data roles permissions restrictions restricted_until shadowbanned last_login last_played last_logout discord_id discord_dm_channel_id steam_id smurf_of_id country bot email_change_code last_login_mins lobby_hash chobby_hash lobby_client discord_dm_channel)a
-       )
+      attrs,
+      ~w(name email password icon colour data roles permissions restrictions restricted_until shadowbanned last_login last_played last_logout discord_id discord_dm_channel_id steam_id smurf_of_id country bot email_change_code last_login_mins lobby_hash chobby_hash lobby_client discord_dm_channel)a
+    )
     |> validate_required([:name, :email, :password, :permissions])
     |> unique_constraint(:email)
     |> validate_change(:name, fn :name, name ->
@@ -110,9 +110,9 @@ defmodule Teiserver.Account.User do
 
     user
     |> cast(
-         attrs,
-         ~w(name email password icon colour data roles permissions restrictions restricted_until shadowbanned last_login last_played last_logout discord_id discord_dm_channel_id steam_id smurf_of_id rank country bot email_change_code last_login_mins lobby_hash chobby_hash lobby_client discord_dm_channel)a
-       )
+      attrs,
+      ~w(name email password icon colour data roles permissions restrictions restricted_until shadowbanned last_login last_played last_logout discord_id discord_dm_channel_id steam_id smurf_of_id rank country bot email_change_code last_login_mins lobby_hash chobby_hash lobby_client discord_dm_channel)a
+    )
     |> validate_required([:name, :email, :password, :permissions])
     |> unique_constraint(:email)
     |> validate_change(:email, fn :email, email ->
@@ -143,9 +143,9 @@ defmodule Teiserver.Account.User do
 
     user
     |> cast(
-         attrs,
-         ~w(name email icon colour data roles permissions)a
-       )
+      attrs,
+      ~w(name email icon colour data roles permissions)a
+    )
     |> validate_required(~w(name email)a)
     |> unique_constraint(:email)
     |> validate_change(:name, fn :name, name ->
@@ -185,9 +185,9 @@ defmodule Teiserver.Account.User do
         |> cast(attrs, [:name, :email])
         |> validate_required([:name, :email])
         |> add_error(
-             :password_confirmation,
-             "Please enter your password to change your account details."
-           )
+          :password_confirmation,
+          "Please enter your password to change your account details."
+        )
 
       Account.verify_plain_password(attrs["password"], user.password) == false ->
         user
@@ -224,9 +224,9 @@ defmodule Teiserver.Account.User do
         user
         |> change_plain_password(attrs)
         |> add_error(
-             :password_confirmation,
-             "Please enter your existing password to change your password."
-           )
+          :password_confirmation,
+          "Please enter your existing password to change your password."
+        )
 
       Account.verify_plain_password(attrs["existing"], user.password) == false ->
         user
@@ -272,9 +272,9 @@ defmodule Teiserver.Account.User do
 
     user
     |> cast(
-         attrs,
-         ~w(name email password icon colour data roles permissions restrictions restricted_until shadowbanned last_login last_played last_logout discord_id discord_dm_channel_id steam_id smurf_of_id)a
-       )
+      attrs,
+      ~w(name email password icon colour data roles permissions restrictions restricted_until shadowbanned last_login last_played last_logout discord_id discord_dm_channel_id steam_id smurf_of_id)a
+    )
     |> validate_required([:name, :email, :password, :permissions])
     |> unique_constraint(:email)
     |> validate_change(:email, fn :email, email ->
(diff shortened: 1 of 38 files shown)


lib/teiserver/account/schemas/user.ex
```diff
@@ -93,6 +93,12 @@ defmodule Teiserver.Account.User do
     )
     |> validate_required([:name, :email, :password, :permissions])
     |> unique_constraint(:email)
+    |> validate_change(:name, fn :name, name ->
+      case Account.valid_name?(name, false) do
+        :ok -> []
+        {:error, reason} -> [{:name, reason}]
+      end
+    end)
```
Teifion: This is used multiple times, I'd rather see:

```ex
defp validate_name_change(%Changesset{} = changeset, name) do
  case Account.valid_name?(name, false) do
        :ok -> []
        {:error, reason} -> [{:name, reason}]
      end
end
```

Then you can use
```ex
|> validate_name_change()
```

lib/teiserver/account/tasks/gdpr_forget_task.ex
```diff
@@ -65,6 +64,11 @@ defmodule Teiserver.Account.Tasks.GdprForgetTask do
   end
 
   defp forget_user_struct(%User{} = user) do
+    name =
+      (Enum.to_list(?A..?Z) ++ Enum.to_list(?0..?9))
+      |> Enum.take_random(20)
+      |> List.to_string()
```
Teifion:
```suggestion
     @alphanum ~c"abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"

     1..20
      |> Enum.map(fn _idx -> Enum.random(@alphanum) end)
      |> List.to_string()
```

lib/teiserver/helpers/general_test_lib.ex
```diff
@@ -4,19 +4,25 @@ defmodule Teiserver.Helpers.GeneralTestLib do
...
   def user_fixture, do: make_user(%{"permissions" => []})
 
+  # TODO: This should be replaced by AccountFixtures.user_fixture()
+  # unfortunately they have different signatures so it's not a find
+  # and replace
   def make_user(params \\ %{}) do
     {:ok, u} =
       Account.create_user(%{
-        "name" => params["name"] || "Test",
+        "name" => params["name"] || "TEST_#{Enum.take_random(@validchars, 15)}",
```
Teifion: Not as important as the GDPR forget one but ideally this would be written the same way as that will be changed.

lib/teiserver/helpers/general_test_lib.ex
```diff
@@ -66,7 +72,9 @@ defmodule Teiserver.Helpers.GeneralTestLib do
     login(build_conn(), user.email)
   end
 
-  # @spec general_setup(list, ) :: tuple
+  # TODO: This should be refactored into a single opts argument
+  # with the possibility for roles, flags and other rarely used arguments
+  # being part of that
```
Teifion, line 81: Don't forget to create the github issue for this.

lib/teiserver/moderation/libs/test_lib.ex
```diff
@@ -7,8 +7,8 @@ defmodule Teiserver.Moderation.ModerationTestLib do
   def report_fixture(attrs \\ %{}) do
     {:ok, report} =
       %{
-        reporter_id: GeneralTestLib.make_user().id,
-        target_id: GeneralTestLib.make_user().id,
+        reporter_id: GeneralTestLib.make_user(%{"name" => "reporter"}).id,
+        target_id: GeneralTestLib.make_user(%{"name" => "target"}).id,
```
Teifion: We don't need their names to be specific, can be removed. Same for `ban_fixture` later.

test/support/fixtures/account/account_fixtures.ex
```diff
@@ -18,7 +20,7 @@ defmodule Teiserver.AccountFixtures do
     {:ok, user} =
       attrs
       |> Enum.into(%{
-        name: "Test",
+        name: "TEST_#{Enum.take_random(@validchars, 15)}",
```
Teifion: As above, use a similar method to the GDPR forget.

test/teiserver/account/user_lib_test.exs
```diff
@@ -8,6 +8,93 @@ defmodule Teiserver.Account.UserLibTest do
 
   use Teiserver.DataCase, async: false
 
+  @disallowed_name ".,:;<>{}()+-*/="
+
+  describe "disallow renaming to names with disallowed characters" do
+    # create_ first
+    test "create_user/1" do
+      user_vars = %{name: @disallowed_name, email: "test@test.test", password: "password"}
+      assert {:error, %{errors: [name: _error]}} = UserLib.create_user(user_vars)
```
Teifion: You are not correctly asserting the result here, ideally we would use the [`match?/2`](https://elixir.hexdocs.pm/1.20.2/Kernel.html#match?/2) function

```suggestion
      assert match?({:error, %Changeset{errors: [name: _error]}}, UserLib.create_user(user_vars))
```

test/teiserver/game/libs/match_rating_lib_test.exs
```diff
@@ -185,7 +184,7 @@ defmodule Teiserver.Game.MatchRatingLibTest do
         processed: true,
         game_type: game_type,
         # All rooms are hosted by the same user for now
-        founder_id: AccountFixtures.user_fixture().id,
+        founder_id: user1_id,
```
Teifion: This potentially changes the nature of the test (the founder is now a player), it's not part of the rename stuff so I would suggest reverting this specific change unless.

test/teiserver/game/balancer_server_test.exs
```diff
@@ -64,7 +64,7 @@ defmodule Teiserver.Game.BalancerServerTest do
     users =
       for {rank, i} <- Enum.with_index(list_of_ranks) do
         user =
-          GeneralTestLib.make_user(%{})
+          GeneralTestLib.make_user(%{"name" => "user_with_ranks#{i}"})
```
Teifion: Why do we need to specify a name here?

test/teiserver/moderation/moderation_test.exs
```diff
@@ -39,8 +39,8 @@ defmodule Teiserver.ModerationTest do
       assert {:ok, %Report{} = report} =
                @valid_attrs
                |> Map.merge(%{
-                 "reporter_id" => GeneralTestLib.make_user().id,
-                 "target_id" => GeneralTestLib.make_user().id
+                 "reporter_id" => GeneralTestLib.make_user(%{"name" => "reporter"}).id,
+                 "target_id" => GeneralTestLib.make_user(%{"name" => "target"}).id
```
Teifion: We don't need to specify names here

test/teiserver_web/tachyon/matchmaking_test.exs
```diff
@@ -410,7 +410,7 @@ defmodule Teiserver.Tachyon.MatchmakingTest do
 
   describe "pairing" do
     defp setup_app(_context) do
-      owner = GeneralTestLib.make_user(%{"roles" => ["Verified"]})
+      owner = GeneralTestLib.make_user(%{"name" => "app_owner", "roles" => ["Verified"]})
```
Teifion: The name doesn't need to be set here, no need to change anything.

test/teiserver_web/tachyon/user_test.exs
```diff
@@ -88,7 +88,7 @@ defmodule TeiserverWeb.Tachyon.UserTest do
 
     test "for offline user", %{client: client} do
       other_user =
-        GeneralTestLib.make_user(%{"roles" => ["Verified"]})
+        GeneralTestLib.make_user(%{"name" => "other_user", "roles" => ["Verified"]})
```
Teifion: Name does not need to be specified here.

Teifion (commented):
A few bits of leftover specification from when you were making the name more explicit, these can be removed to trim down the PR a bit. There are one or two functional changes I would like to see but overall this is a great PR and I expect it will be able to be merged in short order.

lib/teiserver/helpers/general_test_lib.ex
ChocoboCoding: Issue https://github.com/beyond-all-reason/teiserver/issues/1373

Teifion (approved):

PR #1368, opened 2026-07-13
Fix lots of typos

PrivacyIsARight:


lib/teiserver_web/templates/logging/match_log/export_form.html.heex
ChocoboCoding, line 1: Rather delete code that is commented out

lib/teiserver/bridge/bridge_server.ex
```diff
@@ -279,7 +279,7 @@ defmodule Teiserver.Bridge.BridgeServer do
     # what we want
 
     # # Currently having an issue where I can't get the ID for the emoji
-    # # luckily we can just put the emjoi character in and it should work
+    # # luckily we can just put the emoji character in and it should work
     # Nostrum.Api.create_reaction(thread.id, message.id, "👍")
     # Nostrum.Api.create_reaction(thread.id, message.id, "👎")
     # Nostrum.Api.create_reaction(thread.id, message.id, "👐")
```
ChocoboCoding, line 285: Considering that this block is 3-4 years old, and even references deprecated functions `(Nostrum.Api.create_reaction/3)` we might want to completely remove it instead, depending on what @Teifion says

Teifion: Get rid of it completely :D

Teifion:
I appreciate the intent to improve things using a spell checker but I would like to request a second review from @PrivacyIsARight, the Nostrum comments suggest a fire and forget approach which is not what we are after here and smells something like an LLM would do. Please check out the repo's AI policy if you are using an LLM.

Teifion:
How about just don't change any of the Nostrum lines and someone else can deal with that later.

lib/teiserver_web/templates/logging/match_log/export_form.html.heex
```diff
@@ -39,7 +39,7 @@
                 selected: @params[:active],
                 enumerable: [
                   %{id: "Raw data", icon: "fa-solid fa-th", colour: fg}
-                  # %{id: "Aggregrate data", icon: "fa-solid fa-check", colour: fg}
+                  # %{id: "Aggregate data", icon: "fa-solid fa-check", colour: fg}
```
Teifion:
```suggestion
                  
```

Teifion (commented):
One thing to tweak and I'm happy to approve it

ChocoboCoding:
> How about just don't change any of the Nostrum lines and someone else can deal with that later.

Dealt with it in https://github.com/beyond-all-reason/teiserver/pull/1374

PR #1370, opened 2026-07-14
Fix teamSize/nbTeams regex + tests

PrivacyIsARight:


test/teiserver/coordinator/spads_parser_test.exs
```diff
@@ -3,18 +3,42 @@ defmodule Teiserver.Coordinator.SpadsParserTest do
   use Teiserver.DataCase, async: true
   import Teiserver.TeiserverTestLib, only: [new_user: 0]
 
-  test "parsing teamSize information" do
-    result = SpadsParser.handle_in("Global setting changed by marseel (teamSize=5)", %{})
+  test "parsing teamSize information 1/3" do
```
geekingfrog: these test names are horrible, they tell you nothing about what is being tested.
You can put all of them inside a `describe "parsing teamSize` and have each test with a better (and short) description.

geekingfrog: things like "single digit", "double digit", "double digit no trailing zero". Basically what the test is about.

geekingfrog (approved):

PR #1375, opened 2026-07-15
Make command parsing case-insensitive

PrivacyIsARight:


Teifion:
I'm a fan of case-sensitive commands personally. It's a command so being specific is a fair requirement. I'm not prepared to die on this hill so if either Lexon or GeekingFrog approve of this then I'm okay with it.

L-e-x-o-n (approved):
I like it, in our case I dont see the benefit of case sensitive commands and it case insensitvity can save some time occasionally when you accidently use the wrongncase.

PR #1376, opened 2026-07-15
Structify party

geekingfrog:
Create structs to represent party data. And while I'm at it, I also cleaned up a bit the party interface to hide the internals.

### DIFF
--- a/.credo.exs
+++ b/.credo.exs
@@ -81,7 +81,7 @@
              ignore_sigils: true,
              ignore_strings: true,
              ignore_urls: true,
-             max_length: 98,
+             max_length: 103,
              priority: :low
            ]},
           {Credo.Check.Readability.ModuleAttributeNames, []},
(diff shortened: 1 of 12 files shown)


.credo.exs
```diff
@@ -81,7 +81,7 @@
              ignore_sigils: true,
              ignore_strings: true,
              ignore_urls: true,
-             max_length: 98,
+             max_length: 103,
```
Teifion, line 84: Why is this being changed?

L-e-x-o-n (approved):

Teifion (commented):
Sorry, I thought I posted this comment a few days ago but only had a "pending" review >:

.credo.exs
geekingfrog: it's in the commit message, the auto-formatter generate such lines, but credo doesn't like it

Teifion: Sorry, I missed that. Approved.

Teifion (approved):

PR #1377, opened 2026-07-15
Replace web uses of deprecated_get_user_by_id

Teifion:
Given the non-critical nature of most of the pages this is a good first step towards pulling in data directly from the database.

A test is added to ensure updating either the CacheUser or User will result in the cache being cleared or updated for the other.

### DIFF
--- a/.dialyzer_ignore.exs
+++ b/.dialyzer_ignore.exs
@@ -160,5 +160,7 @@
   {"lib/teiserver_web/templates/admin/lobby/lobby_chat.html.heex", :call},
   {"lib/teiserver_web/templates/admin/lobby/server_chat.html.heex", :call},
   {"lib/teiserver_web/live/moderation/action_live/smurf_link.ex", :callback_arg_type_mismatch},
-  {"lib/teiserver_web/live/moderation/action_live/smurf_link.ex", :call}
+  {"lib/teiserver_web/live/moderation/action_live/smurf_link.ex", :call},
+  {"lib/teiserver/account/tasks/merge_accounts_task.ex", :no_return},
+  {"lib/teiserver_web/controllers/account/general_controller.ex", :no_return}
 ]
--- a/lib/teiserver/account/libs/user_lib.ex
+++ b/lib/teiserver/account/libs/user_lib.ex
@@ -6,6 +6,7 @@ defmodule Teiserver.Account.UserLib do
   alias Teiserver.Account.Auth
   alias Teiserver.Account.RoleLib
   alias Teiserver.Account.User
+  alias Teiserver.Account.UserCacheLib
   alias Teiserver.Account.UserQueries
   alias Teiserver.CacheUser
   alias Teiserver.Client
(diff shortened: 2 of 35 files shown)


lib/teiserver/account/libs/user_lib.ex
```diff
@@ -103,6 +108,37 @@ defmodule Teiserver.Account.UserLib do
...
+  end
+
+  @doc """
+  Identical to `get_user_by_id/1` but with a raise instead of a nil result in the event of
+  no user found in the database.
+  """
+  @spec get_user_by_id!(User.id()) :: User.t()
+  def get_user_by_id!(user_id) do
+    get_user_by_id(user_id) || raise "No user of the ID #{inspect(user_id)}"
+  end
```
Teifion, line 129: I have added these as cached functions with the `get_user/2` being designed for things where you might want to bring in joins etc.

lib/teiserver/helpers/cache_helper.ex
```diff
@@ -8,7 +8,7 @@ defmodule Teiserver.Helpers.CacheHelper do
     ConCache.get(table, key) || default
   catch
     :exit, :noproc ->
-      Logger.warning("Cache #{table} is down")
+      Logger.warning("Cache #{table} is down get")
```
Teifion: These warning tweaks were because when I made a mistake with the cache in an earlier attempt it wasn't apparent at which stage there was a problem with the cache and the extra context was helpful. I still think it would be helpful so have included it.

.dialyzer_ignore.exs
```diff
@@ -160,5 +160,7 @@
   {"lib/teiserver_web/templates/admin/lobby/lobby_chat.html.heex", :call},
   {"lib/teiserver_web/templates/admin/lobby/server_chat.html.heex", :call},
   {"lib/teiserver_web/live/moderation/action_live/smurf_link.ex", :callback_arg_type_mismatch},
-  {"lib/teiserver_web/live/moderation/action_live/smurf_link.ex", :call}
+  {"lib/teiserver_web/live/moderation/action_live/smurf_link.ex", :call},
+  {"lib/teiserver/account/tasks/merge_accounts_task.ex", :no_return},
+  {"lib/teiserver_web/controllers/account/general_controller.ex", :no_return}
```
Teifion, line 165: Caused by the updates to the decache function. Both of these use the decache and I am not concerned with it's lack of return at this stage.

lib/teiserver/account/libs/user_lib.ex
```diff
@@ -103,6 +108,37 @@ defmodule Teiserver.Account.UserLib do
...
   end
 
+  @doc """
+  Attempts to get the user from the cache, failing that it will get it from the database.
+
+  Returns nil if no user found.
+  """
+  @spec get_user_by_id(User.id()) :: User.t() | nil
+  def get_user_by_id(user_id) do
+    user_id = int_parse!(user_id)
```
L-e-x-o-n, line 118: Why the `int_parse!`? I realise the idea is to be able to pass strings of integers but `User.id()` should always be a positive integer. I would either change the spec to allows for string or drop string support entirely.

Teifion: The sole reason for this is so we can accept strings (e.g. from web forms) and have a consistent key type. If we don't convert from string to integer we can get invalid caches and duplicate entries.

The alternative is to enforce the `user_id` be an integer and move the problem to the controller/liveview which I considered but would result in code duplication.

L-e-x-o-n: That fair, but I would change the spec then to show it can take a string.
```ex
@spec get_user_by_id(User.id() | String.t()) :: User.t() | nil
```

L-e-x-o-n (approved):

PR #1378, opened 2026-07-15
Use "and" not "or" when filtering players

PrivacyIsARight:


Teifion:
If you think you have found a bug and fixed the bug, it makes perfect sense to also update the tests to prove you have fixed the bug :)

Teifion:
I'll give it some consideration before jumping in but broadly I think this is a sensible fix

PR #1380, opened 2026-07-16
Phoenix to 1.8.9

Teifion:
We want to ensure we keep Phoenix up to date and as part of that we want to upgrade from 1.7 to 1.8.

The biggest obstacle is our continued use of views which appear to no longer be working. This means use of the `render` and `central_component` functions will also need to be removed, replaced with renders from the `HTML` module and converting into Phoenix components.


Teifion:
Closing as now so out of sync as to be pointless. A new PR will be made once more progress towards 1.8.9 is made.

PR #1381, opened 2026-07-16
Replaced all instances of central_component icons

Teifion:
We need to remove all uses of `central_component` for the upgrade to Phoenix 1.8. This removes the icon central component and updates the Fontawesome module slightly.

I used an LLM to help me get started initially, it produced a script which later become this:
```sh
#!/bin/bash

# Replace central_component("icon", ...) with {<Fontawesome.icon ... />}> 
find . -type f \( -name "*.eex" -o -name "*.heex" \) -exec sed -i \
  -E \
  -e 's|\{central_component\("icon",\s*icon:\s*(.+)\)\}|<Fontawesome.icon icon={\1} style="regular" />|g' \
  {} +

echo "Replacements completed!"
```


lib/teiserver/helpers/fontawesome.ex
```diff
@@ -1,46 +1,104 @@
...
+    `class`: css classes placed into the "class" attribute after the above classes
+    `css`: css placed into the "style" attribute
+  """
   @spec icon(map) :: Phoenix.LiveView.Rendered.t()
   def icon(assigns) do
-    style = fa_prefix(assigns[:style])
-    size = fa_prefix(assigns[:size])
-    weight = fa_prefix(assigns[:weight])
+    style =
+      if Application.get_env(:fontawesome_icons, :free_only, false) do
```
geekingfrog: hum what? I'm not super keen on having even more configuration and switches. Having a paid version of font-awesome also created more problem during the deployment because we needed to host the private copy somewhere for the prod server.
For something that is rather unimportant I would vastly prefer not to have to worry about this configuration.

Teifion: Sorry, I forgot we swapped to the free version. I copied the code directly from a different application I have which keeps that distinction.

I have updated the commit to no longer have that `free_only` logic.

lib/teiserver/helpers/fontawesome.ex
```diff
@@ -1,46 +1,90 @@
...
+  # Used to make the styling of icons limited to the free version
+  defp free_style("fa-brand"), do: "fa-brand"
+  defp free_style(""), do: ""
+  defp free_style(nil), do: nil
+  defp free_style(_any), do: "fa-solid"
+
+  # Used to make the styling of icons limited to the free version
+  defp free_weight(nil), do: nil
+  defp free_weight(_any), do: ""
+
```
geekingfrog: these should also be removed since we only have the free version

Teifion: The weight yes but the style should stay because it's swapping non-solid for solid.

geekingfrog (approved):

PR #1387, opened 2026-07-17
Fixes smurf links coming undone

Teifion:
The script_update_user, when not suppied with a smurf_of_id can set the smurf_of_id value to nil. This swaps smurf linking to a different changeset.


L-e-x-o-n:
Hmm, how does it set it to nil? I don't see why we need a new changeset just for this.

Teifion:
The only way I can see it updating is when a nil value is provided as part of the changeset (presumably from a cacheuser somehow) and that's causing this issue.

L-e-x-o-n (approved):

PR #1390, opened 2026-07-17
Added a catch for commands run without options

ChocoboCoding:
When commands are run without any options, the current version will error due to `data.option` being nil

### DIFF
--- a/lib/teiserver/bridge/discord_bridge_bot.ex
+++ b/lib/teiserver/bridge/discord_bridge_bot.ex
@@ -132,10 +132,16 @@ defmodule Teiserver.Bridge.DiscordBridgeBot do
 
   def handle_event({:INTERACTION_CREATE, %{data: data} = interaction, _ws}) do
     options_map =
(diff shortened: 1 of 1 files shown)


lib/teiserver/bridge/discord_bridge_bot.ex
```diff
@@ -132,10 +132,14 @@ defmodule Teiserver.Bridge.DiscordBridgeBot do
...
-        {name, value}
-      end)
+      if data.options == nil do
+        %{}
+      else
+        data.options
+        |> Map.new(fn %{name: name, value: value} ->
+          {name, value}
+        end)
+      end
```
Teifion:
```suggestion
      if data.options do
        data.options
        |> Map.new(fn %{name: name, value: value} ->
          {name, value}
        end)
      else
        %{}
      end
```

If you swap it around this way it reads a little nicer

Teifion (approved):

PR #1391, opened 2026-07-18
Switch to zero indexing in pagination

PrivacyIsARight:


lib/teiserver_web/components/pagination_components.ex
```diff
@@ -243,7 +243,7 @@ defmodule TeiserverWeb.PaginationComponents do
                           @base_url,
                           @current_params,
                           @include_params,
-                          %{"page" => @page + 1},
+                          %{"page" => @page},
```
Teifion: The next button is `@page + 2`, does that mean it's jumping ahead too much?

Teifion: Sorry, I'm confused. If I understand you correctly, `@page` is pointing at the previous page? If that's the case I'd much rather see you refactor this to remove the confusion in the first place.

Teifion: That makes sense, I'm okay with that.

Teifion (approved):

PR #1392, opened 2026-07-18
Fix ip endpoint check

PrivacyIsARight:


lib/teiserver/ip_check/api.ex
```diff
@@ -29,7 +29,7 @@ defmodule Teiserver.IpCheck.Api do
   defp get_api_endpoint do
     endpoint = Config.get_site_config_cache("teiserver.External IP check endpoint")
 
-    if endpoint != nil || endpoint != "",
+    if endpoint != nil && endpoint != "",
```
Teifion:
```suggestion
    if endpoint != nil and endpoint != "",
```

Looks like a mistake was made, `and` would be more appropriate here than `&&` as we are looking for the logical output. Good catch though.

Teifion (approved):

Issue #1394, opened 2026-07-18
Investigate crash 2026-07-18


geekingfrog:
Server just crashed all clients at 18:05 UTC

<img width="598" height="223" alt="Image" src="https://github.com/user-attachments/assets/1ebe04ea-688f-4240-96a2-824b032df5e2" />

The logs are flooded with

```
2026-07-18 18:04:46.975 [error] request_id=SpringTcpServer#607251 pid=<0.12043090.0>  GenServer #PID<0.12043090.0> terminating
** (stop) exited in: GenServer.call(#PID<0.4974.0>, {:lock, 611172, #Reference<0.2660042303.1073741827.191078>}, 5000)
    ** (EXIT) time out
    (elixir 1.19.4) lib/gen_server.ex:1142: GenServer.call/3
    (con_cache 1.1.1) lib/con_cache/lock.ex:22: ConCache.Lock.exec/4
    (teiserver 0.1.0) lib/teiserver/helpers/cache_helper.ex:17: Teiserver.Helpers.CacheHelper.cache_get_or_store/3
    (teiserver 0.1.0) lib/teiserver/tcp/spring/spring_tcp_server.ex:456: Teiserver.SpringTcpServer.handle_info/2
    (stdlib 6.2.2) gen_server.erl:2345: :gen_server.try_handle_info/3
    (stdlib 6.2.2) gen_server.erl:2433: :gen_server.handle_msg/6
    (stdlib 6.2.2) proc_lib.erl:329: :proc_lib.init_p_do_apply/3
Last message: %{reason: ":tcp_closed with socket", event: :disconnect, channel: "client_inout", userid: 611172}
```
and also

```
2026-07-18 18:04:49.571 [error] pid=<0.5604.0>  Ranch listener Teiserver.RawSpringTcpServer had connection process started with Teiserver.SpringTcpServer:start_link/4 at #PID<0.12742059.0> exit with reason: {:timeout, {GenServer, :call, [#PID<0.4974.0>, {:lock, 611172, #Reference<0.2660042303.843317264.28229>}, 5000]}}
```

it seems everything went back to normal without any manual intervention shortly after.

no idea what happened, but we should try to fix that if possible

bruno-dasilva:
I didn't notice before that the cache implementation is locking on writes, and a cache `put` or `get_or_store` (if a miss) both take a lock for the duration of their call.

[...]

PR #1395, opened 2026-07-19
YOLOing issue #1394

geekingfrog:
I suspect the deadlocks are caused by some inconsistent acquisition order on the caches, and some unfortunate timings. Although I couldn't produce a test case or an explanation for such deadlock, I believe removing this cache should help.

We have a cache that does `user.id => %CacheUser{}`, so a different cache `user.id => username` is redundant.


Teifion (commented):
I think this is a good approach, as soon as it's ready for review I'll look at it but I suspect this is an easy approve.

Teifion (approved):

PR #1400, opened 2026-07-20
Update name displayed of bridged messages

ChocoboCoding:
This PR updates the messaged bridged from discord to the game to use the discord server specific nick name, or the global display name instead of the account name

### DIFF
--- a/lib/teiserver/bridge/discord_bridge_bot.ex
+++ b/lib/teiserver/bridge/discord_bridge_bot.ex
@@ -574,7 +574,8 @@ defmodule Teiserver.Bridge.DiscordBridgeBot do
          author: author,
          content: content,
          channel_id: channel_id,
-         mentions: mentions
+         mentions: mentions,
+         member: %{nick: nick}
        }) do
     # Mentions come through encoded in a way we don't want to preserve, this substitutes them
     new_content =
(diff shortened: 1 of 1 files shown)


lib/teiserver/bridge/discord_bridge_bot.ex
```diff
@@ -606,11 +603,18 @@ defmodule Teiserver.Bridge.DiscordBridgeBot do
     #     end
     # end
 
+    name =
+      if nick == nil do
+        author.global_name
+      else
+        nick
+      end
```
L-e-x-o-n:
```suggestion
    name = nick || author.global_name
```
should work here and is cleaner.

L-e-x-o-n:
> Doesn't seem terrible, but couldn't this cause impersonation? If I set my nick to someone elses name and then send a message, then delete it?
> 
> Maybe the fix here is displaying the nick and the username together.

You can already change your Discord username so it is already possible (haven't seen it happen yet), while this does make it slightly easier IMO it shouldn't be something we should worry about. Impersonation is against server rules and such attempts should be dealt with by Discord moderators.

L-e-x-o-n (approved):

ChocoboCoding:
@L-e-x-o-n please check out the changes I did about impersonation (and making it more clear it is bridged)

lib/teiserver/bridge/discord_bridge_bot.ex
```diff
@@ -606,11 +603,13 @@ defmodule Teiserver.Bridge.DiscordBridgeBot do
     #     end
     # end
 
+    name = nick || author.global_name
+
     message =
```
L-e-x-o-n, line 609:
```diff
-"#{author.username}: #{row}"
+"[Bridge] #{name}: #{row}"
```

The message is already seen as sent from `DiscordBridgeBot` so adding `[Bridge]` seems unnecessary.

ChocoboCoding: Oh, nvm then, reverted the commit

PR #1402, opened 2026-07-20
Fix multiple Internal Server Errors

PrivacyIsARight:


lib/teiserver/logging/lib/logging_plug.ex
```diff
@@ -71,8 +71,8 @@ defmodule Teiserver.Logging.LoggingPlug do
 
     page_log =
       PageViewLog.changeset(%PageViewLog{}, %{
-        section: section,
-        path: Enum.join(path, "/"),
+        section: String.slice(section, 0, 255),
+        path: path |> Enum.join("/") |> String.slice(0, 255),
```
geekingfrog: I don't think this is correct to just truncate things. I think a better fix would be to only log page view that return something "normal", so 2xx or 3xx, and just ignore the 4xx/5xx.
There shouldn't be any valid path with more than 255 chars.

lib/teiserver/account/libs/user_lib.ex
```diff
@@ -116,7 +119,12 @@ defmodule Teiserver.Account.UserLib do
   @spec get_user_by_id(User.id() | String.t()) :: User.t() | nil
   def get_user_by_id(user_id) do
     user_id = int_parse!(user_id)
-    cache_get_or_store(:users_by_id, user_id, fn -> get_user(user_id) end)
+
+    if user_id >= @bigint_min and user_id <= @bigint_max do
+      cache_get_or_store(:users_by_id, user_id, fn -> get_user(user_id) end)
+    end
+  rescue
+    ArgumentError -> nil
```
geekingfrog: I don't like silently swallowing argument errors. If it blows up, at least it'll be very visible and we can fix it.

lib/teiserver/account/libs/user_lib.ex
```diff
@@ -116,7 +119,12 @@ defmodule Teiserver.Account.UserLib do
   @spec get_user_by_id(User.id() | String.t()) :: User.t() | nil
   def get_user_by_id(user_id) do
     user_id = int_parse!(user_id)
-    cache_get_or_store(:users_by_id, user_id, fn -> get_user(user_id) end)
+
+    if user_id >= @bigint_min and user_id <= @bigint_max do
```
geekingfrog: This is a fix only for this function, and would still fail on non numeric strings.
A better way is to create a function `parse_user_id(String.t() | User.id()) -> {:ok, integer()} | {:error, reason :: term()}`, and use this instead of `int_parse!`. This new function can do all the required validation.

lib/teiserver/logging/lib/logging_plug.ex
```diff
@@ -81,7 +79,7 @@ defmodule Teiserver.Logging.LoggingPlug do
       })
 
     if conn.assigns[:do_not_log] == nil do
-      Repo.insert!(page_log)
+      Repo.insert(page_log)
```
geekingfrog: why the change there? With your version, the result (potential error) is silently ignored, which is bad.

geekingfrog (commented):
Also one thing, none of this is tested. If you fix something, it's good to first have a test case reproducing the bug, have a corresponding failing test, and then fix that.

lib/teiserver/logging/lib/logging_plug.ex
geekingfrog: because the leading zeroes don't count, so 0002 == 2 which is valid.

geekingfrog:
if you want to fix the log view thingy, I think the correct place is at the router or controller level, to restrict upfront the length of each path param.

geekingfrog:
actually, I am not sure about logging the exact path every time. wouldn't it be better to log the path as defined in the router, without the substituted parameters? that would prevent this logging problem in the first place, and probably also improve what we're logging.
that is, instead of logging `foo/1234/bar` we log `foo/{userid}/bar`
Also, this may very well use grafana+vm instead of our own stuff. 

@Teifion thoughts on that?

Teifion:
It's potentially useful to understand the historical context and thus idea behind the web logging. It was a logging system I'd written for a completely different application and copied wholesale to Teiserver back when we had peaks of around 50 players. The intent was to know which pages were badly designed and to see which pages were actually visited.

tl,dr; it was designed for a different age (fewer players, I knew less) and does not fit this one

I think tracking `foo/{userid}/bar` is a much better way to do it and would make significantly more sense for our current situation. If we could move web-logging off of Teiserver completely I think at this stage that would be a good thing.

geekingfrog:
right. so for this PR I think it's better to just ignore the issue with the log view record persist to the dB, that can be addressed separately.

geekingfrog:
There are still no test for the fix about invalid account user in the path. This is not going in until there are some.

PR #1408, opened 2026-07-22
New "Overwatch Tech" Role

tunelessly:
Adds a role called "Overwatch Tech" under "Staff". Unsure whether this is the right place for it. Picked a color for the icon at random.


Teifion:
What is the purpose of the role? I can't see it doing anything not already covered by other roles.

Teifion:
At this stage we will not be adding the new role.

Issue #1412, opened 2026-07-23
Replace old user picker with new user picker


Teifion:
As part of moving away from the `central_component` setup we have several instances where the UserPicker needs to be used in the place of the `picker_script`. In each of these cases we need to alter the form to use the UserPicker but also the controller to use a form. In an ideal world each instance will be updated in general and re-thought.

PR #1417, opened 2026-07-23
Removes picker and picker_script components

Teifion:
Have created Issue #1412 to track replacement of pickers removed by this change that need to be replaced or in some way refactored.


Teifion:
As per coordination with [the moderation team](https://discord.com/channels/549281623154229250/1529807576311660574) they are happy for the removal of new-user-select pages for actions and bans; hence the removal of those pages and the tests. We will not be merging/deploying this until Monday the 27th to give them time to change their mind should they discover they do need them.

Teifion:
Moderation team have confirmed they do not use these pages in this manner so it is safe to remove them.

L-e-x-o-n (approved):

PR #1421, opened 2026-07-24
Add info to the development guide to make it a bit smoother for those using devcontainers

tunelessly:
When developing with devcontainers I had failing tests at first, regardless of whether I had made any changes or not. The cause was my lack of understanding on how to reconfigure the application for testing without having to stop and reconfigure the container by editing the compose file, an approach I find too awkward- vscode complains immediately and asks you to reload the window when you do it manually. 

This guide (and the application's default configuration) also doesn't cover the steps needed to run the application with containerized dependencies - database, mail server - which requires (to my very new to the project eyes at least) editing either some source and rebuilding or changing some environment variables. This is a commonplace thing to need to do for any project, of course; I just think that this guide should make it a little bit smoother to those completely new to teiserver.

To that end I wrote a small addition to the development guide that I believe is suitable to anyone going with the devcontainer option and doesn't already know how teiserver works more intimately than I and needn't necessarily understand a lot about containers.

Edits, suggestions, etc. all more than welcome.

### DIFF
--- a/documents/guides/development.md
+++ b/documents/guides/development.md
@@ -249,6 +249,30 @@ Eventually we want to move the balance algorithms to some external solver which 
 
 The balance alrogirhtms are located in [this](/lib/teiserver/battle/balance) directory.
 
+### Working under the devcontainer
+
+The commands under [Testing](#testing) and [Useful tools and commands](#useful-tools-and-commands) must be invoked in the container's environment with the environment's variables set accordingly. It's simpler to pass the right values for the variables to the environment compared to reconfiguring the container's environment by editing `compose.yml` and restarting the container. You can do this in the following manners:
+```bash
+docker exec -it teiserver bash
+```
+for an interactive shell; or
+```bash
+docker exec [-e VAR=value]... teiserver [command]
+```
+for single commands.
+
+For instance, to run tests interactively:
+```bash
+docker exec -it teiserver bash
+# We're now using bash in the container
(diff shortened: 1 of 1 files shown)


documents/guides/development.md
```diff
@@ -249,6 +249,30 @@ Eventually we want to move the balance algorithms to some external solver which
...
+### Working under the devcontainer
+
+The commands under [Testing](#testing) and [Useful tools and commands](#useful-tools-and-commands) must be invoked in the container's environment with the environment's variables set accordingly. It's simpler to pass the right values for the variables to the environment compared to reconfiguring the container's environment by editing `compose.yml` and restarting the container. You can do this in the following manners:
+```bash
+docker exec -it teiserver bash
+```
+for an interactive shell; or
+```bash
+docker exec [-e VAR=value]... teiserver [command]
+```
```
geekingfrog: I am dubious about that. This is supposed to be basic docker knowledge. If you are using devcontainers, you should be familiar with that.
Even more so, there are examples a bit further down with that kind of command, so I would just remove this block. Verbosity has a cost

tunelessly: Agreed. I debated whether to include that and decided it'd be easier to later delete than to add more. 

To be clear: delete only the highlighted lines or the whole 'generic command' section (lines 255-262)?

geekingfrog: you have the command to spawn bash in the container again just below, so I think you can delete the one line 256.

geekingfrog (approved):

PR #1422, opened 2026-07-24
Use ceil(log2(n+1)) for bit length, and update tests

PrivacyIsARight:


geekingfrog:
Please, if you use any AI tool, do disclose their use and how, we have an AI policy.

This looks bogus, if you want to represent 4 in base 2 with only 2 bits, you simply cannot. I don't think your "fix" is valid, maybe it would be better to actually raise an error.

And also, I am not going to accept this until you can explain what is this function used for, and where is the relevant protocol specification where this is used. Code doesn't exist in a vacuum.

Teifion (changes requested):
Plus one to what geekingfrog has sent, this is a PR I'd want some explanation of and demonstration of the side effects of such a change.

geekingfrog:
> Nowhere in my description did I ever say that you can represent 4 in base 2 with 2 bits. That's literally what this pr is fixing.

`{"4", 2, [1, 0, 0]},`  This test is exactly that: represent the number 4 as a bitstring of length 2.


> It's also important to note that the existing tests already had this exact kind of behavior.
> "{"12", 3, [1, 1, 0, 0]}" **This is literally exactly what you're concerned about in this pr, as existing behavior before this pr.**

Yep, that's bogus as well, and shouldn't be there.

> If you want a demonstration, you can easily just run my test file without the changes applied and see that the test fails.

I did mind you.

and finally:

> I'm getting incredibly tired of these AI accusations

It's not a fucking accusation, you review AI stuff differently. There has been already plenty of AI generated/assisted code merged in this repo, we do not ban all of them. If you can't comply with simple disclosure or react badly when being ask, perhaps you should consider that you look more like a cultist than a dev.

geekingfrog:
Ah also, the "I want you to explain what's that for": it's a test to see if there is someone at the wheel, I do not want to interact with an llm via github comments. Mindless code changes are just not good, it helps to understand the context.
This is also how you get more familiarity on the overall project, and grow the confidence and ability to push more important and meaningful changes.

Teifion:
There is _much_ to unpack and address here so lets start with a quick summary:

Yes, this is what I want contributors doing. I want contributors to call out sloppy, badly thought out or in any way problematic pull requests. I include my own in this. geekingfrog has done stellar contributions in the code they have written and equally excellent contributions in the items they have planned/reviewed. One day will be the last day they review code I work on and that will be a day the level of quality likely diminishes.

So, lets cut to the chase; LLM usage. Your PRs in this repo and the conversations I've seen; all point to using and LLM to a varying capacity. The recently closed [Fix site config PR](https://github.com/beyond-all-reason/teiserver/pull/1423) was a good example of the PR being very narrowly correct (there was an incorrect label on the config) but contextually incorrect (the config isn't used anywhere and isn't planned to be, it should be removed). I wanted to give you an opportunity to notice this so specifically asked you "Have you looked at **where this is used** and confirmed it is indeed a mode?" (bold added here).

So when geekingfrog challenges you on LLM usage and you overreact massively I am strongly of the belief you are using an LLM without declaring it. As geekingfrog has said, we are not anti-LLM, the fact we have both used them for coding means we know where they are strong and where they are weak. We require their disclosure _because we have used them_, we know their code needs a different style of review.

Now lets focus specifically on this pull request and why I am closing it.

The pull request itself is again a narrow and limited change. Geekingfrog correctly requested you provide contextual information:

> And also, I am not going to accept this until you can explain what is this function used for, and where is the relevant protocol specification where this is used. Code doesn't exist in a vacuum.

As with me you have avoided his question completely. Instead you have become defensive over (in my opinion) completely justified concerns regarding undisclosed LLM usage. Replying that he is able to run the code himself is the opposite of the request he made.

If you want to continue to contribute to this repository you will need to alter you attitude towards the code but most importantly to the maintainers. I do not care how good the code you produce is if you are unable to be respectful to other contributors or to actually respond when they ask you a question.

PR #1423, opened 2026-07-24
Fix site config account registration mode text

PrivacyIsARight:


Teifion (changes requested):
Have you looked at where this is used and confirmed it is indeed a mode?

Teifion:
You did not answer my actual question, if you had you'd have seen this is actually a dead config and not used anywhere. The correct PR will be to remove it completely. I'm closing this PR for now and will post a fuller reply to your other currently open PR regarding LLM usage and the like.

PR #1427, opened 2026-07-26
list allowed origins

geekingfrog:
When we switched off TLS handling for teiserver in https://github.com/beyond-all-reason/ansible-teiserver/pull/29 it broke websockets (liveviews).

This is because the `check_origin: :conn` checks the `scheme`, and in this case the browser would send `https`, but the conn would have `http`. I didn't find a way to check host+port match between the origin header and the connection.
The fix of listing all allowed domain is a bit meh since it's one more thing to support when moving/adding domains and we're going to forget that. But at the same time, it's also simple and shouldn't happen too often.

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -68,7 +68,7 @@ config :teiserver, Teiserver.Repo,
 
 check_origin =
   if Teiserver.ConfigHelpers.get_env("TEI_SHOULD_CHECK_ORIGIN", false, :bool) do
-    :conn
+    ["//#{domain_name}"]
   else
     false
   end
--- a/config/config.exs
+++ b/config/config.exs
@@ -34,7 +34,6 @@ config :teiserver, TeiserverWeb.Endpoint,
   pubsub_server: Teiserver.PubSub,
   debug_errors: Config.config_env() == :dev,
   code_reloader: Config.config_env() == :dev,
-  check_origin: Config.config_env() == :prod,
   version: Mix.Project.config()[:version],
   https: [
     versions: [:"tlsv1.2"],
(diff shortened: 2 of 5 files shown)


bruno-dasilva:
@geekingfrog would just setting check_origin to `true` work? Then it dynamically uses whatever was already set in `config(:url)[:host]`? I double checked [the source](https://github.com/phoenixframework/phoenix/blob/v1.7.24/lib/phoenix/socket/transport.ex#L595-L596) and it indeed only cares about hostname matching if you do this.

https://phoenix.hexdocs.pm/1.7.24/Phoenix.Endpoint.html#socket/3-common-configuration:~:text=If%20true%2C%20the%20header%20is%20checked%20against%20%3Ahost%20in%20YourAppWeb%2EEndpoint%2Econfig%28%3Aurl%29%5B%3Ahost%5D%2E

bruno-dasilva:
alternatively if you prefer the list, I don't see why you couldn't have the list pull the `config.domain_name` rather than hardcoding

geekingfrog:
Take 2, based on the hostname. It was like that [a long time ago](https://github.com/beyond-all-reason/teiserver/pull/913). But it generated some very noisy log, so I also added a log filter to remove these.

config/runtime.exs
```diff
@@ -67,8 +67,8 @@ config :teiserver, Teiserver.Repo,
     Teiserver.ConfigHelpers.get_env("TEI_DB_QUEUE_INTERVAL", repo_env[:queue_interval], :int)
 
 check_origin =
-  if Teiserver.ConfigHelpers.get_env("TEI_SHOULD_CHECK_ORIGIN", false, :bool) do
-    :conn
+  if true || Teiserver.ConfigHelpers.get_env("TEI_SHOULD_CHECK_ORIGIN", false, :bool) do
```
L-e-x-o-n: Always true?

geekingfrog: oops, forgot to take that off after testing, good catch

Teifion (approved):

PR #1451, opened 2026-08-01
$modme and $unmodme commands

geekingfrog:
second attempt after https://github.com/beyond-all-reason/teiserver/pull/1307

`$unmodme` drops all moderator privilege for the user, `$modme` to regain them. This only has effect in the client itself, for the duration of the connection.

The main goal is so that you don't `!resign` the entire team without any vote by mistake, or force a map change when you just wanted a vote.

### DIFF
--- a/lib/teiserver/coordinator/coordinator_commands.ex
+++ b/lib/teiserver/coordinator/coordinator_commands.ex
@@ -18,9 +18,9 @@ defmodule Teiserver.Coordinator.CoordinatorCommands do
   alias Teiserver.Moderation
 
   @splitter "---------------------------"
-  @always_allow ~w(help whoami whois discord coc mute unmute ignore unignore website party)
+  @always_allow ~w(help whoami whois discord coc mute unmute ignore unignore website party modme)
   # These commands are handled by coordinator commands, but are not on the always allow list
-  @mod_allow ~w(modparty unparty)
+  @mod_allow ~w(modparty unparty unmodme)
   @forward_to_consul ~w(s status players follow joinq leaveq splitlobby y yes n no explain)
   @admin_commands ~w(broadcast)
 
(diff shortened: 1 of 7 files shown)


lib/teiserver/coordinator/coordinator_commands.ex
```diff
@@ -49,7 +49,7 @@ defmodule Teiserver.Coordinator.CoordinatorCommands do
         true
 
       # Allow all except Admin only commands for moderators
-      Auth.moderator?(senderid) and not Enum.member?(@admin_commands, cmd.command) ->
+      client.moderator and not Enum.member?(@admin_commands, cmd.command) ->
```
L-e-x-o-n: Why change this? Coordinator commands aren't ran accidently, they should be safe to keep.
More importantly, there is the event organiser role which gives SPADS moderator (client mlderator bit) to non swrver moderators, this would give them access to commands they shouldnt have.

Teifion: They shouldn't be run accidentally but moderators should typically try to remember to `!cv` at the start of their commands so I'm in favour of having moderator commands be inaccessible to non-modded users.

The event organiser is a different matter though and does need to be considered.

L-e-x-o-n: These commands don't have !cv, these are server $ commands.

geekingfrog: makes sense, will remove that. the most important (for me) are the spads commands

lib/teiserver/coordinator/coordinator_commands.ex
```diff
@@ -256,8 +256,10 @@ defmodule Teiserver.Coordinator.CoordinatorCommands do
           @splitter
         ]
 
+        sender_client = Account.get_client_by_id(sender)
+
         mod_parts =
-          if Auth.admin?(sender) or Auth.moderator?(sender) do
```
L-e-x-o-n, line 260: I would revert this as well for the same reason.

L-e-x-o-n (approved):

PR #1452, opened 2026-08-01
Accept bearer tokens in AuthPlug and answer json auth errors with json

WybrenKoelmans:
AI disclaimer: Claude Opus 5 used for transform psuedo-code into proper Elixir, tests fully generated.

### DIFF
--- a/lib/teiserver/account/error_handler.ex
+++ b/lib/teiserver/account/error_handler.ex
@@ -11,29 +11,52 @@ defmodule Teiserver.Account.ErrorHandler do
   @impl Guardian.Plug.ErrorHandler
 
   def auth_error(conn, {:unauthenticated, _reason}, _opts) do
-    redirect_to =
-      if conn.query_string != nil && conn.query_string != "" do
-        "#{conn.request_path}?#{conn.query_string}"
-      else
-        "#{conn.request_path}"
-      end
+    if json_request?(conn) do
+      unauthorized_json(conn, "unauthenticated")
+    else
+      redirect_to =
+        if conn.query_string != nil && conn.query_string != "" do
+          "#{conn.request_path}?#{conn.query_string}"
+        else
+          "#{conn.request_path}"
+        end
 
-    conn
-    |> put_resp_cookie("_redirect_to", redirect_to, sign: true, max_age: 60 * 5)
-    |> Controller.redirect(to: ~p"/login")
+      conn
+      |> put_resp_cookie("_redirect_to", redirect_to, sign: true, max_age: 60 * 5)
+      |> Controller.redirect(to: ~p"/login")
+    end
   end
 
   def auth_error(conn, {:invalid_token, _message}, _opts) do
-    conn
-    |> put_resp_cookie("_teiserver_key", "", max_age: 0)
-    |> Controller.redirect(to: ~p"/login")
+    if json_request?(conn) do
+      unauthorized_json(conn, "invalid_token")
+    else
+      conn
+      |> put_resp_cookie("_teiserver_key", "", max_age: 0)
+      |> Controller.redirect(to: ~p"/login")
+    end
   end
 
   def auth_error(conn, {type, _reason}, _opts) do
-    body = to_string(type)
+    if json_request?(conn) do
+      unauthorized_json(conn, to_string(type))
+    else
+      body = to_string(type)
 
+      conn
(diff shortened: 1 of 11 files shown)


WybrenKoelmans:
relates to https://github.com/WybrenKoelmans/teiserver/pull/3

Teifion:
This is a potentially very significant alteration (it affects authentication), sadly I am away next week so won't be able to review it but I'd like both GeekingFrog and Lexon to feel comfortable with this before it gets merged.

L-e-x-o-n:
I would like some context on this, I can see from the linked PR that it's related to Widget Hub.
What is the end goal here and what role does teiserver have in this?

Has an agreement been reached that this is the way to go forward regarding Widget Hub or is this just a proof of concept?

WybrenKoelmans:
Yeah so the JSON error response for requests that set `Accept: application/json` is just a nice DX improvement, I don't think we need to discuss that much.

The other part of this PR: the bearer token is to authenticate the request is needed for non-web (sessionless) clients trying to communicate with Teiserver.

The linked PR provides an endpoint for a Teiserver authenticated user to request a signed Firebase token. Now because in 'new-client' the request for this must be done using the backend process (the node part, not the 'renderer'/web part) of the Electron application, the request will not have a cookie to authenticate with, but instead will use the oauth token which the user gets when logging in to the new-client to communicate with Teiserver and authenticate using the Bearer header, that's currently not possible without this PR.

Now regardless of how the whole WidgetHub next phase will go. I think this PR stands on it's own as something that is a good addition to have. 

I tried to make sure it is _safe_ making sure to generate the test you see accompanied, but I will also rely on you guys experience to help me make this PR really good, so please let me know what you'd like to see different if anything.

Let's have the discussion for the WidgetHub PR, when this dependent PR is merged (or you can already start on that linked PR if you want ofcourse).

lib/teiserver/account/error_handler.ex
```diff
@@ -11,29 +11,52 @@ defmodule Teiserver.Account.ErrorHandler do
...
+
+      conn
+      |> put_resp_content_type("text/plain")
+      |> send_resp(401, body)
+    end
+  end
+
+  # :accepts sets phoenix_format, so every json pipeline is covered without
+  # hardcoding route prefixes here.
+  defp json_request?(conn), do: conn.private[:phoenix_format] == "json"
```
geekingfrog: looking at private field is a big red flag, I'm sure there is another way to render a response according to the `Accept` header.

geekingfrog (commented):
I have serious doubt about that change. It merges the auth logic from web (with cookies for sessions) and bearer tokens meant for API.
Generally, web and API are served by different endpoint and different pipeline, and I'm not keen on merging them, it increases the complexity of that logic for no real benefit.

The only thing that seems legit is the change around `current_resource` that falls back to `resource_from_token`. Haven't really looked into that, but that seems different from the rest, and potentially more useful.

WybrenKoelmans:
yeah okay indeed I've split them now, am I heading in the right direction now?

lib/teiserver/account/plugs/api_auth_plug.ex
geekingfrog, line 1: This looks **a lot** like `TeiserverWeb.Plugs.OAuthAuthenticatedPlug`, why do you need a different mechanism there?

lib/teiserver/account/plugs/api_auth_plug.ex
```diff
@@ -0,0 +1,80 @@
...
+
+  @behaviour Plug
+
+  def init(opts), do: opts
+
+  def call(conn, _opts) do
+    with {:ok, raw_token} <- bearer_token(conn),
+         {:ok, user} <- user_from_token(raw_token),
+         :ok <- login_allowed(user) do
+      Logger.metadata([user_id: user.id] ++ Logger.metadata())
```
geekingfrog, line 35: as per [doc](https://logger.hexdocs.pm/main/Logger.html#metadata/1):

> This function will merge the given enumerable into the existing metadata, with the exception of setting a key to nil, which will remove that key from the metadata.

geekingfrog (commented):
haven't looked in details at the tests, I'm still not sure about the whole guardian module for bearer token.

WybrenKoelmans:
This functionality is no longer required for my work, other developer may freely pick this up if they require it!

PR #1453, opened 2026-08-01
devcontainer misc fixes

WybrenKoelmans:


.devcontainer/devcontainer.json
```diff
@@ -9,6 +9,10 @@
 		"../compose.yaml"
 	],
 
+	"remoteEnv": {
+		"EDITOR": "code --wait"
+	},
+
```
L-e-x-o-n: What is this for? What if someone doesn't have VS code?

Teifion: I'd love to know the story of how that got there too

WybrenKoelmans: So it's about applications like git having editor integrations, for example when you do `git commit` (note the lack of `-m "my commit"` it will try to start your "editor", many distro's will have some type of editor preinstalled like `nano` `vim` or god forbid `emacs` (😉) and open it to wait for you to enter your commit message there. This devcontainer does not have the `EDITOR` env set by default, so `git commit` will fail. 

With this set to `code --wait`, if you ran `git commit` in the vscode integrated terminal, it will just open a new code tab as your editor, and as you close that tab, it will use it as the commit message (actually vscode will just detect that you are in the middle of a commit and accomate it, but it's beyond the point). 

Now if you are somebody that doesn't use vscode for this devcontainer, you can just set the EDITOR env to something else, like `nano` using .bashrc. 

Maybe Jetbrains also has some type of integration for this, no clue, but basically without it you were getting an error, and after this PR you'd just get a different error if you're not using vscode.

Teifion: > Now if you are somebody that doesn't use vscode for this devcontainer, you can just set the EDITOR env to something else, like nano using .bashrc.

So by default we are expecting people to use vscode? I'm all for supporting editors but making it so you have to change files if you are using a different editor defeats the purpose of devcontainers in my opinion.

WybrenKoelmans: [Yes](https://github.com/beyond-all-reason/teiserver/blob/main/documents/guides/local_setup.md#using-the-devcontainer), and that's actually a good thing. These devcontainers provide a frictionless path, if you are willing to follow the opinionated setup. Setting `EDITOR="code --wait"` reduces friction from trying to use `git commit` that would normally error.

The same goes for assume that the user will use docker and not podman, or work with bash and not zsh, etc. We can not support every combination of tools, but we should strive to support one opinionated frictionless setup.

Teifion: Please talk me through exactly what is going to happen if they have got VSCode installed but are using a different editor and what happens if VSCode isn't installed at all.

WybrenKoelmans: you'll just get `bash: code --wait: command not found` in both cases, as the integrated terminal/vscode devcontainer plugin doesn't shim `code` with a remote container vscode server.

but anyway, i made the  config vscode specific

geekingfrog: devcontainers support multiple editors. There is a per-editor section in the `customizations` field. If someone wants to make them work with a different IDE, I guess they'll add that bit of config there and then.
Looks fine to me tbh

geekingfrog (approved):

L-e-x-o-n (approved):

PR #1459, opened 2026-08-02
Party event sourcing

geekingfrog:
Start refactoring the party state machine with event sourcing. Very similar to what's used for the lobby.
The advantage of event sourcing is that it's easy to replicate events to other nodes.

### DIFF
--- a/lib/teiserver/party/events/leave.ex
+++ b/lib/teiserver/party/events/leave.ex
@@ -1,0 +1,39 @@
+defmodule Teiserver.Party.Events.Leave do
+  @moduledoc """
+  leave the party
+  """
+  alias Teiserver.Account.User
+
+  @enforce_keys [:leaver_id]
+  defstruct [:leaver_id]
+
+  @type t() :: %__MODULE__{
+          leaver_id: User.id()
+        }
+end
+
+defimpl Teiserver.Party.Event, for: Teiserver.Party.Events.Leave do
+  alias Teiserver.Helpers.MonitorCollection, as: MC
+  alias Teiserver.Party.Events.Leave
+  alias Teiserver.Party.Types.Aggregate
+  import Access, only: [key!: 1]
+
+  def apply_event(%Leave{} = ev, %Aggregate{} = agg) do
+    case Map.pop(agg.data.members, ev.leaver_id) do
+      {nil, _members} ->
+        agg
+
+      {_member, rest} when map_size(rest) == 0 ->
+        put_in(agg.data.members, %{})
+
+      {_member, new_members} ->
+        new_aggregate =
+          agg
+          |> Aggregate.put_in_data([key!(:members)], new_members)
+          |> Aggregate.update_in_data([key!(:monitors)], &MC.demonitor_by_val(&1, ev.leaver_id))
+          |> Aggregate.add_side_effect(:notify_updated)
+
+        new_aggregate
+    end
+  end
+end
--- a/lib/teiserver/party/event.ex
+++ b/lib/teiserver/party/event.ex
@@ -1,0 +1,13 @@
+defprotocol Teiserver.Party.Event do
+  alias Teiserver.Party.Types, as: PT
+
+  @doc """
+  Given an event and an aggregate, returns the new aggregate.
+
+  This function must be side-effect free, because many events may be folded
+  together or rolled back before arriving to the final aggregate to be used
+  by the lobby.
+  """
+  @spec apply_event(term(), PT.Aggregate.t()) :: PT.Aggregate.t()
+  def apply_event(event, data)
+end
--- a/lib/teiserver/party/server.ex
+++ b/lib/teiserver/party/server.ex
@@ -253,7 +253,8 @@ defmodule Teiserver.Party.Server do
         {:keep_state, data, [{:reply, from, {:error, :not_a_member}}]}
 
       {_member, rest} when map_size(rest) == 0 ->
-        {:stop_and_reply, :normal, [{:reply, from, :ok}], %{data | members: %{}} |> bump()}
+        new_data = %{data | members: %{}} |> bump()
+        {:keep_state, new_data, [{:reply, from, :ok}, {:next_event, :internal, :empty}]}
 
       {_member, new_members} ->
         new_data =
@@ -597,6 +598,10 @@ defmodule Teiserver.Party.Server do
 
   def handle_event(:info, _event, :starting_up, %PT.Data{} = data) do
     {:keep_state, data, [:postpone]}
+  end
+
+  def handle_event(:internal, :empty, _state, %PT.Data{} = data) do
+    {:stop, {:shutdown, :empty}, data}
   end
 
   def handle_event(:state_timeout, :snapshot_timeout, :starting_up, %PT.Data{} = data) do
--- a/lib/teiserver/party/server.ex
+++ b/lib/teiserver/party/server.ex
@@ -11,6 +11,8 @@ defmodule Teiserver.Party.Server do
   alias Teiserver.Matchmaking
   alias Teiserver.Messaging
   alias Teiserver.Party
+  alias Teiserver.Party.Event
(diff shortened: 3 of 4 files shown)


lib/teiserver/party/events/leave.ex
```diff
@@ -0,0 +1,43 @@
...
+        new_aggregate =
+          agg
+          |> put_in([Access.key!(:data), Access.key!(:members)], new_members)
+          |> update_in(
+            [Access.key!(:data), Access.key!(:monitors)],
+            &MC.demonitor_by_val(&1, ev.leaver_id)
+          )
+          |> update_in([Access.key!(:side_effects)], fn side_effects ->
+            side_effects ++ [:notify_updated]
+          end)
```
Teifion: I'm finding this challenging to read and parse. I can understand each step but I'd much rather see it moved into a function within `Teiserver.Party.Types.Aggregate` with an appropriate comment explaining what is happening for later comprehension.

geekingfrog: I've amended the commit but I'm not convinced by the changes.
I think the helpers `add_side_effect` and `add_action` are good, but the thin wrapper around `put_in` and `update_in`, I'm less convinced.
Importing `key!/1` makes the code a bit less verbose as well.
What do you think?

Teifion: The `key!/1` does make it less verbose and helped.

I think a better way to put it is, this is _very clever_ code. It does a lot in very few lines, here's my understanding of it. I do not have a high confidence in my understanding of it.

Line 1 - `put_data_in`
- Create a function to pull the key `:members`
- Give that and `new_members` to `Aggregate.put_in_data`
- `Aggregate` will put them into the `agg` data structure as per the comment in `Aggregate`

Line 2 - `update_data_in`
- Same principle as the previous line but for the `:monitors` in the data we are updating
- In updating the data it also dereferences the monitor of the person leaving the party
- The code reads at first line it's updating the data but it looks more like it should be removing keys

Line 3 - `add_side_effect`
- This part (I think) I understand better, it's adding a side effect to be actioned/executed after all data inputs have been processed

How did I do?

If I understood it then that's better than I was expecting, if I haven't then the code is simply too clever for me to maintain (I didn't become the first server dev because I was smart...) and I'd like to work to make it easier to understand.

geekingfrog: Reading that, I don't think `put_data_in` and `update_data_in` are worth it since it creates more confusion than anything.
Ultimately I'm not sure how to modify that, the original version modifies a few bits in the aggregate, it's "just" data manipulation. The whole `key!` wrapper is unfortunately needed because `put_in` and the likes can't work with bare keywords on structs, unlike with maps.

for reference, here what it looked like before the helpers.
```elixir
new_aggregate =
  agg
  |> put_in([Access.key!(:data), Access.key!(:members)], new_members)
  |> update_in(
    [Access.key!(:data), Access.key!(:monitors)],
    &MC.demonitor_by_val(&1, ev.leaver_id)
  )
  |> update_in([Access.key!(:side_effects)], fn side_effects ->
    side_effects ++ [:notify_updated]
  end)
```

which could become

```elixir
new_aggregate =
  agg
  |> put_in([key!(:data), key!(:members)], new_members)
  |> update_in([key!(:data), key!(:monitors)], &MC.demonitor_by_val(&1, ev.leaver_id))
  |> Aggregate.add_side_effect(:notify_updated)
```

As for the monitors

> The code reads at first line it's updating the data but it looks more like it should be removing keys

In the internals of [MonitorCollection](https://github.com/geekingfrog/teiserver/blob/9e84bffff9138449e8e0ba9ad70af9b14b74b454/lib/teiserver/helpers/monitor_collection.ex) it's doing that (alongside `Process.demonitor`). The monitor collection thingy is used quite a bit in several modules. It's a way to monitor something and attach a value to it, so that it's easy to demonitor given the value (like a user leaving), or modify some state based on the ref when the `:DOWN` message arrives (like when a user session crashes/disappear).

Teifion: Had a chat about some of this with Lexon and it's making a bit more sense but the fact it wasn't immediately making sense means (to me) it should have a comment.

A lot of the confusion I had was why we have nested layers of things like `put_in` and `Access`. The first suggestion was it should all live in the `Aggregate`, something like this.

```diff
- |> put_in([key!(:data), key!(:members)], new_members)
- |> update_in([key!(:data), key!(:monitors)], &MC.demonitor_by_val(&1, ev.leaver_id))
- |> update_in([Access.key!(:side_effects)], fn side_effects ->
-   side_effects ++ [:notify_updated]
- end)
+ |> Aggregate.put_in([:data, :members], new_members)
+ |> Aggregate.update_in([:data, :monitors], &MC.demonitor_by_val(&1, ev.leaver_id))
+ |> Aggregate.update_in([:side_effects], fn side_effects ->
+   side_effects ++ [:notify_updated]
+ end)
```

Once that's done it's much easier for me to read and comprehend the code (though I am always a fan of more comments around things like this for when I'm debugging at 10pm on a weekend because the server fell over from stuff I wrote 6+ months ago :D).

geekingfrog: There is a problem with `Aggregate.put_in` though, because while the aggregate is a struct, there is no guarantee that what lives under `:data` is also a struct. So a helper cannot magically wrap every key into `Access.key!` since it's going to break as soon as you try to touch a bare map.

Is it a problem of familiarity with the pattern? How would you update several fields in a struct otherwise?

Teifion: I think part of my issue is familiarity with the pattern but I also think in this case, when we call `Aggregate.put_in` we shouldn't be needing to represent the underlying data pattern of the Aggregate in the caller, the `Aggregate` module should be dealing with that for us.

geekingfrog: However, the concept of aggregate is separate from domain logic. We should be able to have one aggregate definition generic over the data. For example:
`Aggregate{data: t()}`

Then, let's say you have `data: %{members: %{string() => %{id: string(), last_seen: timestamp()}}` in the aggregate.

How would you:
1. add a new member?
2. change the `last_seen` field of an existing member?

Teifion: The way I'd expect it to be written for these two is something like this:

Adding a new member
```ex
Aggregate.put_new_in(:members, key, data)
```

Updating the `last_seen`
```ex
data = Aggregate.get(:members, key)
new_data = %{data | last_seen: new_timestamp}
Aggregate.update_in(:members, key, new_data)
```

`Aggregate` is a data structure and in the first case our intent is to put a new item in so we'd say `put_new_in` but the second we want to update it, so we need to extract (`get`), change and update the data.

geekingfrog: How would you write `Aggregate.put_new_in`?  Also, not the biggest fan of the constant var assignment for update. It gets really messy when you have deeper nesting (just one more nesting for example would be already gnarly)

Teifion: Regarding var assignment I'd probably look to find a way it can be done with pipes, I wrote it this way to make it very explicit what each step was for the purpose of the discussion.

Regarding how to write `put_new_in`, something like:
```ex
def put_new_in(%Aggregate{} = agg, field, key, value) do
  new_contents = Map.put_new(agg.field, key, value)
  %Aggregate{agg | field => new_contents}
end
```

geekingfrog: It all falls apart when you deal with more nesting though. That's why there are `put_in` and `update_in` in the standard library, because modifying nested data structure in an immutable world is a pain without these helpers.

I could go with a `Aggregate.update_data(aggregate, mutation_fn)` where `mutation_fn: data -> data`, but I think anything else should be left out the aggregate module.

Teifion (approved):
I'm still not a fan of the put_in stuff but I think it's a hill not worth dying on and perfect is the enemy of good enough.

Issue #1239, opened 2026-06-05
core_components.ex: Checkbox input submits wrong value due to HEEx boolean attribute rendering


BElluu:
The `input/1` component's checkbox clause in `core_components.ex` uses `value={@value}`, where `@value` is typically a boolean derived from an Ecto schema field. HEEx treats boolean attribute values specially:

- `vaule={false}` -> attribute is omitted entirely -> browser submits "on" (HTML default)
- `value={true}` -> bare attributy (no value) -> browser submits "" (empty string)

Neither "on" nor "" passes Ecto's boolean cast, which expects "true". This means any checkbox driven by a boolean Ecto field cannot be submitted correctly through the form - the failed stays false regardless of the checked state.

**Steps to reproduce:**
Add a checkbox backed by a boolean Ecto field (e.g. case_sensitive :boolean), submit the form with the box checked - the field is always false in the changeset.

**Affected code**

```
  # lib/teiserver_web/components/core_components.ex
  def input(%{type: "checkbox", value: value} = assigns) do
    ~H"""
    <input
      type="checkbox"
      value={@value}    ← bug: boolean passed as HEEx attribute
      checked={@checked}
    />
    """
  end
```

**Proposed fix**

```
  <input
    type="checkbox"
    value="true"       ← hardcoded string, matches Phoenix generator default
    checked={@checked}
    {@rest}            ← consistency: other input clauses already pass @rest
  />
```
This matches what phx.new generates in its own core_components.ex template.

Teifion:
I am encountering this issue with my work on replacing central_components checkbox. I will attempt to handle both at once (this is a blocker for my item).

Teifion:
Fixed in https://github.com/beyond-all-reason/teiserver/pull/1460

PR #1460, opened 2026-08-02
Replace central_component checkbox and fix default checkbox value

Teifion:
Along with the central_component removal it also addresses issue https://github.com/beyond-all-reason/teiserver/issues/1239


Teifion:
This will need to be tested on integration and the forms behaviour verified by those using them (thinking moderation in particular). I have tested what I can and it works but there are a lot of combinations and it is possible I have missed one or that the local/fake data was not sufficient to bring about a problem.

L-e-x-o-n (approved):

PR #1471, opened 2026-08-08
feat(misc):  add linter task and small fixes

WybrenKoelmans:


mix.exs
```diff
@@ -201,11 +202,24 @@ defmodule Teiserver.MixProject do
         "sass light --no-source-map --style=compressed",
         "phx.digest"
       ],
+      # Everything the "Lint and style checks" github action runs, in the same
+      # order, so a failure here is a failure there. Keep in step with
+      # .github/workflows/lint.yml
+      lint: [
```
geekingfrog, line 208: does that work? I do `mix lint` and get `(Mix) The task "lint" could not be found`

WybrenKoelmans: Yes definitely, i'm not sure why it wouldn't work for you though. Do you need to reload something for `mix` when running outside of a devcontainer?

```
:/workspaces/teiserver# mix lint
No retired packages found
No vulnerabilities found.
Checking 1084 source files (this might take a while) ...

Please report incorrect results: https://github.com/rrrene/credo/issues

Analysis took 2.7 seconds (0.1s to load, 2.6s running 88 checks on 1084 files)
9308 mods/funs, found no issues.

Use `mix credo explain` to explain issues, `mix credo --help` for options.
```

mix.exs
```diff
@@ -201,11 +202,24 @@ defmodule Teiserver.MixProject do
...
+        "compile --warnings-as-errors",
+        "credo --strict"
+      ],
       precommit: [
         "deps.unlock --check-unused",
-        "compile --force --warning-as-errors",
+        "compile --force --warnings-as-errors",
         "format",
-        "credo",
+        "credo --strict",
```
Teifion: We already have credo set to be strict in the config, this shouldn't be needed.

mix.exs
Teifion: `mix lint` ran fine for me after I checked out the PR. I'm not using a dev container.

.vscode/tasks.json
```diff
@@ -46,6 +46,19 @@
                 "reveal": "always",
                 "panel": "dedicated"
             }
+        },
+        {
+            // Same checks as the "Lint and style checks" github action. MIX_ENV
+            // is set explicitly because the container pins it to dev, which
+            // takes precedence over the preferred_envs in mix.exs.
```
Teifion, line 53: This is a `.json` file and the comments are invalid. If you want comments you will need to use a `.jsonc` file.

.devcontainer/startup.sh
```diff
@@ -27,6 +27,12 @@ cat > /etc/motd << 'MOTD'
     Email:    root@localhost
     Password: password
 
+  Before pushing, run the lint and style checks CI does:
```
Teifion:
```suggestion
  Before pushing, run the same lint and style checks CI does:
```

mix.exs
```diff
@@ -201,11 +202,24 @@ defmodule Teiserver.MixProject do
...
+      # .github/workflows/lint.yml
+      lint: [
+        "deps.unlock --check-unused",
+        "hex.audit",
+        # decimal and gun are transitive dependencies that are a pain to
+        # update, see the workflow for the details
+        "deps.audit --ignore-package-names decimal,gun",
+        "format --check-formatted",
+        "compile --warnings-as-errors",
+        "credo --strict"
```
Teifion: Same as below regarding the strict flag

Teifion (commented):
My main question is why add this rather than tweaking/adding to the `mix precommit` task?

My second question is how to ensure they stay in sync?

.vscode/tasks.json
WybrenKoelmans: vscode is setup by default to treat this file as jsonc, 

https://code.visualstudio.com/docs/debugtest/tasks#:~:text=HelloWorld.js%20file.-,You,-can%20also%20define

it's very common to use comments in .vscode/*.json files

mix.exs
WybrenKoelmans: the `--strict` was part of `lint.yml` as well, but i guess it's superfluous then

WybrenKoelmans:
> My main question is why add this rather than tweaking/adding to the mix precommit task?

This is to make sure we have something that a dev can run exactly what will always be the same result as the github CI checks to avoid surprises at the PR stage.

> My second question is how to ensure they stay in sync?

I considered making github run the same mix command, but since the the github flow will currently run these in parallel we'll lose that if it's merged into a single command. So I ended up with a comment _"# Keep in step with  .github/workflows/lint.yml"_ 

we can make a "holy grail" `check` command and merge everything into a single command (we'll lose the parallelism on github, but impact might be OK) 

something like this? (so delete `lint` and `precommit`)

| Step | ~~lint~~ | ~~precommit~~ | **check**
|---|---|---|---|
| `deps.unlock --check-unused` | ✓ | ✓ | ✓
| `hex.audit` | ✓ | — | ✓
| `deps.audit --ignore-package-names decimal,gun` | ✓ | — | ✓
| `format --check-formatted` / `format` | check-only | auto-fix | **check-only** (probably doesn't make sense to have it auto-fix on github actions?)
| `compile --warnings-as-errors` | ✓ (`--force` missing) | ✓ (`--force`) | ✓ (`--force`)
| `credo` | ✓ | ✓ | ✓
| `test --raise --warnings-as-errors --exclude needs_attention` | — | ✓ | ✓
| `dialyzer` | — | ✓ | ✓

.vscode/tasks.json
Teifion: Will it let us swap to use jsonc for this so we don't get formatting errors when viewing it in something like github?

Teifion:
Rather than adding a new task, I would much rather expand `mix precommit` unless there's a very good reason.

We don't _need_ the github actions to run super fast so if there's an notable advantage I am happy for them to be combined but would want Lexon and GeekingFrog to weigh in at that stage too.

geekingfrog:
The github workflow in `lint.yaml` doesn't run the steps in parallel. As I see it, `mix lint` is the same as [these lines](https://github.com/beyond-all-reason/teiserver/blob/main/.github/workflows/lint.yml#L62-L82).
I think having `mix lint` is useful and I'd rather have the github workflow to just use this command, so this way there is no sync problem.

not sure why `mix lint` failed for me, it now works fine.

Also, wrt precommit, these 2 are basically the same, and should be merged. I prefer the name `lint`, but that's a tiny nitpick and both should be fine. Perhaps the only weirdness with `precommit` is that running `precommit` in a github action without any commit action is going to look out of place.

.vscode/tasks.json
geekingfrog: Personally I don't care about anything under .vscode. If you are using vscode, you should be used to whatever BS microslop is doing or whatnot. Weird syntax highlighting **on github** is a non issue for me.

Teifion:
I think `lint` implies just formatting while `precommit` contains things like tests and could take longer. I'm happy to have two different tasks if they have quite different run cases and workloads. If they're similar then they should indeed be combined.

geekingfrog:
credo, fmt and dialyzer seems all "linter" tool for me.
But to be really honest, I don't give a shit about that kind of semantic, I find it's all irrelevant. If I need a command to be run often, I have shell history and alias. If the alias in mix.exs file don't do it for me, I don't use them and use my own through shell.
If you want to know how to get "proper" code, check what the CI does to enforce whatever and that's the end of it. If you want to run that often, or any variation, you have a shell

Teifion:
Given the lack of energy from the team regarding this PR I'm going to close it.

PR #1476, opened 2026-08-13
Add quantum command

Teifion:
Quantum mode is also known in Starcraft 2 as Archon mode and causes the
commanders to be combined and shared control given. Useful for coaching
or just a different way to cooperate with each other.

I tested the code manually (setting player_number etc) using IEX and got
it to work reliably. In the testing I force-started the game right away
to prevent any fixing of the IDs from taking place but here it should
be fixed by the consul server any time it is needed.

The intent for future commits is to fix any bugs that arise and add
options based on player feedback.


L-e-x-o-n (approved):

L-e-x-o-n (commented):
A message informing lobby members that quantum was turned on/off would be appreciated.

I would make the game unrated when quantum is enabled.

L-e-x-o-n (approved):

L-e-x-o-n (approved):

PR #1480, opened 2026-08-14
Help fix quantum mode being intermittendly buggy

Teifion:
Quantum mode works in an isolated environment but when balance is performed it can break things. The solution for now is to manually tell the ConsulServer to :tick and in the process fix the IDs whenever balance is performed.

### DIFF
--- a/lib/teiserver/coordinator.ex
+++ b/lib/teiserver/coordinator.ex
@@ -175,16 +175,6 @@ defmodule Teiserver.Coordinator do
     end
   end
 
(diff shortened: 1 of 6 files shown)


lib/teiserver/coordinator.ex
```diff
@@ -175,16 +175,6 @@ defmodule Teiserver.Coordinator do
     end
   end
 
-  @spec send_balancer(T.lobby_id(), any) :: any
```
Teifion, line 178: This is never used so has been removed.

L-e-x-o-n (approved):

PR #1487, opened 2026-08-19
Upgrading from OTP-27 to OTP-28

### LINKED ISSUE #1486: OTP 28
We want to upgrade to OTP 28

[Mise versions lists 28.5 as the latest sub-version](https://mise-versions.jdx.dev/tools/erlang)

Teifion:
Minor Elixir version bump to use the version of OTP.

Encountered a number of issues related to cowboy where we had pinned
dependencies so unpinned those and updated the last few places in the
code which relied on them.

Nostrum is the only library using cowlib and given the way it has used
it we have disabled the audit warning for those dependencies.


.dialyzer_ignore.exs
```diff
@@ -159,8 +156,67 @@
   {"lib/teiserver_web/live/moderation/action/smurf_link.ex", :callback_arg_type_mismatch},
   {"lib/teiserver_web/live/moderation/action/smurf_link.ex", :call},
   {"lib/teiserver/account/tasks/merge_accounts_task.ex", :no_return},
+  # {"lib/teiserver_web/controllers/account/general_controller.ex"},
```
L-e-x-o-n: Why is this commented?

.dialyzer_ignore.exs
```diff
@@ -159,8 +156,67 @@
...
+  {"lib/teiserver/game/libs/match_rating_lib.ex"},
+  {"lib/teiserver/logging/tasks/persist_match_day_task.ex"},
   {"test/support/fixtures/moderation_fixtures.ex"},
-  {"lib/teiserver/moderation.ex"},
   {"lib/teiserver_web/live/battles/match/show.ex"},
-  {"lib/teiserver/moderation/tasks/create_anti_abuse_record_task.ex"}
+  {"lib/teiserver/moderation/tasks/create_anti_abuse_record_task.ex"},
+  {"lib/teiserver/moderation.ex"}
+  # ~r/.*persist_match_day_task\.ex.*/,
+  # ~r/.*exact_compare/
```
L-e-x-o-n: Should this be deleted if unused?

L-e-x-o-n (approved):

.dialyzer_ignore.exs
Teifion: Should have been deleted, I will fix before merge. Thank you.

PR #1488, opened 2026-08-20
Updates user email change process

Teifion:
Updating a user email now:
- Tracks previous email
- Tracks datetime of email change
- Emails previous and new email regarding the change
- Audit logs the event
- Performs the entire process inside a transaction

We also disable the Spring commands for email changes and update the web interface to accomodate the new changes.

### DIFF
--- a/lib/teiserver/account/libs/user_lib.ex
+++ b/lib/teiserver/account/libs/user_lib.ex
@@ -1,6 +1,7 @@
 defmodule Teiserver.Account.UserLib do
   @moduledoc false
 
+  alias Ecto.Changeset
   alias Phoenix.PubSub
   alias Teiserver.Account
   alias Teiserver.Account.Auth
@@ -11,6 +12,7 @@ defmodule Teiserver.Account.UserLib do
   alias Teiserver.CacheUser
   alias Teiserver.Client
   alias Teiserver.Config
+  alias Teiserver.EmailHelper
   alias Teiserver.Helper.StylingHelper
   alias Teiserver.Logging
   alias Teiserver.Repo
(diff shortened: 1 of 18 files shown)


lib/teiserver_web/controllers/account/session_controller.ex
```diff
@@ -98,6 +98,41 @@ defmodule TeiserverWeb.Account.SessionController do
     end
   end
 
+  def refresh_totp(conn, %{"user_id" => user_id, "otp" => otp, "redirect" => redirect}) do
```
Teifion, line 101: We add this to have a page where we can force a refresh-check of the token.

lib/teiserver/account/libs/user_lib.ex
```diff
@@ -254,15 +256,56 @@ defmodule Teiserver.Account.UserLib do
...
-    |> UserCacheLib.decache_user_on_ok()
+    # We do everything as a transaction so we can revert the change if something goes wrong
+    # with the email sending. Unfortunately there is no way to un-send the email to the old
+    # address if something goes wrong with sending to the new address
+    result =
+      Repo.transact(fn ->
+        with %Changeset{valid?: true} = changeset <- User.changeset(user, attrs, :email),
+             {:ok, %User{} = user} <- Repo.update(changeset),
+             {:ok, %{}} <- EmailHelper.email_changed(user) do
+          Logging.add_audit_log(user.id, ip_address, "email_change_success", %{})
```
L-e-x-o-n, line 274: Should the audit log details contain old and new email address? I think it could be useful.
(Same for failure audit log.)

lib/teiserver/common/email_helper.ex
```diff
@@ -2,12 +2,18 @@ defmodule Teiserver.EmailHelper do
...
   alias Teiserver.Config
   alias Teiserver.Helper.DateHelper
   alias Teiserver.Mailer
   alias Teiserver.Telemetry
+
   require Logger
 
+  defp host, do: Application.get_env(:teiserver, TeiserverWeb.Endpoint)[:url][:host]
+
+  @privacy_email "privacy@beyondallreason.info"
```
L-e-x-o-n: We already have `privacy_email` in runtime config, maybe that one can be used here.

lib/teiserver/common/email_helper.ex
```diff
@@ -174,4 +178,103 @@ defmodule Teiserver.EmailHelper do
     |> Email.text_body(text_body)
     |> Mailer.deliver(response: true)
   end
+
+  def email_changed(%User{} = user) do
+    # We specifically send to the old address first to ensure if something goes wrong we
+    # do not send anything to the new address
```
L-e-x-o-n, line 184: If something goes wrong when? While sending the old address?

lib/teiserver_web/controllers/account/general_controller.ex
```diff
@@ -21,38 +22,55 @@ defmodule TeiserverWeb.Account.GeneralController do
...
-        user_params
-      else
-        Map.drop(user_params, ["name"])
-      end
+      user_params =
+        if Config.get_site_config_cache("user.Enable renames") do
+          user_params
+        else
+          Map.drop(user_params, ["name"])
+        end
```
L-e-x-o-n: The email changeset doesn't seem to actually support updating the name anymore so the name part should be dropped or the changeset updated IMO.

lib/teiserver_web/templates/admin/user/tab_details.html.heex
```diff
@@ -84,6 +84,12 @@
       value={@user.discord_id}
       width={200}
     />
+
+    <.detail_line
+      label="Last email(s)"
+      value={Enum.join(@user.previous_emails, ", ")}
+      width={200}
+    />
```
L-e-x-o-n, line 92: Minor detail, maybe put it next to email.

lib/teiserver/account/libs/user_lib.ex
Teifion: No, for GDPR reasons we specifically do not want to store them. I almost included them in the first draft though by mistake!

lib/teiserver/common/email_helper.ex
Teifion: I looked for it but didn't spot it, might have been a typo. I agree and will swap it over.

lib/teiserver/common/email_helper.ex
Teifion: Yeah, if there's an error sending the email we want to revert it. If we have a badly configured email service for example. We are not _expecting_ anything to go wrong though.

lib/teiserver_web/controllers/account/general_controller.ex
Teifion: Agreed, will change that.

lib/teiserver_web/templates/admin/user/tab_details.html.heex
Teifion: It's something that should be used infrequently and rarely so I'd rather leave it here. Most of the time it should be empty.

lib/teiserver/common/email_helper.ex
L-e-x-o-n: In that case I am not sure if it will work as you intended. I had some difficult with `Mailer.deliver/2` documentation, but I think it will just return an error tuple without raising an exception. If that's the case then the first delivery failing wouldn't block the second email.

Teifion: Do you have an example or docs showing what it would return? From what I could tell it returns `{:ok, %{}}` on success but I'm not 100% confident.

L-e-x-o-n: Sending a new email through IEX without setting from address results in `{:error, :from_not_set}`. You don't check if the return is `:ok` after the sending to old email.

L-e-x-o-n (approved):

PR #1493, opened 2026-08-22
Add ability to include arguments in consul commands

Teifion:
Adds the ability include named arguments in consul commands, testing it with Quantum mode though the actual change won't do anything so is not documented yet.

### DIFF
--- a/lib/teiserver/coordinator/consul_commands.ex
+++ b/lib/teiserver/coordinator/consul_commands.ex
@@ -1007,33 +1007,9 @@ defmodule Teiserver.Coordinator.ConsulCommands do
     end
   end
 
-  # Enables quantum mode where each team has 1 commander and full shared control
-  # next step will be to add options to it allowing you to pick how many commanders
-  # are shared on a team or which teams are used but for now we're keeping it simple
+  # Disables quantum mode if enabled
   def handle_command(
-        %{command: "quantum", remaining: _remaining, senderid: senderid} = _cmd,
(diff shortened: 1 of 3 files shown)


lib/teiserver/protocols/spring/spring_in.ex
```diff
@@ -1392,7 +1392,7 @@ defmodule Teiserver.Protocols.SpringIn do
       [_full_match, battlestatus, team_colour] ->
         updates =
           Spring.parse_battle_status(battlestatus)
-          |> Map.take([:ready, :player_number, :team_number, :player, :sync, :side])
+          |> Map.take([:ready, :team_number, :player, :sync, :side])
```
Teifion, line 1395: Fixes a potential bug where clients changed player numbers before the server could change them back.

lib/teiserver/coordinator/consul_commands.ex
```diff
@@ -1007,46 +1007,56 @@ defmodule Teiserver.Coordinator.ConsulCommands do
     end
   end
 
-  # Enables quantum mode where each team has 1 commander and full shared control
-  # next step will be to add options to it allowing you to pick how many commanders
-  # are shared on a team or which teams are used but for now we're keeping it simple
+  # Disables quantum mode if enabled
```
Teifion, line 1010: Most of this is me swapping the order of the two commands so the quantum command can be called again to change the arguments without disabling it.

L-e-x-o-n (approved):

PR #1496, opened 2026-08-25
Adds the AntiAbuseRecord struct

Teifion:
Sits inside Moderation, different from other structs in that all access functions will log access via the Scope.

Added a slightly modified Scope from a default Phoenix project to allow more powerful access control in the future and to make passing around acces context for AntiAbuseRecord structs easier.

Also:
- Added a default `get_ip_from_conn` result of "127.0.0.1" for local environments

### DIFF
--- a/lib/teiserver/logging/lib/audit_log_lib.ex
+++ b/lib/teiserver/logging/lib/audit_log_lib.ex
@@ -114,6 +114,6 @@ defmodule Teiserver.Logging.AuditLogLib do
 
   def order_by(query, "Newest first") do
     from logs in query,
-      order_by: [desc: logs.inserted_at]
+      order_by: [desc: logs.inserted_at, desc: logs.id]
   end
 end
--- a/lib/teiserver/logging/lib/logging_helpers.ex
+++ b/lib/teiserver/logging/lib/logging_helpers.ex
@@ -39,7 +39,7 @@ defmodule Teiserver.Logging.Helpers do
     {:ok, the_log} =
       Logging.create_audit_log(%{
         action: action,
-        user_id: scope.user.id,
+        user_id: scope.user && scope.user.id,
         details: details,
         ip: scope.ip
       })
--- a/lib/teiserver/logging/lib/logging_helpers.ex
+++ b/lib/teiserver/logging/lib/logging_helpers.ex
@@ -1,6 +1,7 @@
 defmodule Teiserver.Logging.Helpers do
   @moduledoc false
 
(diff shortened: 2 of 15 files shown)


lib/teiserver/logging/lib/audit_log_lib.ex
```diff
@@ -114,6 +114,6 @@ defmodule Teiserver.Logging.AuditLogLib do
 
   def order_by(query, "Newest first") do
     from logs in query,
-      order_by: [desc: logs.inserted_at]
+      order_by: [desc: logs.inserted_at, desc: logs.id]
```
Teifion, line 117: In tests if you had multiple logs created they had the same insertion timestamp so you wouldn't actually get the newest first. This fixes that edge case without altering behaviour.

lib/teiserver/logging/lib/logging_helpers.ex
```diff
@@ -32,8 +33,20 @@ defmodule Teiserver.Logging.Helpers do
     the_log
   end
 
-  @spec add_audit_log(Plug.Conn.t() | Phoenix.LiveView.Socket.t(), String.t(), map()) ::
+  @spec add_audit_log(Scope.t() | Plug.Conn.t() | Phoenix.LiveView.Socket.t(), String.t(), map()) ::
```
Teifion, line 36: Purely adds a new implementation of the function for when you pass in a `Scope` instead of the usual items.

lib/teiserver/logging/lib/logging_helpers.ex
```diff
@@ -32,8 +33,20 @@ defmodule Teiserver.Logging.Helpers do
...
           Teiserver.Logging.AuditLog.t()
+  def add_audit_log(%Scope{} = scope, action, details) do
+    {:ok, the_log} =
+      Logging.create_audit_log(%{
+        action: action,
+        user_id: scope.user.id,
```
L-e-x-o-n: `scope.user` can be `nil` and then this will break

Teifion: Good catch, I will add something for that. In theory it should always be called with an active scope but yes, we don't want this to fail!

Teifion: Addressed in https://github.com/beyond-all-reason/teiserver/pull/1496/commits/c77160a4eaef7b72b5e90477b8a0a36813964290

L-e-x-o-n (approved):

PR #1501, opened 2026-08-27
Fix mobile ratings filter navigation

### LINKED ISSUE #1500: iPhone/iPad: ratings filter opens a previously viewed match
### What happened

On iPhone and iPad, tapping **Small Team** or **Large Team** under **Matches → Ratings** opens the match overview of a match I viewed previously instead of switching the ratings list.

### Steps to reproduce

1. Open a match overview.
2. Navigate to **Matches → Ratings**.
3. Tap **Small Team** or **Large Team**.

### Expected behaviour

The selected ratings list opens.

### Actual behaviour

The previously viewed match overview opens instead.

### Devices

- iPhone
- iPad

eun-ice:
Fixes #1500

## Cause

The ratings table made each match row clickable with Bootstrap’s `stretched-link`. Its absolutely positioned pseudo-element relied on `position: relative` on a `<tr>`. Safari/WebKit does not reliably use a positioned table row as that pseudo-element’s containing block, allowing the invisible link overlay to extend beyond its row and intercept taps on the rating-type filters.

## Fix

- Navigate from the row with `JS.navigate/1`, matching the existing clickable-table pattern used elsewhere in Teiserver.
- Keep a visually hidden real link for accessible navigation.
- Add regression coverage ensuring match rows remain navigable without `stretched-link` or `position-relative`.

## Testing

- `MIX_ENV=test TEI_PORT=0 TEI_METRICS_SERVER_PORT=0 mix test test/teiserver_web/live/battles/match/ratings_live_test.exs`
- Result: 3 tests, 0 failures.
- Human verification on a real iPhone after reproducing the issue on iPhone and iPad. Author confirmation: “manually confirmed, the fix works!”

## AI usage disclosure

OpenAI Codex was used to investigate the WebKit-specific interaction, author the HEEx change and regression test, and assist with this pull request description. The human contributor reproduced the original bug on iPhone and iPad, reviewed the intended behavior, and manually verified the fix on an iPhone.

### DIFF
--- a/lib/teiserver_web/live/battles/match/ratings.html.heex
+++ b/lib/teiserver_web/live/battles/match/ratings.html.heex
@@ -99,7 +99,7 @@
                 true ->
                   {"text-secondary column-delta", "pause"}
               end %>
-            <tr class="position-relative">
+            <tr phx-click={log.match && JS.navigate(~p"/battle/#{log.match_id}")}>
               <%= if log.match do %>
                 <td>
                   {log.match.map}
@@ -139,7 +139,9 @@
 
               <%= if log.match do %>
                 <td>
-                  <.link navigate={~p"/battle/#{log.match_id}"} class="stretched-link"></.link>
+                  <div class="visually-hidden">
+                    <.link navigate={~p"/battle/#{log.match_id}"}>Show</.link>
+                  </div>
                 </td>
               <% else %>
                 <td>&nbsp;</td>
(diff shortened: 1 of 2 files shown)


Teifion (approved):
Nice first PR. I had a few thoughts about shortening the test before realising the functions I was going to suggest using were not present because we never made them so bonus points for creating the structs used in the test.

PR #1504, opened 2026-08-27
Improves the visual design of the smurf link page, fixes to scopes

Teifion:
Improves the css styling of the smurf link page and fixes issues with the way Scopes are assigned within liveview pages.

Light mode - Empty form
image

Dark mode - Empty form
image

Light mode - Mid search
image

Dark mode - Mid search
image

Additionally:
- Allows running of tests while running the server locally (test port conflict)
- Fixes issue with getting connection IP from LiveView sockets
- Fixes Moderation/Action search button

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -262,5 +262,9 @@ if enable_discord_bridge do
     bot_name: Teiserver.ConfigHelpers.get_env("TEI_DISCORD_BOT_NAME")
 end
 
+# If we are running the server locally this will conflict, this allows us to test
+# if it is already defined, if not then it will set it to a new value as a default
+monitoring_port = Application.get_env(:teiserver, TeiserverWeb.Monitoring)[:port] || 4001
+
 config :teiserver, TeiserverWeb.Monitoring,
-  port: Teiserver.ConfigHelpers.get_env("TEI_METRICS_SERVER_PORT", 4001, :int)
+  port: Teiserver.ConfigHelpers.get_env("TEI_METRICS_SERVER_PORT", monitoring_port, :int)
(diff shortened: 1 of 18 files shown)


lib/teiserver/account/libs/auth_lib.ex
```diff
@@ -117,6 +118,15 @@ defmodule Teiserver.Account.AuthLib do
...
   end
 
+  # Scope
+  def allow?(%Scope{} = scope, permissions_required) do
+    %{id: id, roles: roles} = scope.user
+    permissions_held = get_permissions_from_roles(roles)
+
+    mfa_test(id, permissions_required) and
+      permission_test(permissions_held, permissions_required)
+  end
```
L-e-x-o-n: I think this can be done cleaner. Something like this?
```ex
  def allow?(%Scope{user: user} = scope, permissions_required), do: allow?(user, permissions_required)
```

config/runtime.exs
```diff
@@ -262,5 +262,14 @@ if enable_discord_bridge do
...
+default_monitor_port = 4001
+
+existing_port =
+  Application.get_env(:teiserver, TeiserverWeb.Monitoring)[:port] || default_monitor_port
+
+monitor_port = Teiserver.ConfigHelpers.get_env("TEI_METRICS_SERVER_PORT", existing_port, :int)
+
 config :teiserver, TeiserverWeb.Monitoring,
-  port: Teiserver.ConfigHelpers.get_env("TEI_METRICS_SERVER_PORT", 4001, :int)
+  port: Teiserver.ConfigHelpers.get_env("TEI_METRICS_SERVER_PORT", monitor_port, :int)
```
L-e-x-o-n: Do we even need `monitor_port` here? Seems unnecessary, you could just use `existing_port` instead.

Teifion: You are right, I've overcooked it :dagger:

L-e-x-o-n (approved):

PR #1506, opened 2026-08-27
Add matchmaking state to user/self

### LINKED ISSUE #1450: add matchmaking state to `user/self` event
New data expected for that event: https://github.com/beyond-all-reason/tachyon/pull/150 so that client can know which state they're in wrt matchmaking.

burnhamrobertp:
Adds the player's matchmaking state to the user/self event. Fixes #1450

The role_updated path is in here because it is the other emitter of user/self and it was raising a KeyError: it gets handed the connection state, which has never carried party or battle. Leaving it would mean the new field only ever reaches a client on login and reconnect.

Verified against a local server with two tachyon clients and the bar-lobby client from beyond-all-reason/bar-lobby#711, covering reconnects while queuing, paired, paired after readying up, and queued in two queues at once.


lib/teiserver/player/session.ex
```diff
@@ -1350,7 +1350,8 @@ defmodule Teiserver.Player.Session do
            paired_queue: paired_queue,
            room: room_pid,
            frozen_queues: other_queues,
-           readied?: false
+           readied?: false,
+           timeout_at: DateTime.utc_now() |> DateTime.add(timeout_ms, :millisecond)
```
geekingfrog: Prefer [Datetime.shift/3](https://elixir.hexdocs.pm/main/DateTime.html#shift/3)

lib/teiserver/player/session.ex
```diff
@@ -38,6 +38,14 @@ defmodule Teiserver.Player.Session do
 
   @type conn_state :: :connected | :reconnecting | :disconnected
 
+  @type self_state :: %{
```
geekingfrog, line 44: this could do with a `@typedoc` since it's not clear what `self` is referring to there.

lib/teiserver/player/session.ex
```diff
@@ -669,25 +677,17 @@ defmodule Teiserver.Player.Session do
...
+    self_state = build_self_state(state)
 
     party_state =
       %PT.PartyState{
-        version: if(current_party == nil, do: nil, else: current_party.version),
-        current_party: if(current_party == nil, do: nil, else: current_party.id),
-        invited_to: Enum.map(invited_to, fn st -> {st.version, st.id} end)
+        version: if(self_state.party == nil, do: nil, else: self_state.party.version),
+        current_party: if(self_state.party == nil, do: nil, else: self_state.party.id),
+        invited_to: Enum.map(self_state.invited_to_parties, fn st -> {st.version, st.id} end)
```
geekingfrog: This shouldn't rely on `self_state` but instead keep using the original `state`. `self_state` should only be used to be sent to the handler to then forward that to the user. But for any other internal bookeeping, `state` should be preferred.

This way there is a clear one way relationship between these two piece of data, self_state is fully derived from state.

geekingfrog (commented):
side note: I think it would have been better to split the commits a bit.
The introduction of a named type `queue_ref` is good, but completely unrelated to anything else, and could have been in its own commit. It makes reviewing easier.

similarly for the modification to the role updated path.

just the small datetime.shift thingy and it should be good to go

geekingfrog (approved):

geekingfrog:
I'll let you fix the merge conflicts, and then I'll address https://github.com/beyond-all-reason/teiserver/issues/1517 afterward.

burnhamrobertp:
Merge conflicts? I think those were resolved with the last set of changes I had made

geekingfrog:
hum, somehow github cannot rebase and merge, but is fine with creating a merge commit or squashing.

burnhamrobertp:
Ah if that's what you had meant, I could've done that, certainly. Just let me know if you refer that pattern for teiserver specifically, and I can at least try to remember to adhere to it

PR #1509, opened 2026-08-30
Refactor session battle

geekingfrog:
Some refactoring around entering a battle. see commits for more details

### DIFF
--- a/lib/teiserver/autohost/types/player.ex
+++ b/lib/teiserver/autohost/types/player.ex
@@ -13,4 +13,14 @@ defmodule Teiserver.Autohost.Types.Player do
           name: String.t(),
           password: String.t()
         }
+
+  def new(user_id, name) do
+    password = :crypto.strong_rand_bytes(16) |> Base.encode16()
+
+    %__MODULE__{
+      user_id: user_id,
+      name: name,
+      password: password
+    }
+  end
 end
--- a/lib/teiserver/player/session.ex
+++ b/lib/teiserver/player/session.ex
@@ -221,7 +221,7 @@ defmodule Teiserver.Player.Session do
   def lobby_join_battle(user_id, battle_data, battle_start_data, password) do
     user_id
     |> via_tuple()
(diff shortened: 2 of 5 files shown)


lib/teiserver/player/session.ex
```diff
@@ -221,7 +221,7 @@ defmodule Teiserver.Player.Session do
   def lobby_join_battle(user_id, battle_data, battle_start_data, password) do
     user_id
     |> via_tuple()
-    |> GenServer.cast({:battle, {:lobby_join, battle_data, battle_start_data, password}})
+    |> GenServer.cast({:battle, {:start, battle_data, battle_start_data, password}})
```
Teifion, line 224: We're joining a battle, why are we sending a `:start` atom?

lib/teiserver/autohost/types/player.ex
```diff
@@ -13,4 +13,14 @@ defmodule Teiserver.Autohost.Types.Player do
           name: String.t(),
           password: String.t()
         }
+
+  def new(user_id, name) do
+    password = :crypto.strong_rand_bytes(16) |> Base.encode16()
```
Teifion, line 18: Is there a specific reason we're limiting this to just 16? The password won't be readable so no reason to limit it to just that.

Teifion (approved):
Couple of questions/thoughts but I am okay with it.

lib/teiserver/player/session.ex
geekingfrog: Ultimately everything is a "join battle", since this path is about connecting to a battle running on an autohost. perhaps we should rename everything `:join`, though I feel this is a bit out of scope.

The `:start` is coming from the tachyon side, where such request is known as `battle/start` to the client

lib/teiserver/autohost/types/player.ex
geekingfrog: I kind of randomly picked a number. 128bit of entropy seems enough, but changing it to whatever should be fine

PR #1511, opened 2026-09-02
Adds web interface for viewing Anti-abuse records

Teifion:
- Adds a warning page before viewing any AAR pages in general
- Adds a warning header to the page, warning that this data should not be screenshotted or streamed
- Adds tests and a test helper for tables
- Swaps the admin home page to use the new Tailwind styling

New admin menu page
image

Anti-abuse warning page
image
image

Record list page with fake data
image

Record show page with fake data, expecting to expand this later once we have the creation process hammered out more
image


lib/teiserver_web/components/nav_components.ex
```diff
@@ -197,12 +198,15 @@ defmodule TeiserverWeb.NavComponents do
         <li class={[@active == "microblog" && "menu-active"]}>
           <.link href={~p"/microblog"}>Blog</.link>
         </li>
-        <%= if @current_user do %>
+        <%= if @current_user || @scope do %>
```
Teifion, line 201: In some cases we are still calling this using `current_user` so this allows the best of both worlds during the transition.

lib/teiserver_web/router.ex
```diff
@@ -648,10 +648,36 @@ defmodule TeiserverWeb.Router do
...
+        {UserAuthentication, {:authorise_any, ["Contributor", "Overwatch"]}}
+      ] do
+      live "/", MenuLive, :show
+    end
+
+    live_session :admin_antiabuse,
+      layout: {TeiserverWeb.Layouts, :admin_tw},
+      on_mount: [
+        {Teiserver.Account.DefaultsPlug,
+         {:set, %{site_menu_active: "admin", sensitive_data: true}}},
```
Teifion, line 668: Assigning `sensitive_data` here means the warning banner appears for all pages in the session.

test/support/conn_case.ex
```diff
@@ -63,4 +63,111 @@ defmodule TeiserverWeb.ConnCase do
     Tachyon.tachyon_case_setup(tags)
     {:ok, conn: ConnTest.build_conn()}
   end
+
```
Teifion, line 66: An idea from a previous workplace, it makes validating table contents significantly easier. I used ChatGPT to build out the foundation before tweaking it myself.

mix.exs
```diff
@@ -74,7 +74,7 @@ defmodule Teiserver.MixProject do
       {:lazy_html, ">= 0.1.0", only: :test},
       # see https://hexdocs.pm/phoenix_html/changelog.html#v4-0-0-2023-12-19
       {:phoenix_html_helpers, "~> 1.0"},
-      {:floki, ">= 0.37.0", only: :test},
+      {:floki, ">= 0.37.0"},
```
Teifion, line 77: No longer `test` only because otherwise the compilation of the table extractor will generate warnings for dev.

L-e-x-o-n (approved):

PR #1514, opened 2026-09-02
Add modlobby as available client

tetrisface:
I have made a lobby focused more on tweaking, mods and experimentation.

This PR allows me to use the legacy lobby protocol with the right identifier and publish it. I can of course publish it on your request.

I also included a unit test. I don't have a strong opinion on including it. It could be useful to have for just documenting the parsing as well though. Perhaps wanted to remove just for being a deprecated protocol.

# Testing
I have tested locally that it connected and set lobby_client on the client in iex.

# AI disclosure
Code written by Claude Fable 5. Small parts cleaned up and edited by me.


Teifion:
We are completely dropping support for the SPRING protocol (like, none of it will work at any stage). Can you tell me more about this client you are developing please? In this case I'm not thinking we need the unit test and potentially happy to just add the optimisation level setting as a 1-line change, the tests will just add CI time.

tetrisface:
Will remove the test.

It's pretty geared towards custom lobbies and PvE since that's what I'm interested in.
The core is in rust and frontend is tauri/typescript.
The most alternative things it does is calling a third party api for bringing in some stats per lobby settings and adding a custom lua widget to support fluid switch to the lobby as an immersive overlay while in-game.
But it also brings some obvious improvements like bringing TLS support to the connection with teiserver and using OS keychain for the login. 
At the moment it's Windows only.

Teifion:
Cool, I'm happy to include it for that. Be aware we won't be officially supporting this lobby and the support for the protocol will be dropped abruptly. If you want to contribute to development of the new client I am sure they'd be interested in your support though :)

tetrisface:
great! thanks! :)

tetrisface:
lobby available here for reference https://github.com/tetrisface/modlobby

PR #1525, opened 2026-09-06
Extended $status command

Teifion:


L-e-x-o-n (commented):
Some users are avoiding/ignoring/blocking up to 300 users (configurable, used to be no limit and some people had more than a thousand which caused issues for them). Listing 300 names per relationship could break things and would be hard to read in Chobby anyway.

The user's admin page has a relationship view, I suggest using that instead, maybe put a link to the user's admin page for moderators.

Teifion:
I want to add it to allow me to see if any users I've blocked are joining my games, in this case there should not be that many users in the game so the limit shouldn't matter here.

lib/teiserver/coordinator/consul_commands.ex
```diff
@@ -130,6 +131,32 @@ defmodule Teiserver.Coordinator.ConsulCommands do
       |> List.flatten()
       |> Enum.filter(fn v -> v != nil end)
 
+    # Moderator info regarding relationships
+    player_status =
+      if Auth.moderator?(senderid) or Auth.admin?(senderid) do
```
L-e-x-o-n: Every admin is a moderator, so only `moderator?` is necessary.

Teifion: Sorry, we'd used that pattern elsewhere so I was replicating it but it was only used because of a bug at the time. I will fix :)

L-e-x-o-n (approved):

PR #1530, opened 2026-09-08
First stage of GDPR forget and restore task

Teifion:
Provides the scaffolding around forgetting and restoring users, ensuring the forgotten data is encrypted appropriately.

### DIFF
--- a/lib/teiserver/moderation/tasks/create_anti_abuse_record_task.ex
+++ b/lib/teiserver/moderation/tasks/create_anti_abuse_record_task.ex
@@ -1,0 +1,155 @@
+defmodule Teiserver.Moderation.CreateAntiAbuseRecordTask do
+  @moduledoc """
+  Given a user, create an Anti-Abuse record.
+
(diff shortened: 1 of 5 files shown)


lib/teiserver/moderation/tasks/create_anti_abuse_record_task.ex
```diff
@@ -0,0 +1,158 @@
...
+
+  @doc """
+  Wrapper around &Argon2.hash_pwd_salt/1 to allow us to produce a nil value
+  if there was not a value for this in the first place.
+  """
+  def one_way_encrypt_value(nil), do: nil
+
+  def one_way_encrypt_value(value) do
+    Argon2.hash_pwd_salt(value)
+  end
```
L-e-x-o-n: Unused

L-e-x-o-n (approved):

lib/teiserver/moderation/tasks/create_anti_abuse_record_task.ex
Teifion: Good catch, removed.

PR #1532, opened 2026-09-09
Altered login flow for users with gdpr_forget_after set

Teifion:
Users with `gdpr_forget_after` set are unable to login via the normal protocol and must instead go via the website where we can ensure they are explicitly told that logging in will remove the forget process (and log the entire thing).

Screenshot of the forgotten confirmation page
image

### DIFF
--- a/lib/teiserver/account/libs/user_lib.ex
+++ b/lib/teiserver/account/libs/user_lib.ex
@@ -604,6 +604,9 @@ defmodule Teiserver.Account.UserLib do
       Account.get_user_totp_status(user.id) == :active ->
         {:requires_mfa, user}
 
+      gdpr_forget_in_progress?(user) ->
+        {:gdpr_forget_is_set, user}
+
       true ->
         :ok
     end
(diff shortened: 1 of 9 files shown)


Teifion:
@PtaQQ I'm not done with the PR yet but I need your approval/input on the wording used for this page.

PtaQQ:
Heading: Your account is scheduled for deletion

Body:
You asked us to delete your Beyond All Reason account [USERNAME]. It is scheduled for permanent deletion on [DATE, UTC]. Nothing has been deleted yet.

Logging in now cancels that request and keeps your account.

If you still want your account deleted, do not log in. Close this page and the deletion goes ahead on [DATE, UTC].

Buttons:
- Cancel deletion and log in (primary)
- Keep deletion scheduled (secondary)

Two questions on behaviour:
- what does the second button do, log them out and return to the homepage? Whatever it is, the page should make it obvious they're not logged in afterwards. 
- do we have an email for when deletion is cancelled? I forgor. It should be sent. Let me know if you need wording for this one too.

Teifion:
> what does the second button do, log them out and return to the homepage? Whatever it is, the page should make it obvious they're not logged in afterwards.

When they access this page they are not logged in, this happens before the login is finalised. If they decline it currently just sends them to beyondallreason.info but I can change that easily.

> do we have an email for when deletion is cancelled? I forgor. It should be sent. Let me know if you need wording for this one too.

We do not have one, I don't recall seeing one so wording for that would be appreciated. (Edit: We do, this is not needed)

Teifion:
@PtaQQ this is the warning pane revised based on your input. I plan to keep the design very simple (there are no other elements on the page aside from this), probably no other visual changes unless specifically requested.

In the example the username is `AlsoTeifion`, that's not a weird typo.

image

Teifion:
I realise the email has been defined, I will add it as part of https://github.com/beyond-all-reason/teiserver/pull/1535.

L-e-x-o-n (approved):

PR #1533, opened 2026-09-09
Implement Tachyon battle/ended event

### LINKED ISSUE #1497: Implement `battle/ended` event
Client does not get `battle/ended` after a match is over. It's required for post-match results and permitting the client to reliably know that the host has shut down (e.g. do not allow re-joins for user's whose game crashed, etc).

L-e-x-o-n:
Closes #1497


lib/teiserver/tachyon_battle/battle.ex
```diff
@@ -292,6 +299,46 @@ defmodule Teiserver.TachyonBattle.Battle do
...
+  defp notify_battle_ended(state) do
+    players =
+      for {ally_team, ally_id} <- Enum.with_index(state.start_script.ally_teams),
+          {team, team_id} <- Enum.with_index(ally_team.teams),
+          {player, player_id} <- Enum.with_index(team.players) do
+        %{
+          userId: to_string(player.user_id),
+          allyTeam: to_string(ally_id),
+          team: to_string(team_id),
+          player: to_string(player_id)
```
geekingfrog: this is serialization to tachyon protocol, which should only happen in the handlers. Elsewhere we should stick to elixir data structures, either bare maps or "proper" structs

lib/teiserver/tachyon_battle/battle.ex
```diff
@@ -292,6 +299,46 @@ defmodule Teiserver.TachyonBattle.Battle do
...
+    spectators =
+      for id <- Map.keys(state.participants),
+          not MapSet.member?(player_ids, to_string(id)),
+          do: %{userId: to_string(id)}
+
+    winning_ally_team_ids =
+      Battle.list_match_memberships(search: [match_id: state.match_id, win: true])
+      # Match membership team ID is engine ally team ID
+      |> Enum.map(&to_string(&1.team_id))
+      |> Enum.uniq()
```
geekingfrog: Making a db call there is deeply fishy. The battle process should already have all the info required for that. You get this from the `finished` event. Since you now want to remember it, store it in the process' state.

test/teiserver_web/tachyon/lobby_test.exs
```diff
@@ -289,10 +289,37 @@ defmodule TeiserverWeb.Tachyon.LobbyTest do
...
+      Tachyon.autohost_send_update_event(
+        ctx[:autohost_client],
+        Tachyon.autohost_engine_quit(battle_id)
+      )
+
+      player_id = to_string(ctx[:user].id)
+      spectator_ids = MapSet.new([to_string(ctx2[:user].id), to_string(ctx3[:user].id)])
+
+      for client <- [ctx[:client], ctx2[:client], client3] do
+        messages = Tachyon.drain(client)
```
geekingfrog: Why do you need to drain there? Are there more unwanted messages? drain has a timeout built in and makes the tests much slower. Generally, we have rather tight control over which messages get sent, so it's better to use `Tachyon.recv!` whenever possible.
This also allow simpler assertion and pattern matching afterwards, no need for Enum.find/any.

lib/teiserver/tachyon_battle/battle.ex
```diff
@@ -292,6 +299,46 @@ defmodule Teiserver.TachyonBattle.Battle do
...
+
+    battle_ended_data = %{
+      battleId: state.id,
+      players: players,
+      spectators: spectators,
+      winningAllyTeamIds: winning_ally_team_ids
+    }
+
+    Map.keys(state.participants)
```
geekingfrog, line 346: I am not sure `participants` in its current form is fit for purpose. If a player joins the battle, then leave, they will be in this map, but are no longer interested in the battle_ended event.
perhaps track whether they are currently in the battle as well, using `player_left` events for that? That could probably be in a separate commit however since it's a bit different than the support for the `battle_ended` event.

lib/teiserver/player/session.ex
```diff
@@ -219,6 +219,15 @@ defmodule Teiserver.Player.Session do
     |> GenServer.cast({:battle, {:start, battle_data, battle_start_data, password}})
   end
 
+  @doc """
+  Let players and spectators know a battle has ended
+  """
+  def notify_battle_ended(user_id, battle_id, battle_ended_data) do
+    user_id
+    |> via_tuple()
+    |> GenServer.cast({:battle, {:end, battle_id, battle_ended_data}})
```
geekingfrog: nitpick: perhaps `{:battle, {:ended, ...}}` so it matches the names?

lib/teiserver/player/session.ex
```diff
@@ -1515,6 +1524,19 @@ defmodule Teiserver.Player.Session do
     end
   end
 
+  def handle_cast(
+        {:battle, {:end, battle_id, battle_ended_data}},
+        %PT.Data{} = state
+      ) do
+    case state.battle do
+      %{id: ^battle_id} ->
+        {:noreply, send_to_player(state, {:battle_ended, battle_ended_data})}
```
geekingfrog: I think you can simply use `send_to_player!` which doesn't mutate the state. Because `user/self` has `currentBattle`, which would be null when they reconnect, so we don't need to store this event and retransmit when they reconnect.

However, you also need to adjust the session state because they are not in a battle anymore (don't forget to remove the monitors as well)

lib/teiserver/player/session.ex
geekingfrog, line 1: You also need to send the battle/ended event when the battle process explodes and you get a `:DOWN` message. Although at that point it's a bit fucked because the session doesn't have enough data to send the correct event to the player. We would need to store the start script in the session for that, or at least the subset required to construct the event. You can put a placeholder event with a TODO for now.

lib/teiserver/tachyon_battle/battle.ex
```diff
@@ -292,6 +299,46 @@ defmodule Teiserver.TachyonBattle.Battle do
...
+          allyTeam: to_string(ally_id),
+          team: to_string(team_id),
+          player: to_string(player_id)
+        }
+      end
+
+    player_ids = MapSet.new(players, & &1.userId)
+
+    # All participants who are not players are considered spectators
+    spectators =
```
Teifion, line 334:
```suggestion
    # All participants who are not players are considered spectators
    # An issue to change the way we track spectators - https://github.com/beyond-all-reason/teiserver/issues/1534
    spectators =
```

lib/teiserver/tachyon_battle/battle.ex
L-e-x-o-n: I was thinking a bit more about this. I think tracking `player_left` might not be the best approach, I can't say for certain but I can see autohost sending this when a player leaves the match at its end, e.g. when done checking end game stats... I think just checking if user's session battle id is matching the one for which the notify event is sent is sufficient

Teifion (approved):

PR #1536, opened 2026-09-10
Fix stale connection_timeout stopping a reconnected session

### LINKED ISSUE #1505: Session stopped early when a connection drops twice inside 30s
Session stopped 1.1s after a second abrupt disconnect, instead of 30s.

Join a matchmaking queue, drop the connection, reconnect, drop again inside 30s.

The connection_timeout armed by the first drop is never cancelled when the player reconnects. It fires 30s later, finds conn_pid nil from the second drop, and stops a session that had already reconnected. The timer armed by the second drop still had 29s to run.

Tagged each armed timer with temporary logging to get this:

```
19:24:53.705  CT_ARM id=4  fires_in=30000ms user=3
19:25:03.757  CT_RECONNECT matchmaking={:searching, joined_queues: [{"1v1", "30009617"}]}
19:25:22.590  CT_ARM id=6  fires_in=30000ms user=3
19:25:23.707  CT_FIRE id=4 conn_pid=nil matchmaking={:searching, ...} decision=STOP
19:25:23.704  Player timed out, stopping session
```

Takes party and battle state with it. Paired players get matchmaking/lost when it happens mid-pairing.

Three places arm it (session.ex:584, :1680, :1777), all discarding the ref, and :timer.cancel is only used for lobby_list_subscription.

jadhavgaurav:
A player session that disconnects, reconnects, and disconnects again inside
the 30 second connection_timeout window can be killed by the wrong timer.

Each abrupt disconnect (outside of a battle) arms a fresh connection_timeout
via `:timer.send_after(@connection_timeout, :connection_timeout)`, but the
generic `:connection_timeout` message carries no way to tell which disconnect
armed it. If a player disconnects, reconnects, then disconnects again before
the first timer fires, the first (stale) timer still stops the session when
it eventually fires, even though the second disconnect's own timer still had
time left and the player may already be reconnected or paired for a match.
The session takes its party and battle state down with it.

Fix: tag each armed timeout with a unique reference kept in the session
state, and only act on a `:connection_timeout` message whose reference
matches the one currently stored. A stale timer from an earlier disconnect
is then a no-op; only the most recently armed timer can stop the session.

Added a regression test that disconnects, reconnects, disconnects again, then
delivers the stale timer's message directly and confirms the session survives,
before confirming the current timer still stops it correctly. Verified the
test fails on the unmodified source with a `KeyError` (the field the fix adds
does not exist yet) by reverting only the source change and rerunning it.

Checks run locally (Erlang/OTP 27.3.4, Elixir 1.19.4, matching mise.toml):
- `mix format --check-formatted`: clean
- `mix compile --warnings-as-errors`: clean
- `mix credo --strict` on the changed files: no issues
- `mix test --exclude needs_attention`: full suite passes

AI disclosure: this change (root cause analysis, fix, and the regression
test) was written with Claude Code, per AI_POLICY.md. The fix and tests were
run and verified locally before opening this PR.

Fixes #1505.


Teifion:
Please tell me more about how you encountered/discovered this error and what the effect was. The entire PR description looks a little like it was written by an LLM so I want to check some context around this PR :).

Teifion:
As per the AI_Policy all messages and discussions must be written by a human, not an LLM. The team has agreed this is clearly not the case here.

PR #1550, opened 2026-09-15
Adds GDPR anonymise task

Teifion:


lib/teiserver/coordinator/coordinator_server.ex
```diff
@@ -344,7 +344,7 @@ defmodule Teiserver.Coordinator.CoordinatorServer do
             icon: "fa-solid fa-sitemap",
             colour: "#AA00AA",
             password: Account.make_bot_password(),
-            roles: ["Bot", "Verified"],
+            roles: ["Bot", "Verified", "Server"],
```
Teifion, line 347: We make the coordinator a Server user so nobody but server admins are able to access it.

L-e-x-o-n: Might be a good idea to do the same for other server bots: AutohostMonitor and DiscordbridgeBot.

Teifion: Done, now all 3 have that permission.

L-e-x-o-n (approved):

PR #1549, opened 2026-09-14
Replace Discord prefix commands with / application commands

L-e-x-o-n:
- Prefix commands replaced with application / commands
- Removed whatwas/whatis unit commands
- Removed user message bridging
- Removed bot DM commands and DM channels
- Linking now done through website instead of chat with Coordinator

PR #1551, opened 2026-09-15
Fix decaching users

L-e-x-o-n:
Important for lookup caches where we were decaching the new value instead of the old one


L-e-x-o-n:
Depends on #1549

Teifion (approved):

PR #1564, opened 2026-09-16
Handle user/report for tachyon clients

burnhamrobertp:
Handles user/report from tachyon clients, so the new bar-lobby reporting flow stops falling through to command_unimplemented.

The two json files are load bearing, not spec tidying. The transport validates against the vendored schema before it dispatches, so with user/report missing from priv/tachyon the request never reaches a handler and the new clause on its own would have changed nothing.

Reports for several people arrive as one request but become one row each with nothing tying them together, since the table only has a single target. If they should be linked that wants a real decision rather than me picking one here. I also matched the two guards the web flow has at report_user/index.ex, refusing self-reports and anyone carrying the Reporting restriction, since otherwise this is a way around something the web path enforces.


L-e-x-o-n (approved):
Seems good to me, fix the merge conflict.

Not for this PR, but `user/report` should probably be updated at some point to include a `battle_id` (or some other way to connect a match to a report) and the option to select different report `sub_type` .

L-e-x-o-n (approved):

PR #1580, opened 2026-09-18
Created demo of command palette

Teifion:
The command palette is located only on /admin/palette and only enabled via the component hook. It will not be present anywhere else on the site.

The list of commands is intentionally limited for the purpose of evaluating the command palette.

ChatGPT was used for some troubleshooting.

Palette page
image

Palette in action
image
image


assets/js/app.js
```diff
@@ -6,17 +6,15 @@ import { LiveSocket } from "phoenix_live_view"
 // import {hooks as colocatedHooks} from "phoenix-colocated/teiserver"
 import topbar from "../vendor/topbar_tw"
 import LiveCharts from "live_charts"
-
-// TODO: Is this used any more? I think we can probably remove the UserSocket and thus this line
-// Local files
-import socket from "./socket"
```
Teifion, line 12: This was generating a 2nd websocket connection and from what I can see is no longer used. The UserSocket module is still used from what I can see so I have left that in place.

lib/teiserver_web/components/nav_components.ex
```diff
@@ -377,8 +377,6 @@ defmodule TeiserverWeb.NavComponents do
   """
   attr :url, :string, required: true
   attr :icon, :string, required: true
-  attr :icon_class, :string, default: "solid"
```
Teifion, line 380: These were not actually used

L-e-x-o-n (approved):

